PR #4, opened 2019-09-27
Solution

frucelestine:


src/transformState.js
```diff
@@ -49,7 +49,23 @@
...
  */
 function transformState(state, transforms) {
-  // write code here
+  function deleteProps(obj, props) {
+    for (const p of props) {
+      (p in obj) && (delete obj[p]);
+    }
+  }
+  for (const keys in transforms) {
+    if (transforms[keys].operation === 'addProperties') {
```
earthhuman41: Switch will be better here instead of if

src/transformState.js
```diff
@@ -49,7 +49,23 @@
...
+      Object.assign(state, transforms[keys].properties);
+    }
+    if (transforms[keys].operation === 'removeProperties') {
+      deleteProps(state, transforms[keys].properties);
+    }
+    if (transforms[keys].operation === 'clear') {
+      for (const prop of Object.keys(state)) {
+        delete state[prop];
```
earthhuman41: if you wrote function for deleting properties, why you didn't use here?

PR #29, opened 2019-10-07
Fs aug19 slava

MMashivskij:


src/transformState.js
```diff
@@ -49,7 +49,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  transforms.forEach(function(item, i, arr) {
```
gestap: why you need `arr` and `i`?
use arrow function for method
```suggestion
  transforms.forEach((item) => {
```

src/transformState.js
```diff
@@ -49,7 +49,23 @@
...
+    if (item.operation === 'addProperties') {
+      for (const key in item.properties) {
+        state[key] = item.properties[key];
```
gestap: better to use `Object.assign` for this task

src/transformState.js
```diff
@@ -49,7 +49,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  transforms.forEach(function(item, i, arr) {
+    if (item.operation === 'addProperties') {
```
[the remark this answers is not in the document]
gestap: also, try to use destruction method

src/transformState.js
```diff
@@ -49,7 +49,28 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  transforms.forEach(item => {
```
gestap:
```suggestion
  transforms.forEach((item) => {
  const [operation, properties] = item;
```

ost-ptk (approved):

PR #30, opened 2019-10-08
statefulobject

YeaaaaH:


src/transformState.js
```diff
@@ -49,7 +49,22 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    switch (transforms[i].operation) {
```
gestap: try to use destruction method and use `for...in` instead of `for`

PR #59, opened 2020-01-23
add solution, fix lint errors

danheim:


src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+      case 'addProperties': {
+        return Object.assign(state, properties);
+      }
+
+      case 'removeProperties': {
+        return properties.forEach(property => delete state[property]);
+      }
+
+      case 'clear': {
+        return Object.keys(state).map(key => delete state[key]);
```
YuriiHoliuk: not correct usage of map

src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+      }
+
+      default: {
+        return state;
```
YuriiHoliuk: you do not need to return from forEach, will be better to remove return statements and add breaks

src/transformState.test.js
```diff
@@ -4,126 +4,229 @@ const transformState = require('./transformState');
 
 test('Should add a single property to an empty state', () => {
   const state = {};
+
```
YuriiHoliuk: why do you change test file?

PR #63, opened 2020-01-23
transform

mariiafil:


src/transformState.js
```diff
@@ -47,6 +47,25 @@
...
+        delete state[key];
+      }
+    }
+
+    if (obj.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  });
```
danheim: use switch case

src/transformState.js
```diff
@@ -47,6 +47,27 @@
...
+        for (const key of obj.properties) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
```
danheim: Add brackets
```
case 'clear': {
  // code
}
```

src/transformState.js
```diff
@@ -47,6 +47,27 @@
...
+        for (const key of obj.properties) {
+          delete state[key];
+        };
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
```
danheim: Add brackets
```
case 'clear': {
  // code
}
```

mariiafil: don't understand why I need them

PR #66, opened 2020-01-23
AnastasiiaBardyuzha:  js_stateful-object solution

AnastasiiaBardyuzha:


src/transformState.js
```diff
@@ -46,7 +46,27 @@
...
+
+        delete state[removeKey];
+      }
+    }
+
+    if (transforms[i]['operation'] === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
danheim: switch case

src/transformState.js
```diff
@@ -46,7 +46,30 @@
...
+
+          delete state[removeKey];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
danheim: add brackets
```
case 'clear': {
 // some code
 break;
}
```

PR #67, opened 2020-01-23
solution

LutsenkoAnV:


src/transformState.js
```diff
@@ -46,7 +46,29 @@
...
+      for (const key in resultObj) {
+        delete resultObj[key];
+      }
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (const key of transforms[i].properties) {
+        delete resultObj[key];
+      }
+    }
```
danheim: switch case, не забывай добавлять break

src/transformState.js
```diff
@@ -46,7 +46,34 @@
...
+        for (const key in resultObj) {
+          delete resultObj[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of transforms[i].properties) {
+          delete resultObj[key];
+        }
+        break;
```
danheim: Add brackets
```
case 'clear': {
  // code
}
```

src/transformState.js
```diff
@@ -46,7 +46,34 @@
...
+
+      default:
+        return state;
+    }
```
danheim: in this case, you don't need default

PR #82, opened 2020-01-24
Sapuzhak: solution

Mikhail-88:


src/transformState.js
```diff
@@ -45,8 +45,35 @@
...
+    for (const key of obj) {
+      delete state[key];
+    }
+  }
+
+  function clear() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
```
irenhh: this is almost the same function

src/transformState.js
```diff
@@ -45,8 +45,35 @@
...
+
+  for (const item of transforms) {
+    if (item.operation === 'addProperties') {
```
irenhh: use switch

PR #83, opened 2020-01-24
solution

oksanaDaniluk:


src/transformState.js
```diff
@@ -46,7 +46,28 @@
...
+      case 'addProperties': {
+        Object.assign(state, transforms[i].properties);
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const property of transforms[i].properties) {
+          delete state[property];
+        }
+        break;
```
irenhh: add a new line after block of code

PR #84, opened 2020-01-24
Solution

tynindima:


src/transformState.js
```diff
@@ -46,7 +46,28 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const index of transforms) {
```
irenhh: why index?

tynindima: changed to obj

src/transformState.js
```diff
@@ -46,7 +46,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const obj of transforms) {
+    switch (obj.operation) {
+      case 'addProperties':
+        Object.assign(state, obj.properties);
+        break;
+      case 'removeProperties':
```
irenhh: add a new line before new case

irenhh:
image
what about this?

src/transformState.js
tynindima: Why I needed default case? it will have nothing

src/transformState.js
```diff
@@ -46,7 +46,29 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
+
+      default:
```
irenhh: consider comment above

PR #86, opened 2020-01-25
solution

BohdanSalenik:


src/transformState.js
```diff
@@ -43,10 +43,33 @@
...
+      case 'addProperties': {
+        Object.assign(state, element.properties);
+        break;
+      }
+
+      case 'removeProperties': {
+        element.properties.forEach((value) => {
+          delete state[value];
+        });
+        break;
```
irenhh: add a new line after block of code

PR #88, opened 2020-01-26
solution

BDenysovets:


src/transformState.js
```diff
@@ -46,7 +46,30 @@
...
+      case 'addProperties' : {
+        Object.assign(state, item.properties);
+        break;
+      }
+
+      case 'removeProperties' : {
+        item.properties.forEach(value => {
+          delete state[value];
+        });
+        break;
```
irenhh: add a new line after a block of code for better readability

PR #89, opened 2020-01-26
transform-state

serverieva:


src/transformState.js
```diff
@@ -46,7 +46,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (transform['operation']) {
+      case 'addProperties':
+        for (const property in transform['properties']) {
+          if (transform['properties'].hasOwnProperty(property)) {
```
irenhh: use Object.assign

src/transformState.js
```diff
@@ -51,20 +51,25 @@ function transformState(state, transforms) {
...
         }
         break;
+
       case 'removeProperties':
         for (const propertyToRemove of transform['properties']) {
           delete state[propertyToRemove];
         }
         break;
+
+      default:
```
irenhh: use can combine one of the case with default case so that you don't need to create something like this

PR #91, opened 2020-01-27
'solutions'

maximkacmm:


src/transformState.js
```diff
@@ -46,7 +46,46 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  for (let i = 0; i < transforms.length; i++) {
+    const objects = transforms[i];
```
irenhh: why is it called objects?

src/transformState.js
```diff
@@ -46,7 +46,46 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  for (let i = 0; i < transforms.length; i++) {
```
irenhh: you don't need this cycle

src/transformState.js
maximkacmm: transforms это массив обектов, а transforms[i] это элемент массива (обект). Логичней назвать 
object , а не objects.

src/transformState.js
```diff
@@ -46,7 +46,31 @@
...
+    const operations = transforms[i].operation;
+    const properties = transforms[i].properties;
+
+    switch (operations) {
+      case 'addProperties' :
+        for (const key in properties) {
+          state[key] = properties[key];
+        }
+        break;
```
irenhh: new line after block of code

src/transformState.js
```diff
@@ -46,7 +46,31 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
-}
+  for (let i = 0; i < transforms.length; i++) {
+    const operations = transforms[i].operation;
+    const properties = transforms[i].properties;
```
irenhh: this could be done with destructuring

PR #95, opened 2020-01-27
hardy...

ded-git:


src/transformState.js
```diff
@@ -46,7 +46,31 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    switch (transforms[i]['operation']) {
```
irenhh: use destructuring

src/transformState.js
```diff
@@ -46,7 +46,31 @@
...
+      case 'addProperties':
+        for (const key in transforms[i]['properties']) {
+          state[key] = transforms[i]['properties'][key];
+        };
```
irenhh: new line after the block of code

PR #173, opened 2020-05-27
solution Polishchuk

PolishchukViacheslav:


src/transformState.js
```diff
@@ -47,22 +47,23 @@
  */
 function transformState(state, transforms) {
   for (const [index] of transforms.entries()) {
```
ogsevko, line 49: First of all, you shouldn't use square brackets in variable names. 
Secondly, `for (const variableName of transforms)` would be fine. You don't need entries here. Then you can use variableName.operation or variableName.properties to access keys in objects inside transforms

PolishchukViacheslav: напиши мне в  slack есть пара вопросов

src/transformState.js
```diff
@@ -46,7 +46,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const index of transforms) {
```
danheim, line 49:
```suggestion
  for (const action of transforms) {
```

src/transformState.js
```diff
@@ -46,7 +46,30 @@
...
+    switch (index.operation) {
+      case 'addProperties':
+        for (const prop in index.properties) {
```
danheim, line 52:
```suggestion
        for (const key in action.properties) {
```

src/transformState.js
```diff
@@ -46,7 +46,30 @@
...
+          state[prop] = index.properties[prop];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const value of index.properties) {
```
danheim, line 58: лучше по ключам

PR #196, opened 2020-08-03
done

dan3si:


src/transformState.js
```diff
@@ -47,6 +47,25 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const trans of transforms) {
```
```suggestion
  for (const transform of transforms) {
```
why did you shorten it?

src/transformState.js
```diff
@@ -47,6 +47,25 @@
...
+    if (trans.operation === 'addProperties') {
+      for (const proper in trans.properties) {
```
vkryvytskyy, line 52:
```suggestion
      for (const property in trans.properties) {
```

src/transformState.js
```diff
@@ -47,6 +47,25 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const trans of transforms) {
+    if (trans.operation === 'addProperties') {
```
vkryvytskyy, line 51: consider to use switch..case here

PR #197, opened 2020-08-03
solution

ArtemiiKab:


ArtemiiKab:
Did that,  but why switch preferable?

PR #198, opened 2020-08-03
solved

DmJSCourse:


src/transformState.js
```diff
@@ -46,7 +46,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const action of transforms) {
```
vsm2k17, line 49: that's not an action that's an object with action and properties
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -46,7 +46,27 @@
...
+    if (action['operation'] === 'addProperties') {
+      Object.assign(state, action['properties']);
+    }
+
+    if (action['operation'] === 'removeProperties') {
+      for (const key in action['properties']) {
```
vsm2k17, line 55:
```suggestion
      for (const key in action.properties) {
```

PR #200, opened 2020-08-03
solution

DolgKreker:


asbh25 (approved):

src/transformState.js
```diff
@@ -46,7 +46,31 @@
...
  */
 function transformState(state, transforms) {
-  // write code here
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+  const clear = 'clear';
+
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === add) {
+      for (const key in transforms[i].properties) {
```
vkryvytskyy, line 55:
```suggestion
  for (const transform of ) {
    if (transform.operation === add) {
      for (const key in transforms[i].properties) {
```

src/transformState.js
```diff
@@ -46,7 +46,31 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+  const clear = 'clear';
+
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === add) {
```
vkryvytskyy, line 54: consider to use if..else or even better switch..case

vkryvytskyi (changes requested):

src/transformState.js
```diff
@@ -46,7 +46,31 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 52:
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -46,7 +46,31 @@
...
+    switch (transforms[i].operation) {
```
```suggestion
    switch (transforms.operation) {
```

vkryvytskyi (approved):
fix this

PR #201, opened 2020-08-03
Solution

YuriiRysyn:


src/transformState.js
```diff
@@ -46,7 +46,27 @@
...
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (item.operation === 'removeProperties') {
+      for (const char of item.properties) {
+        delete state[char];
+      }
+    }
```
vkryvytskyy, line 66: use if..else, or even better switch..case

PR #202, opened 2020-08-03
Transform state solution

asbh25:


src/transformState.js
```diff
@@ -46,7 +46,21 @@
...
+      }
+    } else if (transform.operation === 'removeProperties') {
+      for (const prop of transform.properties) {
+        delete state[prop];
+      }
+    } else if (transform.operation === 'clear') {
+      for (const prop in state) {
+        delete state[prop];
+      }
+    }
```
vkryvytskyy, line 62: consider to use switch..case instead of if..else

PR #203, opened 2020-08-03
solution

oggyrock:


src/transformState.js
```diff
@@ -46,7 +46,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i in transforms) {
+    if (transforms[i].operation === 'addProperties') {
```
nanccyy, line 50: it'd be better to replace iterators i, y, etc. with something more meaningful

src/transformState.js
```diff
@@ -46,7 +46,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i in transforms) {
```
vkryvytskyy, line 49:
```suggestion
  for (const transform in transforms) {
```

src/transformState.js
```suggestion
  for (const transform of transforms) {
    if (transform.operation === 'addProperties') {
```

vkryvytskyy, line 50: consider to use switch..case here, instead of if's(or if..else at least)

oggyrock: а як же:
"РАБОТАЕТ - НЕ ТРОГАЙ!"
?
з.ы. виправив)

src/transformState.js
```diff
@@ -46,7 +46,26 @@
...
+        for (const prop of transform.properties) {
+          delete state[prop];
+        }
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+    }
```
vkryvytskyy, line 65: add default case to switch..case

PR #204, opened 2020-08-03
solution

krajnova:


src/transformState.js
```diff
@@ -46,7 +46,33 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
DmJSCourse, line 72: You can add default: break; to this switch for best practice, even though it is not required with current tests

PR #208, opened 2020-08-03
stateful-object

Elena-brv:


Andrii256:
> let's try to solve this task using "switch"

I think it can be bad practise. Switch is usefull for simple construction. But here it can be difficult to read

src/transformState.js
```diff
@@ -45,8 +45,27 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const obj of transforms) {
```
vsm2k17, line 50: use more meaningful names

PR #209, opened 2020-08-03
Solution

KaterinaSavvova:


src/transformState.js
```diff
@@ -47,6 +47,27 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const i of (transforms)) {
```
```suggestion
  for (const i of transforms) {
```
redundant

src/transformState.js
```diff
@@ -47,6 +47,27 @@
...
+    if (i.operation === 'addProperties') {
+      for (const n in i['properties']) {
```
vkryvytskyy, line 52: what does `n` means?

src/transformState.js
```diff
@@ -47,6 +47,27 @@
...
+        state[n] = i['properties'][n];
+      }
+    }
+
+    if (i.operation === 'clear') {
+      for (const k in state) {
```
vkryvytskyy, line 58: what does `k` means?

src/transformState.js
```diff
@@ -47,6 +47,27 @@
...
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (value.operation === 'removeProperties') {
+      for (const prop of value.properties) {
+        delete state[prop];
+      }
+    }
```
vkryvytskyy, line 67: use if..else or switch..case

PR #210, opened 2020-08-03
solution

VenomSRT:


src/transformState.js
```diff
@@ -46,7 +48,39 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  let currentOperation = '';
+
+  for (const transform of transforms) {
+    for (const transformProp in transform) {
```
vsm2k17, line 54: no need to use second loop

src/transformState.js
```diff
@@ -46,7 +48,39 @@
...
+      if (transformProp === 'operation') {
+        currentOperation = transform[transformProp];
```
vsm2k17, line 56: redundant, no need to extra variable

PR #213, opened 2020-08-04
solved

nanccyy:


src/transformState.js
```diff
@@ -45,8 +45,33 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const key in transforms) {
```
```suggestion
  for (const transform in transforms) {
```

src/transformState.js
```diff
@@ -45,8 +45,33 @@
...
+    if (transforms[key].operation === 'addProperties') {
+      Object.assign(state, transforms[key].properties);
+    }
```
```suggestion
  for (const key of transforms) {
    if (key.operation === 'addProperties') {
      Object.assign(state, transforms[key].properties);
    }
```

src/transformState.js
```diff
@@ -45,8 +45,33 @@
...
+
+    if (transforms[key].operation === 'removeProperties') {
```
vkryvytskyy, line 55: consider to use switch..case here, instead of if..else

PR #215, opened 2020-08-04
solution

oivannikov:


src/transformState.js
```diff
@@ -45,8 +45,48 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 50: try "for of" loop

src/transformState.js
```diff
@@ -45,8 +45,48 @@
...
+}
+
+function addProperties(state, properties) {
+  Object.assign(state, properties);
+
+  return state;
+}
+
+function removeProperties(state, properties) {
+  for (let i = 0; i < properties.length; i++) {
```
vsm2k17, line 74: try "for in" loop

src/transformState.js
```diff
@@ -45,8 +45,47 @@
...
+      addProperties(state, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      removeProperties(state, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'clear') {
+      clear(state, transforms[i].properties);
+    }
```
vsm2k17, line 61: consider of using switch construction or if => else (there is no way that you received two different operations, so no need letting your program acting in this way)

src/transformState.js
vsm2k17: for of loop in this situation much more comfortable to use

src/transformState.js
```diff
@@ -45,8 +45,43 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const transform in transforms) {
+    if (transforms[transform].operation === 'addProperties') {
```
vsm2k17, line 51:
```suggestion
  for (const transform of transforms) {
    if (transform.operation === 'addProperties') {
```

vsm2k17: "for of" with arrays
"for in" with objects

src/transformState.js
```diff
@@ -45,8 +45,43 @@
...
+  Object.assign(state, properties);
+
+  return state;
+}
+
+function removeProperties(state, properties) {
+  for (const properti in properties) {
+    if (state.hasOwnProperty(properties[properti])) {
+      delete state[properties[properti]];
+    }
```
vsm2k17, line 73:
```suggestion
  for (const property of properties) {
    if (state.hasOwnProperty(property)) {
      delete state[property];
    }
```

PR #223, opened 2020-08-04
solution

Oleksandr-Rohatnov:


src/transformState.js
```diff
@@ -40,13 +40,36 @@
...
- * {foo: 'bar', name: 'Jim'}.
- *
- * @param {Object} state
- * @param {Object[]} transforms
- */
+//  * {foo: 'bar', name: 'Jim'}.
+//  *
+//  * @param {Object} state
+//  * @param {Object[]} transforms
+//  */
```
vkryvytskyy, line 47: why are you added single line comments here?

src/transformState.js
```diff
@@ -40,13 +40,36 @@
...
+
 function transformState(state, transforms) {
-  // write code here
+  for (const obj in transforms) {
+    const main = transforms[obj];
```
vkryvytskyy, line 51:
```suggestion
  for (const tranform of transforms) {
```

src/transformState.js
```diff
@@ -40,13 +40,36 @@
...
+
+    switch (main.operation) {
```
```suggestion
    switch (transform.operation) {
```

src/transformState.js
```diff
@@ -40,13 +40,36 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
vkryvytskyy, line 69: always add default case for switch

PR #227, opened 2020-08-04
Solution

kostyasabada:


package-lock.json
```diff
@@ -463,12 +463,13 @@
       }
```
vkryvytskyy, line 463: remove this file from PR

src/transformState.js
```diff
@@ -46,7 +46,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i of transforms) {
```
vkryvytskyy, line 49:
```suggestion
  for (const transform of transforms) {
```
what does i means?

src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+
+    if (i.operation === 'addProperties') {
+      Object.assign(state, i.properties);
+    }
+
+    if (i.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
vkryvytskyy, line 64: use if..else, or even better switch..case

src/transformState.js
```diff
@@ -46,7 +46,24 @@
...
+    if (transform.operation === 'removeProperties') {
+      for (const key of transform.properties) {
+        delete state[key];
+      }
+    } else
+    if (transform.operation === 'addProperties') {
+      Object.assign(state, transform.properties);
+    } else
+
+    if (transform.operation === 'clear') {
```
vkryvytskyy, line 59:
```suggestion
    } else if (transform.operation === 'clear') {
```

PR #230, opened 2020-08-04
Solution

leonbohdan:


src/transformState.js
```diff
@@ -46,7 +46,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const obj in transforms) {
+    const currentObj = transforms[obj];
```
vsm2k17, line 50: at this moment you got in your obj indexes of the array, use "for of" loop for arrays

src/transformState.js
```diff
@@ -46,7 +46,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const obj in transforms) {
```
vsm2k17, line 49: use more meaningful names

src/transformState.js
leonbohdan: transforms is an object, that why I use for...in loop

src/transformState.js
leonbohdan: renamed

PR #231, opened 2020-08-04
Soluton

iamtheru:


src/transformState.js
```diff
@@ -46,7 +46,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 49: use `for of` loop instead of usual for

PR #238, opened 2020-08-04
solution

nikita-parkhomenko:


src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+      for (const prop of transform.properties) {
+        delete state[prop];
+      }
+    }
+
+    if (transform.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
vkryvytskyy, line 64: use if..else or switch..case instead of separate ifs

PR #241, opened 2020-08-04
Solution

1nsider21:


src/transformState.js
```diff
@@ -46,7 +46,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const value of transforms) {
+    if (value.operation === 'addProperties') {
+      for (const i in value.properties) {
+        state[i] = value.properties[i];
```
tykhan, line 52: щоб не робити ще один цикл, можна скористатись Object.assign

tykhan (approved):

src/transformState.js
vsm2k17, line 52:
```suggestion

        state = {
        ...state,
        ...properties,
```

src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+      }
+    } else if (value.operation === 'removeProperties') {
+      for (const i of value.properties) {
```
vsm2k17, line 55: "i" - that bad name, when you use "for of" loop, better to use something more meaningful

src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+        if (state[i]) {
+          delete state[i];
+        }
+      }
+    } else if (value.operation === 'clear') {
+      for (const i in state) {
```
vsm2k17, line 61: "i" that's a bad name

mykola-yarchuk (approved):
consider of using switch construction

PR #243, opened 2020-08-05
solution

ChrisZinch:


src/transformState.js
```diff
@@ -46,7 +46,25 @@
...
+        for (const key in properties) {
+          delete state[properties[key]];
+        }
+        break;
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
+    }
```
vkryvytskyy, line 64: always add default case for switch

PR #321, opened 2020-11-26
Solution

anatoliy-dotsenko:


src/transformState.js
```diff
@@ -62,7 +62,33 @@
...
+    }
+
+    if (addPror.operation === 'removeProperties') {
+      for (const removeKey of addPror.properties) {
+        for (const key in state) {
+          if (removeKey === key) {
+            delete state[key];
+          }
+        }
+      }
```
vkryvytskyy, line 87: you need only one loop here, if state hasn't this key, it's do nothing

vkryvytskyy (changes requested):

src/transformState.js
anatoliy-dotsenko: I don't understand, that should I change. If I'm using one loop, it doesn't work correctly

anatoliy-dotsenko: for (const removeKey of addPror.properties) {
  delete state[removeKey];
}

vkryvytskyy (approved):

PR #322, opened 2020-11-30
Solution

mark-kochura:


src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i of transforms) {
```
vkryvytskyy, line 65:
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    for (const j in i) {
```
vkryvytskyy, line 66:
```suggestion
    for (const key in transform) {
```

vkryvytskyy (changes requested):
fix naming, it's unreadable in this way

vkryvytskyy (approved):

PR #327, opened 2020-12-04
Develop

Vasilkoff:


src/transformState.js
```diff
@@ -62,7 +62,26 @@
...
+      case 'addProperties':
+        Object.assign(state, transforms[i].properties);
+        break;
+      case 'removeProperties':
+        for (let j = 0; j < transforms[i].properties.length; j++) {
+          for (const key in state) {
+            if (key === transforms[i].properties[j]) {
+              delete state[key];
+            }
+          }
```
vkryvytskyy, line 76: just delete all props from transform[i], you don't need inner for loop

vkryvytskyy (approved):

PR #332, opened 2020-12-08
Stateful object solution

VinnieJ-2k20:
Линтер на дает прописать
`state = {}`
Пришлось идти длинным путем.

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
+      Object.assign(state, transform.properties);
+    }
+
+    if (transform.operation === 'removeProperties') {
+      for (const property of transform.properties) {
+        delete state[property];
+      }
+    }
+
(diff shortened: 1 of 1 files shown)


Lara-l-art (approved):
well done

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+      for (const property of transform.properties) {
+        delete state[property];
+      }
+    }
+
+    if (transform.operation === 'clear') {
+      const keys = Object.keys(state);
+
+      for (const key of keys) {
+        delete state[key];
```
varseniuk, line 80:
```suggestion
      for (const key in state) {
        delete state[key];
```

Loop `for...in` iterates through properties of the object. There's no need to create a new array of keys

varseniuk (changes requested):
Let's use here `switch-case` instead of `if`

varseniuk (changes requested):
'Switch' statements always should contain a 'default' case.
Reasons why:
- always should be an error-handling
- there may be unknown type of arguments
- there may be any kinds of exception

varseniuk (approved):
Good job! 👍

PR #338, opened 2020-12-09
Solution

vorfolomeevav:


Serhii-Alieksieievych (approved):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === 'addProperties') {
```
varseniuk, line 66: Instead of 'if... else if... else if...' use 'switch...case' statement. It's more appropriate in such kind of tasks

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+      const propertiesKeys = Object.keys(transforms[i].properties);
+
+      for (let j = 0; j < propertiesKeys.length; j++) {
+        state[propertiesKeys[j]] = transforms[i].properties[propertiesKeys[j]];
+      }
```
varseniuk, line 71: You can simplify your code here by removing the loop and using Object.assign for transforms[i].properties

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+    } else if (transforms[i].operation === 'clear') {
+      const stateKeys = Object.keys(state);
+
+      for (let k = 0; k < stateKeys.length; k++) {
+        delete state[stateKeys[k]];
+      }
```
varseniuk, line 77:
```suggestion
      for (const key in state) {
        delete state[key];
      }
```
Why not use 'for...in' loop here and just iterate through the keys of state?
In such case you will not need to create `const stateKeys = Object.keys(state);` and it will make code clearer

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+    } else if (transforms[i].operation === 'removeProperties') {
+      for (let m = 0; m < transforms[i].properties.length; m++) {
+        if (state.hasOwnProperty(transforms[i].properties[m])) {
+          delete state[transforms[i].properties[m]];
+        }
```
varseniuk, line 82:
```suggestion
          delete state[transforms[i].properties[m]];
```
There's no need to check if the 'state' has some property here. because if it hasn't, so 'delete' just will do nothing.

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+        state[propertiesKeys[j]] = transforms[i].properties[propertiesKeys[j]];
+      }
+    } else if (transforms[i].operation === 'clear') {
+      const stateKeys = Object.keys(state);
+
+      for (let k = 0; k < stateKeys.length; k++) {
+        delete state[stateKeys[k]];
+      }
+    } else if (transforms[i].operation === 'removeProperties') {
+      for (let m = 0; m < transforms[i].properties.length; m++) {
```
varseniuk, line 79:
```suggestion
      for (const property of transforms[i].properties) {
```
Here it's better and clearer to use 'for...of' loop to iterate through array. It's much more readable:

varseniuk (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const property of transforms[i].properties) {
+          delete state[property];
+        }
+        break;
+    }
```
varseniuk, line 82: Always try to add 'default' condition to 'switch...case' statements

varseniuk (changes requested):
Good work! 👍 
Just add 'default' case.

varseniuk (approved):

PR #340, opened 2020-12-10
Solution

Iryna713:


AlieksieienkoAndriy (approved):

src/transformState.js
```diff
@@ -62,7 +62,30 @@
...
+    if (transforms[i]['operation'] === 'removeProperties') {
+      for (const key of transforms[i]['properties']) {
+        if (state[key]) {
+          delete state[key];
+        }
+      }
+    }
+
+    if (transforms[i]['operation'] === 'clear') {
+      // state = {};
```
vsm2k17, line 81: linter sad that's incorrect, you can't reassign function arguments

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 65: "for of" loop looks better here

vsm2k17 (changes requested):
great work, but consider use switch case construction or "if-else" at least because at this moment your code can be triggered on every if block (it's impossible, we know it, but still no need to organize logic this way), need to write blocks of code counting on behavior when only one block can be triggered. that's why I'm suggesting to use a switch case.

Denchepornyuk (approved):

PR #341, opened 2020-12-10
Solution

Layrovell:


dtsehelnyk (approved):
хорошая работа)

src/transformState.js
```diff
@@ -61,8 +61,27 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const i of key.properties) {
+          delete state[i];
+        }
+        break;
+
+      case 'clear':
+        for (const i in state) {
```
varseniuk, line 79:
```suggestion
        for (const property of key.properties) {
          delete state[property];
        }
        break;

      case 'clear':
        for (const value in state) {
```

try to use more meaningful names for variables

src/transformState.js
```diff
@@ -61,8 +61,27 @@
...
+          delete state[i];
+        }
+        break;
+    }
```
varseniuk, line 83: Add 'default' case

varseniuk (changes requested):
Always try to add the 'default' case for 'switch...case' statements

varseniuk (approved):
On 'default case' you can log smth to console (like 'Error! Unknown type of data!').
Or just 'break' and stop the function

PR #343, opened 2020-12-10
solution

xipholena:


5Mountains:
Huh, really clever use of destruction.  🧐

5Mountains (approved):

varseniuk (approved):
Good solution 👍 
Use the key word 'default' for default cases

PR #347, opened 2020-12-10
Solution

khrulenko:


dima-bond (approved):

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 65: "for of" loop looks better here, using this type of loop make your code shorter and more readable

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    if (transforms[i].operation === 'addProperties') {
+      Object.assign(state, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (let s = 0; s < transforms[i].properties.length; s++) {
```
vsm2k17, line 71: "for of" loop looks better here, using this type of loop make your code shorter and more readable

vsm2k17 (changes requested):
consider using switch case construction.

vsm2k17 (approved):
great job, but try to always add a default case

PR #348, opened 2020-12-10
solution

NazarDovhanchuk:


olesiatk (approved):

src/transformState.js
```diff
@@ -61,8 +61,27 @@
...
+      case 'removeProperties':
+        for (const key of transforms.properties) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
varseniuk, line 80: where is 'break' here?

src/transformState.js
```diff
@@ -61,8 +61,27 @@
...
+    }
```
varseniuk, line 81: Add 'default' case

varseniuk (changes requested):
Add 'break' to the last case and also add 'default' case.

src/transformState.js
```diff
@@ -61,8 +61,33 @@
...
+          delete state[key];
+        };
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
```
varseniuk, line 82:
```suggestion
        }
        break;
```
Stick to one style of formatting in your code. In other cases there are no empty lines before 'break', but here you insert it for some reason

src/transformState.js
```diff
@@ -61,8 +61,33 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
-function transformState(state, transforms) {
-  // write code here
+
+function transformState(state, transform) {
+  for (const transforms of transform) {
+    switch (transforms.operation) {
+      case 'addProperties' : Object.assign(state, transforms.properties);
```
varseniuk, line 68:
```suggestion
      case 'addProperties' : 
        Object.assign(state, transforms.properties);
```
Formatting

varseniuk (changes requested):
All is good. Bit fix formatting

Denchepornyuk (approved):

Denchepornyuk (approved):

PR #350, opened 2020-12-11
Solution

olesiatk:


NazarDovhanchuk (approved):

src/transformState.js
```diff
@@ -62,7 +62,56 @@
...
+    //       delete state[transforms[i].properties[k]];
+    //     }
+    //   }
+    // }
+
+    // if (transforms[i].operation === `clear`) {
+    //   for (const key in state) {
+    //     delete state[key];
+    //   }
+    // }
```
vsm2k17, line 111: no need to save this, the switch case is better

src/transformState.js
```diff
@@ -62,7 +62,56 @@
...
+
+    switch (operation) {
+      case ('addProperties'):
+        for (const key in properties) {
+          state[key] = properties[key];
+        }
+        break;
+
+      case ('removeProperties'):
+        for (let k = 0; k < properties.length; k++) {
```
vsm2k17, line 76: "for of" loop looks better here

src/transformState.js
```diff
@@ -62,7 +62,56 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 65: "for of" loop looks better here

vsm2k17 (changes requested):
delete comments

src/transformState.js
olesiatk: deleted comments

src/transformState.js
olesiatk: changed

src/transformState.js
olesiatk: changed

aionova (approved):
Good job!

Denchepornyuk (approved):

PR #351, opened 2020-12-11
Develop

DarMatt:


taniazt (approved):

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const obj of transforms) {
```
aionova, line 65: Name your variables properly. 'Obj' doesn't tell anything about your variable. 'Transformation' would fit it better

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    if (obj.operation === 'addProperties') {
```
aionova, line 66: When you have more than 2 conditions, it's better to use switch case.

aionova (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transformation of transforms) {
+    switch (transformation.operation) {
+      case 'addProperties':
+        for (const key in transformation.properties) {
+          state[key] = transformation.properties[key];
+        }
```
Denchepornyuk, line 70: для addProperties не потрібно робити цикл, зроби це через Object.assign

Denchepornyuk (changes requested):

aionova (approved):
Good Job!

PR #353, opened 2020-12-11
Solution

AlieksieienkoAndriy:


Iryna713 (approved):

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vsm2k17, line 65: use "for of" loop, it looks better here

vsm2k17 (changes requested):
great work, but consider use switch case construction or "if-else" at least because at this moment your code can be triggered on every if block (it's impossible, we know it, but still no need to organize logic this way), need to write blocks of code counting on behavior when only one block can be triggered. that's why I'm suggesting to use a switch case.

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i of transforms) {
```
aionova, line 65: Name your variables properly. What does 'i' mean?

aionova (approved):

src/transformState.js
```diff
@@ -62,7 +62,30 @@
...
+        break;
+      }
+
+      case 'clear': {
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+      }
+    }
```
aionova, line 85: add default case in case of error

Denchepornyuk (approved):

PR #358, opened 2020-12-11
solution

deelray:


AlieksieienkoAndriy (approved):

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const item of transforms) {
```
aionova, line 65: 'Transformation' would describe this variable better than 'item'

aionova (approved):
Good job!

aionova (approved):

PR #363, opened 2020-12-12
Solution

HaidaiAndrii:


src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+        delete state[del];
+      }
+    }
+
+    if (transforms[name].operation === 'clear') {
+      for (const del in state) {
+        delete state[del];
+      }
+    }
+  }
```
kerpichonka, line 83: використай switch case і буде бомба))

kerpichonka (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,29 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const name in transforms) {
```
aionova, line 65: Choose proper names for your variables.
It's not name

aionova, line 65: Transforms is an array. Why not use for of loop?

src/transformState.js
```diff
@@ -62,7 +62,29 @@
...
+    switch (transforms[name].operation) {
+      case 'addProperties':
+        for (const prop in transforms[name].properties) {
```
aionova, line 68: You can use Object.assign here

src/transformState.js
```diff
@@ -62,7 +62,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const del of transforms[name].properties) {
+          delete state[del];
+        }
+        break;
+
+      case 'clear':
+        for (const del in state) {
```
aionova, line 80: Name 'del' doesn't tell anything about this variable

aionova (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+        for (const key of transformation.properties) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
aionova, line 81: In switch case it's recommended to include default case for error handling

aionova (approved):

PR #365, opened 2020-12-12
solution

chri911:


AlieksieienkoAndriy (approved):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const key of transforms) {
+    for (const value in key) {
```
aionova, line 66: What do key and value mean here? Transformation and command would be better names for these variables

aionova (approved):

aionova (approved):

PR #373, opened 2020-12-13
solution

Droshik:


Alexandr-Petrenko (approved):

Alexandr-Petrenko:
Good job!

src/transformState.js
```diff
@@ -63,6 +63,23 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
aionova, line 66: why not use 'for of' loop?

src/transformState.js
```diff
@@ -63,6 +63,23 @@
...
+    if (transforms[i].operation === 'clear') {
```
aionova, line 67: Switch case will be more appropriate here than 3 if conditions.

aionova (changes requested):

Denchepornyuk (approved):

PR #375, opened 2020-12-13
task was completed

alexmogwaiii:


Nvalja (approved):

src/transformState.js
```diff
@@ -62,7 +62,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const item of transforms) {
+    if (item.operation === 'addProperties') {
+      for (const key in item.properties) {
```
aionova, line 67: You can do it without loop, using Object.assign

src/transformState.js
```diff
@@ -62,7 +62,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const item of transforms) {
```
aionova, line 65: Name variables properly. It's not item, it's transformation or command

aionova (approved):

PR #377, opened 2020-12-14
Solution

Baburineu:


src/transformState.js
```diff
@@ -63,6 +63,23 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const i of transforms) {
```
vpolets, line 66: transforns - plural
transform - singular
```suggestion
  for (const transform of transforms) {
```

vpolets (approved):

vkryvytskyy (approved):

PR #421, opened 2021-02-11
transform state

yurasokal:


src/transformState.js
```diff
@@ -61,8 +61,27 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
```
AnnaHarshyna, line 67: **switch** may be used here as well

AnnaHarshyna (approved):
lgtm!

src/transformState.js
yurasokal: Please wait, I'm in a proces now)

solaryasha (approved):

PR #428, opened 2021-02-22
Solution

oleksii-slobodian:


src/transformState.js
```diff
@@ -62,7 +62,35 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 65:
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -62,7 +62,35 @@
...
+      Object.assign(state, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (const key in state) {
+        for (let j = 0; j < transforms[i].properties.length; j++) {
+          if (transforms[i].properties[j] === key) {
+            delete state[key];
+          }
+        }
```
vkryvytskyy, line 76: you could omit this if statement, if state doen't have such property, your code will do nothing

vkryvytskyy (approved):

PR #434, opened 2021-03-01
Solution

herkoss:


src/transformState.js
```diff
@@ -62,7 +62,20 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let index = 0; index < transforms.length; index++) {
```
aionova, line 65: 'for of' loop will fit better here

src/transformState.js
```diff
@@ -62,7 +62,20 @@
...
+    if (transforms[index]['operation'] === 'addProperties') {
```
aionova, line 66: use switch case if there are more than 1 else if.

src/transformState.js
```diff
@@ -62,7 +62,24 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (transform['operation']) {
```
```suggestion
    switch (transform.operation) {
```

src/transformState.js
```diff
@@ -62,7 +62,24 @@
...
+      case 'addProperties':
+        for (const key in transform['properties']) {
+          state[key] = transform['properties'][key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
aionova, line 75:
```suggestion
      case 'clear':
        state = {};
```

src/transformState.js
```diff
@@ -62,7 +62,24 @@
...
+        break;
+      case 'removeProperties':
+        for (const key in transform['properties']) {
```
aionova, line 78: better to create a variable for transform['properties'].
state[transform['properties'][key]] is hard to read

aionova (changes requested):

src/transformState.js
herkoss: Linter error

Assignment to function parameter 'state'  no-param-reassign

aionova: Sorry, my bad

aionova (approved):
Looks great now!

PR #440, opened 2021-03-03
Solution

IhorDrul:


src/transformState.js
```diff
@@ -63,6 +63,21 @@
...
+  for (let i = 0; i < transforms.length; i++) {
+    switch (transforms[i].operation) {
+      case 'addProperties' :
+        Object.assign(state, transforms[i].properties);
+        break;
+      case 'removeProperties' :
+        transforms[i].properties.map(prop => delete state[prop]);
+        break;
+      case 'clear' :
+        Object.keys(state).map(prop => delete state[prop]);
```
danheim, line 75: use forEach, you don't need map here

src/transformState.js
```diff
@@ -63,6 +63,21 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    switch (transforms[i].operation) {
+      case 'addProperties' :
+        Object.assign(state, transforms[i].properties);
+        break;
+      case 'removeProperties' :
+        transforms[i].properties.map(prop => delete state[prop]);
```
danheim, line 72: what is prop? that's key

danheim (changes requested):

danheim (approved):

PR #443, opened 2021-03-03
solution

Daaiif:


src/transformState.js
```diff
@@ -62,7 +62,23 @@
...
+        delete state[key];
+      }
+    }
+
+    if (prop.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
danheim, line 81: use switch case

danheim (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const prop of transforms) {
```
danheim, line 65: what is prop? do not use short forms
```suggestion
  for (const transform of transforms) {
```

danheim (changes requested):

danheim (approved):
great

PR #444, opened 2021-03-03
solution

Serhii-Nikitin:


src/transformState.js
```diff
@@ -63,6 +63,26 @@
  */
 function transformState(state, transforms) {
   // write code here
+
+  for (const obj of transforms) {
```
danheim, line 67: don't short names

src/transformState.js
```diff
@@ -63,6 +63,26 @@
...
+    switch (true) {
```
danheim, line 68: google how to use switch case

danheim (changes requested):

danheim (approved):

PR #448, opened 2021-03-03
solution

vitalii-nepomiashchyi:


src/transformState.js
```diff
@@ -62,7 +62,29 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    const currentTransform = transforms[i];
+    const currentTransformName = currentTransform.operation;
+    const currentTransformValue = currentTransform.properties;
```
danheim, line 68: hack (destructuring)
```suggestion    
    const { operation, properties } = currentTransform;
```

danheim (approved):
try to use `for of` loop

PR #455, opened 2021-03-04
Solution

ruslanvasylyshyn:


TarasChaban:
good job!

src/transformState.js
```diff
@@ -63,6 +63,30 @@
  */
 function transformState(state, transforms) {
   // write code here
+
+  for (const key of transforms) {
```
danheim, line 67: call it `transform`

src/transformState.js
```diff
@@ -63,6 +63,30 @@
...
+
+      case 'removeProperties':
+        for (const i of key.properties) {
+          delete state[i];
+        }
+
+        break;
+
+      case 'clear':
+        for (const k in state) {
```
danheim, line 81: you iterate keys, 
```suggestion
        for (const key of transform.properties) {
          delete state[i];
        }

        break;

      case 'clear':
        for (const key in state) {
```

danheim (changes requested):

danheim (approved):

PR #457, opened 2021-03-08
solution

den-bulaev:


src/transformState.js
```diff
@@ -63,6 +63,27 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const object of transforms) {
+    if (object.operation === 'addProperties') {
+      for (const item in object.properties) {
```
aionova, line 68:
```suggestion
      for (const propertyName in object.properties) {
```

src/transformState.js
```diff
@@ -63,6 +63,27 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const object of transforms) {
```
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -63,6 +63,27 @@
...
+    if (object.operation === 'addProperties') {
+      for (const item in object.properties) {
+        state[item] = object.properties[item];
+      }
+    } else if (object.operation === 'removeProperties') {
+      for (const item of object.properties) {
+        if (item in state) {
```
aionova, line 73:
```suggestion
```
this check is redundant

aionova (changes requested):
Good start!
Let's use switch case construction here

src/transformState.js
```diff
@@ -63,6 +63,25 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
+      for (const propertyName in transform.properties) {
+        state[propertyName] = transform.properties[propertyName];
+      }
```
aionova, line 70:
```suggestion
    if (transform.operation === 'addProperties') {
     Object.assign(state, transform.properties);
```

src/transformState.js
```diff
@@ -63,6 +63,25 @@
...
+    } else if (transform.operation === 'removeProperties') {
+      for (const item of transform.properties) {
```
aionova, line 72:
```suggestion
      for (const property of transform.properties) {
```

aionova (changes requested):
Let's use switch case

aionova (changes requested):
Fix comments

aionova (changes requested):
You still didn't fix the comment about switch case

aionova (approved):
Good job)

PR #461, opened 2021-03-11
Solution

mralexfed007:


src/transformState.js
```diff
@@ -62,7 +62,31 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vpolets, line 65: for (const transfrom of transforms)

src/transformState.js
```diff
@@ -62,7 +62,31 @@
...
+          }
+        }
+      }
+    }
+
+    if (transforms[i].operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
vpolets, line 86: use switch case

vpolets (changes requested):

src/transformState.js
vkryvytskyy: use for of loop

vkryvytskyy (changes requested):

vpolets (approved):

PR #463, opened 2021-03-12
solution

SerhiiKolesnyk:


IvanKolodiazhnyi:
Looks good)
Good job!

src/transformState.js
```diff
@@ -63,6 +63,29 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
aliceblunt8, line 66: let's use for of loop instead for with index

src/transformState.js
```diff
@@ -63,6 +63,29 @@
...
+    if (transforms[i].operation === 'addProperties') {
```
aliceblunt8, line 67: let's use switch case here instead of if

src/transformState.js
```diff
@@ -63,6 +63,29 @@
...
+      for (const key in transforms[i].properties) {
+        state[key] = transforms[i].properties[key];
+      }
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (const key of transforms[i].properties) {
+        if (state.hasOwnProperty(key)) {
```
aliceblunt8, line 75: no need for hasOwnProperty check

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #466, opened 2021-03-16
Solution

IvanGrekov:


src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const i of transforms) {
```
```suggestion
  for (const transform of transforms) {
```

aionova (changes requested):
Let's use switch case here. It will fit perfect here instead of 2 'else if' blocks

aionova (approved):
Good job!
Although switch case construction would be better here

IvanGrekov:
> Good job!
> Although switch case construction would be better here

Yes, agree) Thank you) Already remade

PR #468, opened 2021-03-21
Devepol

senkiv-oleh:


den-bulaev:
Good!

VinnieJ-2k20 (changes requested):
Let's rewrite the solution using `for ... of` cycle and `switch/case`

VinnieJ-2k20 (changes requested):
`switch/case` block should always have default option in case of irregular inputs.
In this task you can simply perform `continue` by default.

src/transformState.js
```diff
@@ -62,7 +63,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const k of value.properties) {
+          delete state[k];
+        }
+        break;
+
+      default:
+        break;
```
VinnieJ-2k20, line 85: For the future, in case of irregular inputs you may throw an error.
You'll learn how to do it further in the program.

VinnieJ-2k20 (approved):

PR #469, opened 2021-03-22
Solution

mariiakornieva:


src/transformState.js
```diff
@@ -61,8 +61,33 @@
...
 function transformState(state, transforms) {
-  // write code here
+  for (const { operation, properties } of transforms) {
+    if (operation === 'addProperties') {
+      addProperties(state, properties);
+    } else if (operation === 'removeProperties') {
+      removeProperties(state, properties);
+    } else if (operation === 'clear') {
+      clear(state);
+    }
```
vkryvytskyy, line 89: use switch..case

vkryvytskyy (approved):

src/transformState.js
mariiakornieva: ok, thanks for the suggestion

PR #470, opened 2021-03-23
Solution

SergiyKhomchenko:
Task solution.

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === 'addProperties') {
+      Object.assign(state, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'clear') {
+      for (const keys in state) {
+        delete state[keys];
+      }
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+      for (const keys in state) {
+        delete state[keys];
+      }
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (const key in transforms[i].properties) {
+        delete state[transforms[i].properties[key]];
+      }
+    }
```
vkryvytskyy, line 80: use switch case for such operations

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
vkryvytskyy, line 65: use for..of

vkryvytskyy (changes requested):

SergiyKhomchenko (commented):
Fixed

src/transformState.js
SergiyKhomchenko: Fixed. Code updated.

vkryvytskyy (approved):

PR #472, opened 2021-03-30
Developed solution

RomanDubovskyi:
Object state is mutated according to instruction from transforms object

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -64,12 +64,16 @@
 function transformState(state, transforms) {
   for (const transform of transforms) {
     if (transform.operation === 'clear') {
-      Object.keys(state).forEach(p => delete state[p]);
-    } else if (transform.operation === 'removeProperties') {
+      Object.keys(state).forEach(k => delete state[k]);
+    }
+
+    if (transform.operation === 'removeProperties') {
       for (const forDeletion of transform.properties) {
         delete state[forDeletion];
       }
-    } else if (transform.operation === 'addProperties') {
+    }
+
+    if (transform.operation === 'addProperties') {
       for (const key in transform.properties) {
         state[key] = transform.properties[key];
       }
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'clear') {
+      Object.keys(state).forEach(p => delete state[p]);
+    } else if (transform.operation === 'removeProperties') {
+      for (const forDeletion of transform.properties) {
+        delete state[forDeletion];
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
```
vpolets, line 65:
```suggestion
  for (const { operation, properties } of transforms) {
```

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+      for (const forDeletion of transform.properties) {
+        delete state[forDeletion];
+      }
+    }
+
+    if (transform.operation === 'addProperties') {
+      for (const key in transform.properties) {
+        state[key] = transform.properties[key];
+      }
+    }
```
vpolets, line 80: про switch case знаєш?)

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+      Object.keys(state).forEach(k => delete state[k]);
+    }
+
+    if (transform.operation === 'removeProperties') {
+      for (const forDeletion of transform.properties) {
+        delete state[forDeletion];
+      }
+    }
+
+    if (transform.operation === 'addProperties') {
```
vpolets, line 76: краще виносити літерали в константи

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'clear') {
+      Object.keys(state).forEach(k => delete state[k]);
```
vpolets, line 67:
```suggestion
      Object.keys(state).forEach(key=> delete state[key]);
```

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    }
+
+    if (transform.operation === 'removeProperties') {
+      for (const forDeletion of transform.properties) {
```
vpolets, line 71:
```suggestion
      for (const property of transform.properties) {
```

vpolets (approved):

src/transformState.js
RomanDubovskyi: nice, thanks, didn't think of it

src/transformState.js
RomanDubovskyi: fixed

src/transformState.js
RomanDubovskyi: fixed

src/transformState.js
RomanDubovskyi: it's the same for in sense of readability and scaling, ideally, I would use here some kind of map with handler for each operation)

PR #478, opened 2021-04-21
solution

Andriykooo:


src/transformState.js
```diff
@@ -58,11 +58,36 @@
...
  * { foo: 'bar', name: 'Jim' }
  *
- * @param {Object} state
- * @param {Object[]} transforms
+ * @param {objectect} state
+ * @param {objectect[]} transforms
  */
 function transformState(state, transforms) {
   // write code here
+  for (const object of transforms) {
```
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -58,11 +58,36 @@
...
+
+    if (object.operation === 'addProperties') {
+      for (const key in object.properties) {
+        state[key] = object.properties[key];
+      };
+    };
+    // REMOVE
+
+    if (object.operation === 'removeProperties') {
+      for (const element of object.properties) {
```
aionova, line 77:
```suggestion
      for (const property of object.properties) {
```

aionova (changes requested):
Good start!
Let's use switch case here

aionova (approved):
Good job!

PR #479, opened 2021-04-21
Solution

iryna-vys:


src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    const { operation, properties } = transforms[i];
```
illisogori, line 66:
```suggestion
  for (const { operation, properties } of transforms) {
```

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+
+    if (operation === 'addProperties') {
+      for (const key in properties) {
+        state[key] = properties[key];
+      }
+    }
```
illisogori, line 72:
```suggestion
    if (operation === 'addProperties') {
      Object.assign(state, properties);
    }
```

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+
+    if (operation === 'removeProperties') {
+      for (let n = 0; n < properties.length; n++) {
+        delete state[properties[n]];
+      }
```
illisogori, line 77:
```suggestion
      for (const property of properties) {
        delete state[property];
      }
```

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    const { operation, properties } = transforms[i];
+
+    if (operation === 'addProperties') {
```
illisogori, line 68: Можно сделать через switch.
switch (operation) {
      case 'addProperties' :
      .......
        break;

illisogori (approved):
Good job.

VinnieJ-2k20 (changes requested):
`switch/case` should always have default option in case of some irregular inputs.
You can simply return null or error message by default.

VinnieJ-2k20 (approved):

PR #481, opened 2021-04-21
Solution

igor-stasiv:


Ihor-Karpyn (approved):
In this case use switch is better.  Everything else looks good

VinnieJ-2k20 (changes requested):
Agreed. Let's use `switch/case` structure for this task.

src/transformState.js
```diff
@@ -62,7 +62,28 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const { operation, properties } of transforms) {
```
VinnieJ-2k20, line 65: 👍

VinnieJ-2k20 (approved):
Good job 👍

PR #484, opened 2021-04-21
solution

victor-arutiunov:


aionova (changes requested):
Looks good, but switch case construction will fit perfect here. 
Let's fix it

src/transformState.js
```diff
@@ -62,7 +62,23 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (true) {
+      case transform.operation === 'addProperties':
```
aionova, line 67:
```suggestion
    switch (transform.operation) {
      case 'addProperties':
```

aionova (changes requested):
Let's fix it once again

aionova (approved):
Looks great now!

PR #485, opened 2021-04-21
solution

Namart:


iryna-vys (approved):
Looks good to me

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const property of transforms) {
```
aliceblunt8, line 65:
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    if (property.operation === 'removeProperties') {
+      for (const value of property.properties) {
+        delete state[value];
+      }
+    }
+
+    if (property.operation === 'clear') {
+      for (const item of Object.keys(state)) {
+        delete state[item];
+      }
```
aliceblunt8, line 79: let's use 'for in' loop here

aliceblunt8 (changes requested):
let's rewrite this to switch case

src/transformState.js
```diff
@@ -61,8 +61,28 @@
...
+        break;
+      case 'removeProperties':
+        for (const value of transform.properties) {
+          delete state[value];
+        }
+        break;
+
+      case 'clear':
+
+        for (const item of Object.keys(state)) {
```
aliceblunt8, line 79: don't use Object.keys here, let's use 'for in' loop to iterate through state's keys

src/transformState.js
```diff
@@ -61,8 +61,28 @@
...
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (transform.operation) {
+      case 'addProperties':
+        Object.assign(state, transform.properties);
+        break;
+      case 'removeProperties':
```
aliceblunt8, line 71: add an empty line between each case
```suggestion

      case 'removeProperties':
```

src/transformState.js
```diff
@@ -61,8 +61,28 @@
...
+        for (const value of transform.properties) {
+          delete state[value];
+        }
```
aliceblunt8, line 74: it's not value, it's key
```suggestion
        for (const key of transform.properties) {
          delete state[key];
        }
```

aliceblunt8 (changes requested):

mykola-yarchuk (approved):
Good job

PR #486, opened 2021-04-21
first solution

scarlettjohanssons:


src/transformState.js
```diff
@@ -63,6 +63,37 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    const obj = transforms[i];
```
Ihor-Karpyn, line 67: You can use for in, and get rid line 67

src/transformState.js
```diff
@@ -63,6 +63,37 @@
...
+    if (obj.operation === 'removeProperties') {
+      const properties = obj.properties;
+
+      for (let j = 0; j < properties.length; j++) {
+        for (const key in state) {
+          if (properties[j] === key) {
+            delete state[key];
+          }
+        }
+      }
```
Ihor-Karpyn, line 92: Hard solution. You can only iterate transorm.properties and delete keys

Ihor-Karpyn (approved):

Ihor-Karpyn:
And for this case switch will better

src/transformState.js
```diff
@@ -63,6 +63,37 @@
  */
 function transformState(state, transforms) {
   // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
aliceblunt8, line 66: let's use 'for of' loop

src/transformState.js
```diff
@@ -63,6 +63,37 @@
...
+    const obj = transforms[i];
+
+    if (obj.operation === 'addProperties') {
```
aliceblunt8, line 69: let's use switch case instead of if

src/transformState.js
```diff
@@ -63,6 +63,37 @@
...
+      const properties = obj.properties;
+
+      for (const key in properties) {
+        state[key] = properties[key];
+      }
```
aliceblunt8, line 74: why not using Object.assign?

src/transformState.js
aliceblunt8: just iterate transform properties, get rid of state loop

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #487, opened 2021-04-22
solution

IvanKolodiazhnyi:


src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === 'addProperties') {
+      Object.assign(state, transforms[i].properties);
+    } else if (transforms[i].operation === 'removeProperties') {
+      for (let y = 0; y < transforms[i].properties.length; y++) {
+        delete state[transforms[i].properties[y]];
```
victor-arutiunov, line 70: You can use loop "for (let i of object)"

src/transformState.js
```diff
@@ -62,7 +62,21 @@
...
+      }
+    } else if (transforms[i].operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
+  return state;
```
victor-arutiunov, line 79: Good job! Also you can use switch case instead of if else.

src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
aliceblunt8, line 65: use 'for of' loop

src/transformState.js
```diff
@@ -62,7 +62,21 @@
...
+    if (transforms[i].operation === 'addProperties') {
```
aliceblunt8, line 66: let's use switch case instead of if

aliceblunt8 (changes requested):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const value of transforms) {
```
aliceblunt8, line 65:
```suggestion
  for (const transform of transforms) {
```

aliceblunt8 (approved):

PR #488, opened 2021-04-22
-solution

vitalya2626:


src/transformState.js
```diff
@@ -62,7 +62,28 @@
...
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+
+      default:
+        break;
```
IvanGrekov, line 84: Good and pretty laconic solution, but this part of code is unnecessary because `case 'clear'` is the last option and  we don't have to do something with our object by default.

IvanGrekov (changes requested):

src/transformState.js
vitalya2626: Thanks! Did this 'На Автомате') Пофиксил)

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+        for (const property of transform.properties) {
+          delete state[property];
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
```
IvanGrekov, line 81: `break` is still here, but ok

IvanGrekov (approved):

VinnieJ-2k20 (changes requested):
Actually, `switch/case` should always have default option in case of some irregular inputs.
You can simply return null by default.

VinnieJ-2k20 (approved):

PR #492, opened 2021-04-22
solution

provolovskyi-illia:


nazarmatsevych (approved):
Well done :)

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (transform.operation) {
+      case 'addProperties':
+        Object.assign(state, transform.properties);
+        break;
+      case 'removeProperties':
```
VinnieJ-2k20, line 70:
```suggestion
        break;
        
      case 'removeProperties':
```

VinnieJ-2k20 (changes requested):
`switch/case` block should always have default option in case of irregular inputs.
In this task you can simply perform `continue` by default.

src/transformState.js
```diff
@@ -62,7 +62,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
VinnieJ-2k20, line 84: For the future, in case of irregular inputs you may throw an error.
You'll learn how to do it further in the program.

VinnieJ-2k20 (approved):

PR #493, opened 2021-04-22
Solution 

SkyPKWolf:


src/transformState.js
```diff
@@ -62,7 +62,26 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
+      Object.assign(state, transform.properties);
+      continue;
```
vitaliy-buchinchik, line 68: I think that it`s unnecessary

src/transformState.js
```diff
@@ -62,7 +62,26 @@
...
+    }
+
+    if (transform.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+      continue;
```
vitaliy-buchinchik, line 75: Also this continue You can remote

src/transformState.js
```diff
@@ -62,7 +62,26 @@
...
+    }
+
+    if (transform.operation === 'removeProperties') {
+      for (const key of transform.properties) {
+        delete state[key];
+      }
+      continue;
```
vitaliy-buchinchik, line 82: One more continue :)

src/transformState.js
```diff
@@ -62,7 +62,26 @@
...
+    }
+  }
```
vitaliy-buchinchik, line 84: It would be better to return state object from function

vitaliy-buchinchik (approved):

aliceblunt8 (approved):

PR #494, opened 2021-04-22
Solution

IvanVasylenko:


src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (transform.operation === 'removeProperties') {
+      for (const prop of transform['properties']) {
```
SkyPKWolf, line 73:
```suggestion
      for (const prop of transform.properties) {
```

SkyPKWolf (approved):
good

VinnieJ-2k20 (changes requested):
Let's rewqrite the solution using `switch/case` structure.

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    switch (transform.operation) {
+      case 'removeProperties':
+        for (const prop of transform.properties) {
```
varseniuk, line 68:
```suggestion
        for (const property of transform.properties) {
```
`prop` is a not very meaningful name

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+        for (const prop in transform.properties) {
+          state[prop] = transform.properties[prop];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
varseniuk, line 82: Add `default` case

varseniuk (approved):
Good! 👍 
But you should always add `default` case for your `switch` statements

PR #496, opened 2021-04-22
solution

amithos:


progmari (approved):
Great job 👍

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const input of transforms) {
+    const operation = input.operation;
+    const props = input.properties;
```
aionova, line 67: You can do destructuring of input

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const input of transforms) {
```
```suggestion
  for (const { operation, properties } of transforms) {
```

aionova (changes requested):
Let's use switch case here

src/transformState.js
amithos: No, I can't. With **const {operation, props} = input;** the array becomes not iterable,
I don't know any other way to do this.

aionova (approved):
LGTM!

PR #497, opened 2021-04-22
Solution

progmari:


amithos (approved):
Well done!))

VinnieJ-2k20 (changes requested):
Good that you are already using array methods like `forEach` 👍 

VinnieJ-2k20 (changes requested):
`switch/case` structure should always have default option for irregular inputs.
For now it can do nothing, but usually it returns an `error.`

VinnieJ-2k20 (approved):

PR #498, opened 2021-04-22
Solution

leonid-vegera:


ivan-shpynda (approved):
well done!

VinnieJ-2k20 (changes requested):
`switch/case` structure should always have default option for irregular inputs.
For now it can do nothing, but usually it returns an error.

VinnieJ-2k20 (approved):
Good job 👍

PR #499, opened 2021-04-22
Solution

TimurHrynenko:


src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  transforms.forEach(move => {
+    switch (move.operation) {
+      case 'addProperties':
+        Object.assign(state, move.properties);
```
BoikoEva, line 68: Do you really need forEach() here? I think that 'for' loop is easier, and you will need less code, without something like 'move.properties);'

BoikoEva (approved):

src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+        break;
+      case 'removeProperties':
+        for (const prop in move.properties) {
+          if (state.hasOwnProperty(move.properties[prop])) {
```
aionova, line 72: this check is redundant

src/transformState.js
```diff
@@ -62,7 +62,27 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  transforms.forEach(move => {
```
```suggestion
  transforms.forEach(transform => {
```

aionova (approved):
Good job overall!

PR #502, opened 2021-04-22
solution

KapToH337:


BoikoEva (approved):
nice! anyway, you can also try switch case

src/transformState.js
```diff
@@ -62,7 +62,30 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    const operation = transform.operation;
+    const properties = transform.properties;
```
aionova, line 67:
```suggestion
  for (const { operation, properties } of transforms) {
```

aionova: You can use destructuring here

aionova (approved):
Switch case construction would fit perfect here

PR #503, opened 2021-04-22
solution

ivan-shpynda:


leonid-vegera (approved):
Good job!

src/transformState.js
```diff
@@ -61,8 +61,25 @@
  * @param {Object} state
  * @param {Object[]} transforms
  */
+
 function transformState(state, transforms) {
-  // write code here
+  for (const obj of transforms) {
```
VinnieJ-2k20, line 66:
```suggestion
  for (const transform of transforms) {
```

src/transformState.js
```diff
@@ -61,8 +61,25 @@
...
+    if (obj['operation'] === 'addProperties') {
```
VinnieJ-2k20, line 67:
```suggestion
    if (obj.operation === 'addProperties') {
```

src/transformState.js
```diff
@@ -61,8 +61,25 @@
...
+      Object.assign(state, obj['properties']);
```
VinnieJ-2k20, line 68:
```suggestion
      Object.assign(state, obj.properties);
```

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (changes requested):
`switch/case` structure should always have default option in case of irregular inputs.
In this task it may do nothing, or throw an error.

VinnieJ-2k20 (approved):

PR #504, opened 2021-04-22
'Solution'

ViPiven:


BoikoEva (approved):
Good job, also you can try switch case for it.

VinnieJ-2k20 (changes requested):

ViPiven:
> Let's rewrite the solution using `switch/case` structure.

done

VinnieJ-2k20 (changes requested):
`switch/case` structure should always have default option in case of irregular inputs.
In this task it may do nothing, or throw an error.

VinnieJ-2k20 (approved):

PR #506, opened 2021-04-22
solution

bohdanklius:


src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  const newState = state;
+
+  for (let i = 0; i < transforms.length; i++) {
```
BoikoEva, line 67: you can use here 'for (const obj of transforms)'  and at all you can do all the task with switch case, it's more understandable and just easy:)

BoikoEva (approved):

src/transformState.js
aliceblunt8, line 67: use 'for of' loop

src/transformState.js
```diff
@@ -62,7 +62,25 @@
...
+    if (transforms[i].operation === 'addProperties') {
+      Object.assign(newState, transforms[i].properties);
+    }
+
+    if (transforms[i].operation === 'removeProperties') {
+      for (let x = 0; x < transforms[i].properties.length; x++) {
+        delete newState[transforms[i].properties[x]];
+      }
```
aliceblunt8, line 75: use 'for in' loop

src/transformState.js
```diff
@@ -62,7 +62,25 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  const newState = state;
+
+  for (let i = 0; i < transforms.length; i++) {
+    if (transforms[i].operation === 'addProperties') {
```
aliceblunt8, line 68: use switch case

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #508, opened 2021-04-22
solution by ched

ched88xtc:


BoikoEva (approved):
Good job! Also you can try to use switch case

src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
+    if (transform.operation === 'addProperties') {
+      Object.assign(state, transform.properties);
+    } else if (transform.operation === 'removeProperties') {
+      for (const i of transform.properties) {
```
aionova, line 69:
```suggestion
      for (const property of transform.properties) {
```

src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const transform of transforms) {
```
```suggestion
  for (const { operation, properties } of transforms) {
```
it would be nice to use destructuring here

aionova (changes requested):
Instead of using 2 'else if' blocks, it would be easier to read switch case construction

aionova (approved):
Awesome!

PR #510, opened 2021-04-22
Yuhuuu!

Allrightman:


BoikoEva (approved):
Good, but i think it`s easier in switch case, try it:)

src/transformState.js
```diff
@@ -62,7 +62,21 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (const obj of transforms) {
```
```suggestion
  for (const { operation, properties } of transforms) {
```
let's destructure it

src/transformState.js
```diff
@@ -62,7 +62,21 @@
...
+    if (obj.operation === 'addProperties') {
+      Object.assign(state, obj.properties);
+    } else if (obj.operation === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    } else if (obj.operation === 'removeProperties') {
+      for (let i = 0; i < obj.properties.length; i++) {
```
aionova, line 73: for of loop would be better here

aionova (changes requested):
Let's use switch case. It fits perfect here

Allrightman:
Ok, thanks!

src/transformState.js
Allrightman: Sure I will change it, thanks!

ArtemTopolskyi (approved):
Good job!

Allrightman:
Thanks!

ArtemTopolskyi (changes requested):
fix

ArtemTopolskyi (approved):

PR #511, opened 2021-04-22
solution

dmitry-puhliakov:


BoikoEva (approved):
Great! Also you can use switch case as an alternative

VinnieJ-2k20 (changes requested):

VinnieJ-2k20 (changes requested):
`switch/case` should always have `default` option in case of irregular input.
In this task it can do nothing or throw an error.

varseniuk (approved):
Great 🔥

ArtemTopolskyi (changes requested):
fix

ArtemTopolskyi (approved):

PR #519, opened 2021-04-23
solution

Taya294:


Grygoriy-Shytikov (approved):
Good
Але думаю тут краще використати switch case щоб не повторювати "transform.operation === "

varseniuk (changes requested):
Good job! 🔥 Your logic is correct, but for the cases where you need to choose one option from the many, it's better to use `switch-case` statement instead of `if-then`.
Let's change it a bit and it's done
P.S: don't forget to add `default` case for your `switch-case`. You can throw an error or just `break` by default

varseniuk (approved):
Great 🔥

PR #523, opened 2021-04-23
ready for check

Bogdan-Kotsupey:


src/transformState.js
```diff
@@ -62,7 +62,36 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  for (let i = 0; i < transforms.length; i++) {
```
mykola-yarchuk, line 65: is better to use for of loop here, it's easier to read

src/transformState.js
```diff
@@ -62,7 +62,36 @@
...
+    for (const key in transforms[i]) {
+      if (transforms[i][key] === 'addProperties') {
+        continue;
```
mykola-yarchuk, line 68: that's fire, but there is a better approach, your solution use one loop and trying to handle different cases, which makes your code hard to read, would be better if you create different cases (switch-case construction), and run different loops(if it needed, depend on case)

src/transformState.js
```diff
@@ -62,7 +62,36 @@
...
+      }
+
+      if (Array.isArray(transforms[i][key]) === true) {
+        for (let k = 0; k < transforms[i][key].length; k++) {
+          delete state[transforms[i][key][k]];
+        }
+      }
+
+      if (transforms[i][key] === 'clear') {
+        for (const key3 in state) {
```
mykola-yarchuk, line 89: don't use numbers in variable names 
`key3` that's a bad name

mykola-yarchuk (changes requested):

mykola-yarchuk (approved):
great

PR #531, opened 2021-04-28
Solution

katya-suvorova:


src/transformState.js
```diff
@@ -62,7 +62,27 @@
...
+      for (const value of objectProperties) {
+        delete state[value];
+      }
+    }
+
+    if (object.operation === 'clear') {
+      for (const values in state) {
+        delete state[values];
+      }
+    }
```
vkryvytskyy, line 82: use switch..case instead of separate ifs

vkryvytskyy (changes requested):

vpolets (approved):

PR #548, opened 2021-06-09
My solution

pavlenkoam:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+    if (key.type === 'addProperties') {
+      Object.assign(state, key.extraData);
+    } else if (key.type === 'removeProperties') {
+      for (const keys of key.keysToRemove) {
+        delete state[keys];
+      }
+    } else if (key.type === 'clear') {
+      for (const k in state) {
+        delete state[k];
+      }
```
vkryvytskyy, line 18: use `switch..case` instead of `if`s to check key.type

vkryvytskyy (changes requested):

vpolets (approved):

PR #618, opened 2021-06-28
Solution

MariiaSpiridonova:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+        delete state[actions[i].keysToRemove[j]];
+      }
+    }
```
vkryvytskyy, line 25: use switch..case to do smth based on action.type

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const prop in state) {
+        if (state.hasOwnProperty(prop)) {
```
vkryvytskyy, line 11: this if is redundant

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[prop];
+        }
```
vkryvytskyy, line 13:
```suggestion

        delete state[prop];

```

vkryvytskyy (changes requested):

src/transformState.js
MariiaSpiridonova: Done. The solution was updated with suggested changes.

src/transformState.js
MariiaSpiridonova: Done. The solution was updated with suggested changes.

src/transformState.js
MariiaSpiridonova: Done. The solution was updated with suggested changes.

vkryvytskyy (approved):

PR #634, opened 2021-07-01
Solution

olenataranenko:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act in actions) {
```
vkryvytskyy, line 8: use for of loop with arrays

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (actions[act].type === 'removeProperties') {
+      for (let i = 0; i < actions[act].keysToRemove.length; i++) {
+        delete state[actions[act].keysToRemove[i]];
+      };
+    };
+
+    if (actions[act].type === 'clear') {
+      for (const j in state) {
+        delete state[j];
+      }
```
vkryvytskyy, line 22: use switch case there

vkryvytskyy (changes requested):

vkryvytskyy (approved):

PR #639, opened 2021-07-04
Transformer

HelenaMar:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const act in actions) {
```
vkryvytskyy, line 8: use for of loop

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      for (let i = 0; i < actions[act].keysToRemove.length; i++) {
+        delete state[actions[act].keysToRemove[i]];
+      };
+    };
 
+    if (actions[act].type === 'clear') {
+      for (const j in state) {
+        delete state[j];
+      }
+    }
```
vkryvytskyy, line 23: use switch case

vkryvytskyy (changes requested):

PR #652, opened 2021-07-17
transformState task

darynakravchuk:


src/transformState.js
```diff
@@ -6,6 +6,23 @@
...
+    if (keys.type === 'removeProperties') {
+      for (const i of keys.keysToRemove) {
+        delete state[i];
+      }
+      continue;
+    }
+
+    for (const j in state) {
+      delete state[j];
+    }
```
vkryvytskyy, line 24: consider to use switch case

src/transformState.js
```diff
@@ -6,6 +6,23 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const keys of actions) {
```
vkryvytskyy, line 9:
```suggestion
  for (const action of actions) {
```
you get single object from there

vkryvytskyy (approved):

PR #658, opened 2021-08-02
Solution

ivanchenko-kirill:


Moroz-Dmytro (approved):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        if (state[keyToRemove]) {
```
varseniuk, line 15:
```suggestion
```
not necessary here

varseniuk (changes requested):
All is good, but let's use `switch - case` construction for this purpose instead of `if`

varseniuk (approved):
Great!

PR #665, opened 2021-08-03
transformState was done

illiarerun:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        state[key] = action.extraData[key];
+      }
```
Anna-Tkachenko, line 12: You can also use Object.assign(state, action.extraData) here

illiarerun: Thanks for reminder, i'll use this method in next task))

danheim (commented):
Great, like it

danheim (approved):

PR #667, opened 2021-08-03
solution

DmytroMykoliv:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
sasha-morozov, line 8: Would suggest using the "for..in" loop instead of this one..Thus, instead of "actions[i]" you could use more readable variable. The hint is below:
Array => word + "s";
Object => word

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (actions[i].type === 'addProperties') {
```
sasha-morozov, line 9: Nice job. But I've heard that the mentors highly recommend using "switch" instead of "if's". So, I would suggest redoing this part of the task.

sasha-morozov (commented):
Good job!:)
Left some suggestions in the PR

src/transformState.js
DmytroMykoliv: Agree ))

src/transformState.js
DmytroMykoliv: agree, forgot about switch ))

sasha-morozov (approved):

Ihor-Karpyn (changes requested):
Variables name ⚠️

Ihor-Karpyn (approved):
LGTM 👌

PR #670, opened 2021-08-03
Solution

LazurNiko:


DmytroMykoliv (approved):
Great work ! Recommend you to use the switch method, it is will be more readable )

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
- */
+/*
+* @param {Object} state
+* @param {Object[]} actions
+*/
+
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
+    if (key.type === 'addProperties') {
```
Ihor-Karpyn, line 10: switch will better in this case

Ihor-Karpyn (changes requested):
Rewrite to `switch`

Ihor-Karpyn (approved):
LGTM 👌

PR #671, opened 2021-08-03
solved

lera-yaskevych:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
DmytroMykoliv, line 9: you can use here for ... of, it will be more readable

DmytroMykoliv (approved):
I did this using if's to, but I rewrote the task using the switch method as recommended, it looks more readable

DmytroMykoliv:
> Currently working on switch version
ok, don't  see this massage )

lera-yaskevych:
Thanks for the review! You gave me valuable remarks)

danheim (approved):

PR #672, opened 2021-08-03
Solution

denyschemeriakov:


varseniuk (changes requested):
All is good, but let's use `switch ... case` if we have many options to check

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
```
varseniuk, line 22: add `break` here as well

varseniuk (changes requested):
Add `break` to the last statement and add default case (just `break` in it)

varseniuk (approved):

PR #674, opened 2021-08-04
switch solution

zikozakazabiyaka:


UlaDutchak (approved):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Ihor-Karpyn, line 8: `for of`  will better here

Ihor-Karpyn (changes requested):
Almost done👌

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const data of actions) {
```
Ihor-Karpyn, line 8:
```suggestion
  for (const action of actions) {
```

Ihor-Karpyn (approved):
Great, pay attention to my comment

PR #675, opened 2021-08-04
Solution

poliandiya:


george-w-boo (commented):
Все супер

src/transformState.js
```diff
@@ -6,6 +6,24 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (let i = 0; i < actions.length; i++) {
```
Ihor-Karpyn, line 10: `for of` and `switch` will better in this case

Ihor-Karpyn (changes requested):
Almost done

Ihor-Karpyn (approved):
Great 👌

PR #678, opened 2021-08-04
solution

derelicttt:


UlaDutchak (approved):

varseniuk (changes requested):
All is good, but add `default` case

varseniuk (changes requested):
I see no changes after my last comment.
Add `default` case to your `switch...case` construction

varseniuk (approved):
🔥

varseniuk (approved):

PR #679, opened 2021-08-04
JS Advanced - Stateful object

george-w-boo:


UlaDutchak (approved):
Можна ще зробити через switch case

Ihor-Karpyn (changes requested):
`Switch` will better in this case, rewrite to `switch`

george-w-boo:
> `Switch` will better in this case, rewrite to `switch`

Переписав на switch

Ihor-Karpyn (approved):
LGTM

PR #680, opened 2021-08-04
js_stateful-object

RomanFedusevych:


UlaDutchak (approved):
Можна також використати switch case

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
```
varseniuk:
```suggestion
```
No need in this loop. Use Object.assign without loop

varseniuk (changes requested):
Let's use `switch...case` construction when we have many different options (don't forget about `default` case)

src/transformState.js
RomanFedusevych: поміняв хвилину тому

varseniuk (changes requested):
You forgot about `default` case

RomanFedusevych:
додав👌🏻

varseniuk (approved):

PR #681, opened 2021-08-04
transform state

set1one:


alex-glotov67 (approved):
Good solution, but after all cases, I prefer to add `default: break;`

varseniuk (changes requested):
All is good, but don't forget to add `default` case

varseniuk (approved):

PR #692, opened 2021-08-04
solution

roman-hado:


varseniuk (changes requested):
Let's use `switch...case` when we have many different options (don't forget about `default` case)

varseniuk (approved):
Good!

PR #693, opened 2021-08-04
Solution

TarasHoliuk:


alex-glotov67 (approved):
Looks cool, but for delete some property you can use only word delete

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          Reflect.deleteProperty(state, key);
```
varseniuk, line 15:
```suggestion
```
just use `delete`

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        }
+        break;
+      case 'clear':
+        for (const parameter in state) {
+          Reflect.deleteProperty(state, parameter);
+        }
+        break;
+    }
+  }
```
varseniuk, line 24: add `default` case

varseniuk (changes requested):

varseniuk (approved):
Good!

PR #696, opened 2021-08-04
Solution

Ingvvvar:


VVHimiak (approved):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
varseniuk, line 26: Add `default` case

varseniuk (changes requested):
All is good, but add `default` case

src/transformState.js
Ingvvvar: default case added

varseniuk (approved):
🔥

PR #701, opened 2021-08-04
done stateful object task

Uran-web:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const property of actions) {
```
Ihor-Karpyn, line 8:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    for (const type in property) {
```
Ihor-Karpyn, line 9:
```suggestion
```
This loop is redundant

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      if (property[type] === 'addProperties') {
```
Ihor-Karpyn, line 10: `Switch` will better here, rewrite to `switch`.

Ihor-Karpyn (changes requested):
Almost done

Ihor-Karpyn (approved):

PR #706, opened 2021-08-05
Task done

Virass:


alex-glotov67 (approved):
Perfect solution!

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
```
danheim, line 8: apples - apple
books - book
users - user
tables - table
actions - ?

```suggestion
  for (const key of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+          delete state[i];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, key.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const arr of key.keysToRemove) {
```
danheim, line 21: this is not `arr`, this is key

danheim (changes requested):
Gr

danheim (approved):

PR #737, opened 2021-09-14
solution

VladSaen:


oksana-onopa (approved):
LGTM 👍

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const keys in actions[i].extraData) {
```
Anna-Tkachenko, line 10: Use switch/case instead of if

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const keys in actions[i].extraData) {
```
Anna-Tkachenko, line 11:
```suggestion
        for (const key in actions[i].extraData) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          state[keys] = actions[i].extraData[keys];
+        }
+        break;
+
+      case 'removeProperties':
+        for (let y = 0; y < actions[i].keysToRemove.length; y++) {
+          delete state[actions[i].keysToRemove[y]];
+        }
```
Anna-Tkachenko, line 19: You can use for of here
```suggestion
        for (const key of actions[i].keysToRemove) {
          delete state[key];
        }
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+
+      case 'clear':
+        for (const keys in state) {
```
Anna-Tkachenko, line 23:
```suggestion
        for (const key in state) {
```

Anna-Tkachenko (approved):
Good job👍

PR #748, opened 2021-09-15
Solution transformState

EnCRT:


Denis-Demyanok (approved):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
Ihor-Karpyn, line 9: `switch` will beter here

Ihor-Karpyn (approved):

PR #766, opened 2021-09-16
solution

Tarasulko:


just-a-vlad (approved):
Good job)

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let modifiedState = Object.assign(state);
```
Anna-Tkachenko, line 8: This assignment is redundant here, you don't create new object, just copy link from state to modifiedState

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        modifiedState = Object.assign(modifiedState, action.extraData);
```
Anna-Tkachenko, line 13: You can change here state directly
```suggestion
        Object.assign(state, action.extraData);
```

Anna-Tkachenko (changes requested):

aliceblunt8 (approved):

PR #767, opened 2021-09-16
solution

Bogdansgit:


Anna-Tkachenko (changes requested):
Rewrite code logic using switch/case

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case (action.type === 'clear') :
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
Anna-Tkachenko, line 24:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;

      case 'clear':
        for (const key in state) {
          delete state[key];
        }
    }
```

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
Good job!👍

PR #780, opened 2021-10-05
Task solved

cyb1ove:


src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+      default:
+        break;
```
Olexandr6, line 22: I would add here some return string, for catching errors, in case you get different type.🤔
```suggestion
      default:
      	return 'Undefined action type: ' action.type; 
        break;
```

G-YaSha (approved):
Good solution👍

src/transformState.js
VinnieJ-2k20: @Olexandr6 
Great comment and error handling  👍

VinnieJ-2k20 (approved):
Good job 👍

PR #781, opened 2021-10-05
Task was solved

Igor-MateAcademy:


src/transformState.js
```diff
@@ -5,7 +5,50 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
TRSKZK, line 8: You can give  key variable more clear name I would say action in actions

src/transformState.js
```diff
@@ -5,7 +5,50 @@
...
+    const operationTypeObject = { ...actions[key] };
+
+    for (const operationKey in operationTypeObject) {
+      switch (operationTypeObject[operationKey]) {
+        case 'addProperties': {
+          addProperties(state, operationTypeObject.extraData);
```
TRSKZK, line 14: You can destructure data that you need  from operatioTypeObject.extraData, it will make your code look neat and clear

src/transformState.js
```diff
@@ -5,7 +5,50 @@
...
+          break;
+        }
+
+        case 'removeProperties': {
+          removeProperties(state, operationTypeObject.keysToRemove);
```
TRSKZK, line 19: Same applies to this data, just destructure it

TRSKZK (changes requested):
Nice practice with separate functions for each test case,  but you can solve this task using just one loop instead of nesting loops. Need some refactoring

TRSKZK (approved):
Well done!

VinnieJ-2k20 (approved):
Good job 👍

PR #785, opened 2021-10-05
Solution

MariiaObraztsova:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const count in actions) {
+    switch (actions[count].type) {
+      case 'addProperties':
+        for (const key in actions[count].extraData) {
```
TRSKZK, line 11: then you can destructure extraData and keysToRemove from action

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          state[key] = actions[count].extraData[key];
```
TRSKZK, line 12: I guess you can add new properties with Object.asign()

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in actions[count].keysToRemove) {
```
TRSKZK, line 17: This one can be done easier, you can loop over destructured keysToRemove and delete state properties with assigned keys.

TRSKZK (changes requested):
Well Done Masha!!! Need some minor fixes)

src/transformState.js
MariiaObraztsova: Got what you meant, but now I'm wondering what the difference between these approaches in general and in this task. Will ask on q&a :)

TRSKZK (approved):
Well Done !

VinnieJ-2k20 (approved):
Good job 👍

PR #788, opened 2021-10-06
solution

HaqpaHR:


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+        break;
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
Olexandr6, line 29: You should add here some return string, for catching errors, in case you get different type.🤔
```suggestion
      default:
      	return 'Undefined action type: ' + action.type; 
```

VinnieJ-2k20 (approved):
Good job 👍

PR #792, opened 2021-10-06
transformState

misharosa:


MariiaObraztsova (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+      case 'removeProperties':
+        for (const remove of key.keysToRemove) {
+          delete state[remove];
+        }
+        break;
+
+      case 'clear':
+        for (const char in state) {
+          delete state[char];
+        }
```
cyb1ove, line 26: Each of case block should ends with break statement.

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+    }
+  }
```
cyb1ove, line 28: Switch should ends width default case block.

VinnieJ-2k20 (changes requested):
Fix @cyb1ove comments.

PR #794, opened 2021-10-07
solution

ivarovskiy:


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+        break;
+
+      case 'clear':
+        for (const clearProp in state) {
+          delete state[clearProp];
+        }
+        break;
+
+      default:
+    }
```
cyb1ove, line 30: The `default` block should ending with a `break` statement too.

cyb1ove (approved):

PR #796, opened 2021-10-07
solution

Aleksey060790:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+  for (const key of actions) {
+    switch (key.type) {
+      case 'addProperties':
+        for (const add in key.extraData) {
+          state[add] = key.extraData[add];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const remove of key.keysToRemove) {
```
dima-narozhnyi, line 17: It is better to just use the word 'key' in the name of the variable

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+
+      default:
```
dima-narozhnyi, line 28: If no case is matched then the default code is executed (if it exists)!
You have it in your code, but it does nothing.

PR #797, opened 2021-10-07
solution

Linetstg:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+  for (const key of actions) {
+    switch (key.type) {
+      case 'addProperties':
+        for (const add in key.extraData) {
+          state[add] = key.extraData[add];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const remove of key.keysToRemove) {
```
dima-narozhnyi, line 17: It is better to just use the word 'key' in the name of the variable

dima-narozhnyi (approved):
The solution looks good! Great job!

PR #812, opened 2021-10-14
Initial commit - solution

paralele-piped:
Thanks for a review!

Note: final version with <code> else { continue; } </code> works 1s faster (to 5.5s from 6.3s)

Artem Motalov / GetLucky23

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -6,18 +6,25 @@
  */
 function transformState(state, actions) {
   for (const action of actions) {
-    if (action.type === 'addProperties') {
-      Object.assign(state, action.extraData);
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const keyRemoved of action.keysToRemove) {
+        delete state[keyRemoved];
+      }
+    } else if (action.type === 'clear') {
+      for (const keyCleared in state) {
+        delete state[keyCleared];
+      }
+    } else {
+      continue;
+    }
```
vkryvytskyy, line 21: Перебирай action.type через switch case

PR #833, opened 2021-11-08
First solution

WorkAccCom:


dariia-romanova (approved):
Everything looks good

src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```

AlinaMatuschak (changes requested):
Good start 👌

src/transformState.js
```diff
@@ -4,8 +4,25 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
AlinaMatuschak, line 9:
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
```

AlinaMatuschak (approved):

PR #834, opened 2021-11-08
sateful object task solution

JamesSunderlanddd:


vlad-tuzenko (approved):
GJ 👍

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const props of obj.keysToRemove) {
+          delete state[props];
+        }
+      }
+
+      if (obj[key] === 'clear') {
+        for (const keys in state) {
+          delete state[keys];
+        }
+      }
```
illiarerun, line 24: What if first condition turns out true? Computer would do 2 unnecessary checks after that. Switch/case would be a perfect fit here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    }
+  }
+
+  return state;
```
illiarerun, line 28: In this task there is no need for returning value at the end

illiarerun (changes requested):
Good start 🔥

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    for (const key in action) {
```
illiarerun, line 9: That's a useless loop, that makes the complexity of your solution not that great.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+      switch (true) {
+        case (action[key] === 'addProperties'): {
```
illiarerun, line 11: You can just check if `action.type === <someType>`

illiarerun (changes requested):
Good start 🔥

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+            delete state[props];
+          }
+          break;
+        }
+
+        case (action[key] === 'clear'): {
+          for (const keys in state) {
+            delete state[keys];
+          }
+          break;
```
illiarerun, line 27: Redundant break statement, you have no cases after that one

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    for (const key in action) {
+      switch (true) {
```
illiarerun, line 10: Also, i didn't notice, you can just pass action.type as the value for switch

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          break;
+        }
+
+        case ('clear'): {
+          for (const keys in state) {
+            delete state[keys];
+          }
+        }
+      }
+    }
```
illiarerun, line 28: Use 2 spaces indentation, not tabs

illiarerun (changes requested):

src/transformState.js
JamesSunderlanddd: @illiarerun thank you for that, would be impossible to become a programmer without that stuff.

illiarerun (approved):
LGTM 🔥

src/transformState.js
illiarerun: ))

PR #835, opened 2021-11-08
solution

mariia-bila:


dmtrhrytsak (approved):
Pretty awesome! Would leave as it is, except I would probably replace **for of loop's** in two places:

```
      for (const removeKeys of element.keysToRemove) {
        delete state[removeKeys];
      }
```

and 

```
      for (const key in state) {
        delete state[key];
      }
```

with [for each](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach) construction. That way, I think, it would be more clear and readable. But that's completely up to, since the code is already fire🔥

src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+    } else if (element.type === 'removeProperties') {
+      for (const removeKeys of element.keysToRemove) {
+        delete state[removeKeys];
+      }
+    } else if (element.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
illiarerun, line 20: You have a lot of nesting `if`'s here, it'd be better to use switch/case construction here

src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const element of actions) {
```
illiarerun, line 8: Okay, but what is the element? If you iterate through the `actions` array, then the value in it will be `action`

illiarerun (changes requested):
Good start 🔥  Now lets work on fixes))

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const removeKeys of action.keysToRemove) {
+          delete state[removeKeys];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
illiarerun, line 24: There is no need in `break` statement in the last case

illiarerun (approved):
Good job 🔥

illiarerun (approved):
LGTM 🔥

PR #836, opened 2021-11-08
Solution

msaikun:


volodymyr-soltys97 (approved):
Very well 👍

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
```suggestion
  for (const action of actions) {
    const {type, extraData, keysToRemove} = action;
   
```

AlinaMatuschak (changes requested):
Good start 👌

AlinaMatuschak (approved):

PR #837, opened 2021-11-08
solution

Val-space:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'addProperties':
+        Object.assign(state, obj.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of obj.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
+        }
```
dmtrhrytsak, line 19: You used the forEach construction below, then why not to use it here?) It's more clear and readable, and moreover, it will make the code look more flat, and not like this (with ступенями от лестници)

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
dmtrhrytsak, line 8: Use more descriptive name then obj, com'on Lera, what a naming is that!👀

dmtrhrytsak (approved):

src/transformState.js
illiarerun: That's not that scary to use just common loops instead of array methods))

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of obj.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(prop => delete state[prop]);
+        break;
```
illiarerun, line 24: There's actually no need in this break, cause you don't have conditions after this one

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        for (const key of obj.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(prop => delete state[prop]);
```
illiarerun, line 23: If you iterate through keys, it'd be better to name the element like `key`

src/transformState.js
Val-space: what's wrong with the name? We iterate through objects in actions. What can be more clear?

dmtrhrytsak: We are humans, and as a human It makes more sense to us when we say we iterate over **single** action in actions 🙌

Be more abstract, for your own sake!

illiarerun (approved):
Great job 🔥

PR #838, opened 2021-11-08
implemented task

tmrsl:


JamesSunderlanddd (approved):
Good Job!

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
    const {type, extraData, keysToRemove} = action;
   
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (obj.type === 'addProperties') {
```

AlinaMatuschak (changes requested):
Almost perfect 👍

AlinaMatuschak (approved):

PR #840, opened 2021-11-09
solution: Transforming state task

lanebx:


TatianaMorozova04:
I think it will be more convenient to read the code with more understandable variables.

VladDyrenkov:
Good job
Use more understandable variables =)

VladDyrenkov (approved):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'clear') {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      for (const keyInState in state) {
+        delete state[keyInState];
+      }
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const keyToRemove of obj.keysToRemove) {
+        for (const keyInState in state) {
```
AlinaMatuschak, line 17: When you write cycle in cycle think one more time: `is it necessary?`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          if (keyInState === keyToRemove) {
+            delete state[keyInState];
```
AlinaMatuschak, line 19: you can keyToRemove instead keyInState. If keyToRemove exists in the state, it will remove. If not - not

AlinaMatuschak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const { type, keysToRemove, extraData } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    switch (obj.type) {
+      case 'removeProperties':
+        for (const keyInState in state) {
+          if (obj.keysToRemove.includes(keyInState)) {
+            delete state[keyInState];
+          }
+        }
```
AlinaMatuschak, line 15:
```suggestion
        for (const keyToRemove of obj.keysToRemove) {
          delete state[keyToRemove];
        }
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      case 'clear':
+        for (const keyInState in state) {
+          delete state[keyInState];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, obj.extraData);
+        break;
```
AlinaMatuschak, line 26: for the last case, you can delete the break

AlinaMatuschak (approved):

AlinaMatuschak (approved):

PR #841, opened 2021-11-09
Solution

volodymyr-soltys97:


msaikun (approved):

msaikun (approved):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const value of actions) {
```
```suggestion
  for (const action of actions) {
    const {type, extraData, keysToRemove} = action;
   
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (value.type === 'addProperties') {
```

AlinaMatuschak (changes requested):
Good start 👌

AlinaMatuschak (approved):

PR #842, opened 2021-11-09
transform state solution

Vitaliy-0:


tmrsl (approved):

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
  
```

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
AlinaMatuschak, line 13: you can use Object assign

AlinaMatuschak (approved):
Good work 💪

PR #844, opened 2021-11-09
Solution stateful-object

Dima-Ziakhor:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (actions[i].type === 'addProperties') {
+      for (const action in actions[i].extraData) {
+        state[action] = actions[i].extraData[action];
+      }
```
alex-kostomarov, line 18: i think you can use "Object.assign(state, action.extraData)" instead of loop

alex-kostomarov (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+    if (actions[i].type === 'addProperties') {
+      for (const action in actions[i].extraData) {
+        state[action] = actions[i].extraData[action];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in state) {
+        for (const elem of actions[i].keysToRemove) {
```
AlinaMatuschak, line 23: When you write cycle in cycle think one more time: `is it necessary?`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          if (elem === key) {
+            delete state[key];
```
AlinaMatuschak, line 25: you can elem instead key. If elem exists in the state, it will remove. If not - not

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
AlinaMatuschak, line 8: better use for of with destructuring

AlinaMatuschak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear': {
+        if (action.type === 'clear') {
```
AlinaMatuschak, line 11: Read more about the switch case. If case 'clear' start to run, your action.type always will be equal 'clear'

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        Object.assign(state, action.extraData);
+
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const elem of action.keysToRemove) {
+          if (state.hasOwnProperty(elem)) {
+            delete state[elem];
+          }
```
AlinaMatuschak, line 30: you can delete the condition. If elem exists in the state, it will remove. If not - not

```suggestion
          delete state[elem];
```

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
```suggestion
  for (const { type, extraData, keysToRemove } of actions) {
```

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🤘

PR #846, opened 2021-11-09
init

junglq:


lera-permiakova (approved):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
+    if (key.type === 'addProperties') {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
```
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = action;
  
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (key.type === 'addProperties') {
+      for (const add in key.extraData) {
+        state[add] = key.extraData[add];
+      }
```
AlinaMatuschak, line 12: you can use object assign

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🚀

PR #847, opened 2021-11-09
First solution

ImIkator:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
-  // write code here
+  for (const direct of actions) {
+    switch (direct.type) {
+      case 'addProperties':
+        Object.assign(state, direct.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of direct.keysToRemove) {
+          if (key in state) {
```
sviatoslav-myhal, line 16: i don`t think , this condition makes some sens, mb remove it

sviatoslav-myhal (commented):
Good job!

src/transformState.js
illiarerun: Actually it makes sense, he checks if key from action.keysToRemove is in the `state`, so an unnecessary delete statement won't happen if there is no such key

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const direct of actions) {
```
illiarerun, line 8: action of actions

illiarerun (approved):
Good job 🔥

PR #849, opened 2021-11-09
transformState

TatianaMorozova04:


lanebx (approved):
Good work

src/transformState.js
```diff
@@ -4,8 +4,35 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const obj in actions) {
```
illiarerun, line 9: If you iterate through an array of `actions`, the element should be named `action`, right?

src/transformState.js
```diff
@@ -4,8 +4,35 @@
...
+      for (const key in keys) {
+        delete state[keys[key]];
+      }
+    }
+
+    if (type === 'clear') {
+      for (const prop in state) {
+        delete state[prop];
+      }
+    }
```
illiarerun, line 32: Too many if's, use switch/case here

illiarerun (changes requested):
Good start 🔥  But now let's fix some things:

src/transformState.js
```diff
@@ -4,8 +4,40 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
illiarerun, line 9: Actions is an array, you can iterate through, using for...of loop, in that way you wouldn't need to use indexes to find the right action in your loop

illiarerun (changes requested):

illiarerun (approved):
LGTM 🔥

PR #851, opened 2021-11-09
solution

oleksiy-pochkayev:


src/transformState.js
```diff
@@ -4,8 +4,23 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
SerhiiKirik, line 10: You can use `switch` here
```suggestion
   for (const action of actions) {
    switch (action.type) {
    case 'addProperties':
```

src/transformState.js
```diff
@@ -4,8 +4,23 @@
...
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
```
SerhiiKirik, line 18: You can just return empty object

SerhiiKirik (approved):
Good joob 🔥

AlinaMatuschak (changes requested):
fix @SerhiiKirik comments 👆

AlinaMatuschak (approved):
💪

PR #856, opened 2021-11-09
Solution

Veronika-Korona:


alinaShchavinska:
In my opinion, the task is done well

alinaShchavinska (approved):

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
illiarerun, line 26: Redundant break statement, you don't have any cases after this one

illiarerun (changes requested):
Good start 🔥

src/transformState.js
alinaShchavinska: but if there will be added a new case statement in the future. There is a chance to forget to add `break` to the previous one😉

alinaShchavinska: It is like a “trailing” or “hanging” comma in object declaration. Makes it easier to add/remove/move around properties ("cases"), because all lines become alike.

Veronika-Korona: Agree with Alina. On my opinion, it is better to leave it that way.

illiarerun: Then you should one, remember about `default:`

illiarerun (changes requested):

illiarerun (approved):
Great job 🔥

PR #857, opened 2021-11-09
Solution

SerhiiKirik:


Veronika-Korona (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (true) {
```
illiarerun, line 9: This type of check is valid, but we can use switch in more convenient way, just pass action.type here

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
illiarerun, line 8: Also, be more descriptive with your namings, if we iterate through an array of actions the element would be an action, right?

illiarerun (changes requested):
Good start 🔥  Let's improve your code a bit:

illiarerun (approved):
LGTM 🔥

PR #858, opened 2021-11-09
transform state

lera-permiakova:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        state[add] = prop.extraData[add];
+      }
+    }
+
+    if (prop.type === 'removeProperties') {
+      for (const rem of prop.keysToRemove) {
+        for (const key in state) {
+          if (key === rem) {
+            delete state[key];
+          }
```
junglq, line 26: u can optimize this block write only

delete state[rem];

junglq (approved):

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const prop of actions) {
+    if (prop.type === 'clear') {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const prop of actions) {
```
```suggestion
  for (const prop of actions) {
    const { type, extraData, keysToRemove } = prop;
  
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    if (prop.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (prop.type === 'addProperties') {
+      for (const add in prop.extraData) {
+        state[add] = prop.extraData[add];
+      }
```
AlinaMatuschak, line 18: you can use object assign

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    }
+
+    if (prop.type === 'removeProperties') {
+      for (const rem of prop.keysToRemove) {
+        for (const key in state) {
```
AlinaMatuschak, line 23: When you write cycle in cycle think one more time: `is it necessary?`

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):
🔥

PR #859, opened 2021-11-09
solution

MagicBox15:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (true) {
+      case
+        obj.type === 'addProperties':
```
SerhiiKirik, line 11: You can write here `obj.type`, and in the following conditions it will not be necessary to specify it repeatedly.
And, can you rename obj?
If we iterate through an array of `actions` then the element of this array would be called `action`, right?

```suggestion
    switch (obj.type) {
      case 'addProperties':
```

SerhiiKirik (approved):
Well done 🔥

src/transformState.js
MagicBox15: very well) done!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case
+        'addProperties':
```
illiarerun, line 11: There's no need to put the value of the case on a newline, fix it

illiarerun (approved):
Good job 🔥  Make sure to fix this one:

PR #860, opened 2021-11-09
state and action solution

skachkov23:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 9: I think it will be better.
```suggestion
  for (const action of actions) {
    const {type, extraData, keysToRemove} = action;
    
    switch (type) {
```

volodymyr-soltys97 (approved):
GJ

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
```
AlinaMatuschak, line 23: 🔥

AlinaMatuschak (approved):

PR #861, opened 2021-11-09
solution

Nadiyahr:


src/transformState.js
```diff
@@ -4,8 +4,23 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(x => {
```
IrynaHerasymchuk, line 9: Would be better to name arguments not just with single letters, but words (for example 'action')

IrynaHerasymchuk (approved):
Great! Btw, another option for solving this task is switch case 😉

src/transformState.js
Nadiyahr: Thanks :) , it is really more understandable to read

src/transformState.js
```diff
@@ -4,8 +4,23 @@
...
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(item => delete state[item]);
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
+    }
```
illiarerun, line 20: So let's use switch case, as Iryna suggested

src/transformState.js
```diff
@@ -4,8 +4,23 @@
...
+
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(item => delete state[item]);
```
illiarerun, line 15: If you iterate through keys, let's use `key` as the name of an element, and not abstract `item`

illiarerun (changes requested):
Good job 🔥  Let's work on fixes now:

illiarerun (approved):
LGTM 🔥

PR #863, opened 2021-11-10
Solution

alex-kostomarov:


4qwerty (approved):
Good job🔥

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        break;
```
illiarerun, line 27: There is no need to put break statement in default case

illiarerun (approved):
Good job 🔥  Make sure to fix this one:

PR #866, opened 2021-11-10
Solution

aldmpr:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
```
sviatoslav-myhal, line 16: you can just assign an empty object instead of this loop

sviatoslav-myhal (approved):
Good :heart_eyes:

AlinaMatuschak (approved):
👌

PR #880, opened 2021-12-12
commit added

Oleksandr-Leshchenko:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      }
+
+      if (element[key] === 'addProperties') {
+        Object.assign(state, element.extraData);
+      }
+
+      if (element[key] === 'clear') {
+        for (const item in state) {
+          delete state[item];
+        }
```
MaksimusCRS, line 23: 2. [CODE STYLE]: Use `switch` statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[element.keysToRemove[i]];
+        }
+      }
+
+      if (element[key] === 'addProperties') {
+        Object.assign(state, element.extraData);
+      }
+
+      if (element[key] === 'clear') {
+        for (const item in state) {
```
MaksimusCRS, line 21: 1. [CODE STYLE]: don't use `for in` loop for iterating over array

MaksimusCRS (changes requested):

src/transformState.js
Oleksandr-Leshchenko: Стейт - это не массив

MaksimusCRS (approved):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const element of actions) {
+    for (const key in element) {
```
Denchepornyuk, line 9: It isn't a good idea to use loop `for` inside another loop `for`
You can remove the second loop

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      switch (element[key]) {
+        case 'removeProperties':
+          for (let i = 0; i < element.keysToRemove.length; i++) {
+            delete state[element.keysToRemove[i]];
+          }
```
Denchepornyuk, line 14: loop `for in` is better here

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #884, opened 2021-12-13
solution

samson-sim:


Niki-veb-dev (commented):
GJ

Olexandr6 (changes requested):
It's better to use `switch` statement here  or rewrite your solution without `else`, cause it looks more complex

Olexandr6 (approved):
Good solution 👍

But `default` block in `switch` statement is required, so you should add it. It can prevent some errors

PR #896, opened 2021-12-14
Solution

destroythemoon:


ArturErmolenko (approved):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        delete state[i];
+      }
+    }
+
+    if (properties.type === 'clear') {
+      for (const i in state) {
+        delete state[i];
+      }
+    }
+  }
```
Denchepornyuk, line 24: use switch case instead of three if conditions

Denchepornyuk (changes requested):

PR #899, opened 2021-12-14
solution

Valeria-Valuyskaya:


Volodymyr1806 (approved):

Olexandr6 (approved):
Good solution 👍

PR #933, opened 2021-12-27
transform state

Maria-Lushchik:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (type === 'removeProperties') {
+      for (const key of keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
aliceblunt8, line 24: let's use switch case, not if blocks

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #935, opened 2022-01-03
Solution 1

Ihorhavryliak:


src/transformState.js
```diff
@@ -6,6 +6,24 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (const ket in actions) {
```
aliceblunt8, line 10: let's use switch case

aliceblunt8, line 10: use for of loop
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -6,6 +6,24 @@
...
+    }
+
+    if (actions[ket].type === 'clear') {
+      for (const up in state) {
+        delete state[up];
+      }
+    }
+
+    if (actions[ket].type === 'removeProperties') {
+      for (const ho of actions[ket].keysToRemove) {
```
aliceblunt8, line 22: use more understandable names for variables

aliceblunt8 (changes requested):

PR #939, opened 2022-01-09
solution [Yuriy Shvorob]

djtyrf312:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
aliceblunt8, line 9: in such cases it's better to use switch case

aliceblunt8 (approved):

PR #961, opened 2022-01-20
Solution

mariiaStorozhyk:


pogrebnyakp (approved):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties' :
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
```
Anna-Tkachenko, line 15: Create more meaningful variable name instead of i

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          if (Object.hasOwnProperty.call(state, property)) {
+            delete state[property];
+          }
+        }
+        break;
```
Anna-Tkachenko, line 26: Check point 4 in checklist

Anna-Tkachenko (changes requested):

Anna-Tkachenko (changes requested):
Add default case to switch/case statement

Anna-Tkachenko (approved):

PR #987, opened 2022-01-30
Solution

KovalVladuslav:


Ihor-Karpyn (changes requested):
rewrite on `switch case`, when you have list of options `switch` better than `if`

PR #998, opened 2022-02-08
solution

SanyaBratashchuk:


aspurgilus (approved):
Почти как у меня))

Ihor-Karpyn (changes requested):
rewrite to `switch`, if you have list of options `switch` better

Ihor-Karpyn (approved):

PR #1025, opened 2022-02-20
Solution

7dimnik:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
```
aliceblunt8, line 9: use switch case

aliceblunt8 (changes requested):

src/transformState.js
7dimnik: Why is the switch case better?

aliceblunt8: cause when you have one value to compare it's more declarative to use switch case

aliceblunt8 (approved):

PR #1029, opened 2022-03-05
Solution

ghost:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const value of actions) {
+    if (value.type === 'addProperties') {
+      Object.assign(state, value['extraData']);
```
vitaliy-martynyuk, line 10:
```suggestion
      Object.assign(state, value.extraData);
```

src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const value of actions) {
+    if (value.type === 'addProperties') {
```
vitaliy-martynyuk, line 9: try using `switch` here instead of multiple `if`

vitaliy-martynyuk (changes requested):

vitaliy-martynyuk (approved):

PR #1033, opened 2022-03-10
solution

OlexandrIstomin:


src/transformState.js
```diff
@@ -4,8 +4,24 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function transformState(object, array) {
+  for (const i of array) {
+    for (const k in i) {
```
vitaliy-martynyuk, line 9: this loop is not necessary here

src/transformState.js
```diff
@@ -4,8 +4,24 @@
...
+      if (i[k] === 'addProperties') {
```
vitaliy-martynyuk, line 10: use `switch` here

vitaliy-martynyuk (changes requested):

vitaliy-martynyuk (approved):

PR #1046, opened 2022-03-28
Solution

yaremenko-maksym:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
```
OksanaTyshchenko, line 9: You can use: for (const item of actions), because you don't use index of each array item, only value of item

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+    switch (true) {
+      case actions[i].type === 'addProperties':
+        for (const key in actions[i].extraData) {
```
OksanaTyshchenko, line 12: try to use destructuring to receive keys of extraData

OksanaTyshchenko (changes requested):
Good job, but some things could be made simpler.

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const item of actions) {
```
Anna-Tkachenko, line 9:
```suggestion
  for (const action of actions) {
```

Anna-Tkachenko (approved):
Good job 👍

PR #1049, opened 2022-03-29
Initial solution

Max-Pochepets:


src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+function removeProperties(state, keys) {
+  for (const key of keys) {
+    delete state[key];
+  }
+}
+
+function clear(state) {
+  for (const key in state) {
+    delete state[key];
+  }
```
vkryvytskyy, line 38: Separate functions looks like overkill for this, but no strong opinion

vkryvytskyy (approved):

PR #1052, opened 2022-03-30
transformState function was created

OksanaTyshchenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const item of actions) {
+    if (item.type === 'addProperties') {
```
yaremenko-maksym, line 9: For me, `switch` statement is better in this task, instead of `if/else`.

yaremenko-maksym (changes requested):
Might use some minor changes.

Anna-Tkachenko (approved):
Check point 4 in checklist: don't forget to add default case

PR #1054, opened 2022-03-30
state was added

HelenPetrovska:


ivan-kadykalo (approved):
Looks good

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
```
Anna-Tkachenko, line 25: It's not default case, it's logic for 'clear' action.type

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,32 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
Anna-Tkachenko, line 10: 🔥

Anna-Tkachenko (approved):
Good job! 👍

PR #1055, opened 2022-03-30
Solution

strukk311:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
YegorKochetkov, line 11:
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
```
I think it will be better:
the expression inside switch (action.type) is compared with the condition in case. And if true - code will be executed, if false - break will be executed

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+          delete state[key];
+        }
+        break;
+
+      case action.type === 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
YegorKochetkov, line 28: 4.[CODE STYLE]: switch/case should always have default case for error handling.

YegorKochetkov (changes requested):

Anna-Tkachenko (approved):
Good job! 👍

PR #1056, opened 2022-03-30
solution

Jenia80:


strukk311 (approved):
good job)

Anna-Tkachenko (changes requested):
Good job! 👍 
Just try to rewrite this logic using switch / case
Such way will be better if you should do some logic depending on the value of some property

Anna-Tkachenko (approved):
Good job!
Don't forget to add default case for switch / case statement

PR #1057, opened 2022-03-30
Solution

Vakolyyk:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      for (const key of arr) {
+        delete state[key];
+      }
+    }
+
+    if (actions[i]['type'] === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VETALEK, line 26: u can use `switch` satement for `actions[i].type`

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
```
VETALEK, line 8: comment

VETALEK (approved):
All good

Anna-Tkachenko (approved):
Good job!
Just use switch / case if you need to do some logic depending on the value of some variable

PR #1058, opened 2022-03-30
TransformeState solution

vlasiuk-anatolii:


VladSkoromnyi (approved):
Good job

Anna-Tkachenko (approved):
Good job!
Just use switch / case if you need to do some logic depending on the value of some variable

PR #1059, opened 2022-03-31
Solution

YegorKochetkov:


HelenPetrovska (approved):
Hi) You could use switch, but it's your option. Good job!

Anna-Tkachenko (approved):
Good job!
Just use switch / case in cases where you should do some logic depending on the value of some variable

PR #1060, opened 2022-03-31
Solution

ElizaChernysh:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return state;
```
YegorKochetkov, line 27: according to the condition, it seems that it is not necessary to return anything from the function

YegorKochetkov (approved):
goog job)

Anna-Tkachenko (approved):
Good job! 👍

PR #1062, opened 2022-03-31
solution Transform State

maxxnikiforov:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i]['type']) {
+      case 'addProperties':
+        Object.assign(state, actions[i]['extraData']);
+        break;
```
YegorKochetkov, line 12: I think it would be clearer and more beautiful:
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+      case 'removeProperties':
+        for (let x = 0; x < actions[i]['keysToRemove'].length; x++) {
+          delete state[actions[i]['keysToRemove'][x]];
```
YegorKochetkov, line 16: and here:
```suggestion
        for (const key of action.keysToRemove) {
          delete state[key];
```

YegorKochetkov (approved):
looks good ;)

src/transformState.js
maxxnikiforov: Согласен, красивее и понятней - исправляю!

danylo-onishchuk (approved):

PR #1067, opened 2022-03-31
add task

Helexi:


kkiirill (approved):
great

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const a of actions) {
```
Anna-Tkachenko, line 8: What is a?

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    switch (a.type) {
+      case 'addProperties':
+        Object.assign(state, a.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const m of a.keysToRemove) {
```
Anna-Tkachenko, line 15: What is m?

Anna-Tkachenko (changes requested):

src/transformState.js
Helexi: This is item in array) But I undestand that I incorrectly named the variables

danylo-onishchuk (approved):

PR #1069, opened 2022-03-31
solution

OleksandrNi:


ElizaChernysh (approved):
Good job)

src/transformState.js
```diff
@@ -1,11 +1,25 @@
...
+    }
+
+    if (key.type === 'removeProperties') {
+      for (const rem of key.keysToRemove) {
+        delete state[rem];
+      }
+    }
+
+    if (key.type === 'clear') {
+      for (const clear in state) {
```
Anna-Tkachenko, line 16:
```suggestion
      for (const key in state) {
```

src/transformState.js
```diff
@@ -1,11 +1,25 @@
...
- */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
+    if (key.type === 'addProperties') {
+      Object.assign(state, key.extraData);
+    }
+
+    if (key.type === 'removeProperties') {
+      for (const rem of key.keysToRemove) {
```
Anna-Tkachenko, line 10: What is rem?

Anna-Tkachenko (changes requested):
Good job!
Just fix variable naming
And if you need to do some logic depending on the value of some variable - will be better to use switch / case

danylo-onishchuk (approved):

danylo-onishchuk (commented):
need default

PR #1070, opened 2022-03-31
solve

kkiirill:


Anastasiia-Sharkova (approved):

Anna-Tkachenko (approved):
Good job!
Don't forget to add default case to switch / case statement

PR #1071, opened 2022-03-31
state solution

Anastasiia-Sharkova:


kkiirill (approved):
great!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const extra of action.keysToRemove) {
+          delete state[extra];
+        };
+        break;
+
+      default:
+        for (const item in state) {
+          delete state[item];
+        }
+        break;
```
Anna-Tkachenko, line 24: It's not default case, it's case for 'clear' action.type

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'removeProperties':
+        for (const extra of action.keysToRemove) {
+          delete state[extra];
+        };
+        break;
+
+      case 'clear':
+        for (const item in state) {
```
Anna-Tkachenko, line 21:
```suggestion
        for (const key in state) {
```

Anna-Tkachenko (approved):
Good job! 👍

PR #1072, opened 2022-03-31
Finish task

SeKaM0:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties' :
+        Object.assign(state, actions[i].extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const key in state) {
+          if (actions[i].keysToRemove.includes(key)) {
+            delete state[key];
```
Anna-Tkachenko, line 17: This if is redundant here

Anna-Tkachenko (approved):
Don't forget to add default case for switch / case statement

PR #1077, opened 2022-03-31
Solution

AntonLevchuk:


src/transformState.js
```diff
@@ -4,8 +4,34 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
IlliaBezditnyi, line 9:
```suggestion
```
i think you can use for of loop here

IlliaBezditnyi (approved):
Looks good) Check comment

Anna-Tkachenko (approved):
Good job! 👍

PR #1078, opened 2022-04-01
ready

Nikitiliok:


SeKaM0 (approved):
Looks good

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+
+      case 'removeProperties':
+        for (let item = 0; item < actions[i].keysToRemove.length; item++) {
+          delete state[actions[i].keysToRemove[item]];
+        }
+        break;
+
+      case 'clear':
+        for (const item in state) {
```
Anna-Tkachenko, line 23:
```suggestion
        for (const key in state) {
```

Anna-Tkachenko (approved):
Don't forget to add default case to switch / case statement

PR #1091, opened 2022-04-06
Solution

AndriiRut:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function transformState(state, ...actions) {
```
Bohdan-Maksymiuk, line 7:
```suggestion
function transformState(state, actions) {
```
why do you need this?

Bohdan-Maksymiuk (changes requested):
So many loops and if's 😨 
You can just iterate through actions and then write logic in `switch case` dependent on action type

src/transformState.js
AndriiRut: I accidentally add it, and now after I deleted it my code go wrong.

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of actions[i].keysToRemove) {
+          if (!state[key]) {
+            continue;
+          }
+
+          delete state[key];
```
Denchepornyuk, line 22: do you need this `if`? I believe not

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1094, opened 2022-04-08
transform state task solution

koshachina:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+      case 'addProperties':
+        for (const key in action.extraData) {
+          newState[key] = action.extraData[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in newState) {
+          delete newState[key];
+        }
```
ArtemTopolskyi, line 22: or just newState = {};

ArtemTopolskyi (approved):

PR #1105, opened 2022-04-15
Solution

TanyaPanfyorova:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
Oleksandra-Pasenchenko, line 8:
```suggestion
  for (const action of actions) {
```

Oleksandra-Pasenchenko (approved):
Good job, but have a look at the comment: it's better to name variables that you iterate over as singular, like this:
action of actions
state of states

src/transformState.js
TanyaPanfyorova: Got it! Fixed

Oleksandra-Pasenchenko (approved):

PR #1107, opened 2022-04-18
Solution

oles-partyka:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const property of actions) {
```
```suggestion
  for (const action of actions) {
```

Bohdan-Maksymiuk: semantic naming

Bohdan-Maksymiuk (changes requested):
Good logic but fix naming

Bohdan-Maksymiuk (approved):
Great

PR #1110, opened 2022-04-20
Solution

59MrRobot:


src/transformState.js
```diff
@@ -6,6 +6,21 @@
...
+      Object.assign(state, act.extraData);
+    } else if (act.type === 'removeProperties') {
+      for (const i of act.keysToRemove) {
+        delete state[i];
+      }
+    } else {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
andreisucman, line 20: It appears to me that if the user enters a random command as a second argument your function will delete all the data. This is not cool. It should delete only if 'clear' is passed.

andreisucman (changes requested):
A clean solution, but it has a logical mistake IMO. Also if you remove the 'else' part of the loop you will be able to put blank spaces between the 'if' blocks which improves the readability IMO).

andreisucman (approved):
Nice and clean.

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+  for (const act of actions) {
+    const { type, extraData, keysToRemove } = act;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const i of keysToRemove) {
```
VinnieJ-2k20, line 18: Let's name variables appropriately
```suggestion
        for (const key of keysToRemove) {
```

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'action does no exist';
```
VinnieJ-2k20, line 30:
```suggestion
        return 'action does not exist';
```

VinnieJ-2k20 (approved):
Great job 🥇

PR #1111, opened 2022-04-20
Task completed

andreisucman:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
59MrRobot, line 8: I think here that using the for...in loop would make your code look more cleaner and easier to read.

59MrRobot (changes requested):
Good work, just one suggestion

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    if (actions[i]['type'] === 'addProperties') {
+      Object.assign(state, actions[i]['extraData']);
+    } else if (actions[i]['type'] === 'removeProperties') {
```
59MrRobot, line 11: Also maybe using only if statements as Ihor suggested in the calculate cost task?

andreisucman: Yes, will do.

src/transformState.js
andreisucman: Makes sense

59MrRobot (approved):
Great work!

src/transformState.js
ihor-jpeg: but keep in mind that `for of` is about 6 times slower than `for`. it's okay for small inputs, but can have a big impact with big inputs

VinnieJ-2k20 (changes requested):
Everything looks good.
But in this situation where there's limited number of options, it's better to use `switch/case` instead of `if`.
Also, `switch/case` should always have `default` case that may return error message or throw an error.

VinnieJ-2k20 (approved):
Good job 👍

PR #1114, opened 2022-04-21
solution

efdalyalcin:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const stateObject = state;
```
amitpatiljc, line 8: I think variable is not required for this task. You can simply use 'state' as a object to modify

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+  for (const key of actions) {
+    if (key.type === 'addProperties') {
```
amitpatiljc, line 11: This is clean code but as per checklist
2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

amitpatiljc (changes requested):
Good job Efdal , Checkout the comments for suggested changes.

src/transformState.js
efdalyalcin: it's a work of habit :) removed.

src/transformState.js
efdalyalcin: changed to switch case.

amitpatiljc (approved):
Great work!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
meforoma, line 9: >4. [CODE STYLE]: switch/case should always have default case for error handling.

meforoma (changes requested):
all's great - except missing `default case for error handling`

PR #1116, opened 2022-04-21
Solution

amitpatiljc:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty([key])) {
```
iLebo-stack, line 16: There's really no need to check if `state` has property key `key`, in this scenario your goal is to delete the property anywhere

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'error!';
```
iLebo-stack, line 29: After default you could just end with `break`. There's no need for the function to return anything

iLebo-stack (changes requested):
Great Amit👍, take a look at my comments

src/transformState.js
amitpatiljc: Done

src/transformState.js
amitpatiljc: Done

iLebo-stack (approved):
Nicely done

VinnieJ-2k20 (approved):
Good job 👍

PR #1117, opened 2022-04-22
solution

iLebo-stack:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
59MrRobot, line 8: You should destructure the parameters here in the loop declaration the same way we learnt today to optimize code

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    const instruction = action.type;
```
59MrRobot, line 9: After destructuring your code, you wont need this line

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (instruction === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
59MrRobot, line 25: According to the checklist, we should use a switch for this part, with a default case too

59MrRobot (changes requested):
Good job, just a couple of suggestions

59MrRobot (approved):
Good job

meforoma (approved):
🎉

PR #1118, opened 2022-04-23
Solution

vitalii-heiko:


src/transformState.js
```diff
@@ -6,6 +6,23 @@
...
+      Object.assign(state, actions[i].extraData);
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key of actions[i].keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
```
ArturErmolenko, line 20: Better use Switch Case

ArturErmolenko (changes requested):

ArturErmolenko (approved):
Well done 🔥

PR #1122, opened 2022-04-26
stateful object

elkhayate:


meforoma (approved):
neat, although when `throw Error` it's better to provide some relevant wording to let user know what's wrong.

PR #1124, opened 2022-04-26
too dificult, perhaps my solution not one of the simplest

ZUBOV-ILLIA:


EnjSadman (approved):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
Denchepornyuk, line 28: use `switch case` instead of `if`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      Object.assign(state, obj.extraData);
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key in state) {
+        for (const ch of obj.keysToRemove) {
+          if (key === ch) {
+            delete state[key];
+          }
+        }
```
Denchepornyuk, line 19: you don't need two loops here
make one loop for `obj.keysToRemove` and delete every key from `state` witch presented in `obj.keysToRemove`


also, you don't need `if`
If key doesn't exist in object and you are trying to remove this key, JS will skip it without any error
Screenshot 2022-04-26 at 20 48 35

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1127, opened 2022-04-26
solution

lytvyshko:


iwanttosmile4u (approved):

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
+ * @param {action[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
Denchepornyuk, line 13: use `Object.assign` instead of loop, for adding values to object

denys-cheporniuk (changes requested):

denys-cheporniuk (approved):

PR #1129, opened 2022-04-26
create transformState

Andriyk0:


AnastasiyaSopotnytska (approved):

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      Object.assign(state, key.extraData);
+    } else if (key.type === 'removeProperties') {
+      for (const j of key.keysToRemove) {
+        delete state[j];
+      }
+    } else if (key.type === 'clear') {
+      for (const k in state) {
+        delete state[k];
+      }
+    }
```
Denchepornyuk, line 19: use `switch case` instead of `if, else if, else`

Denchepornyuk (changes requested):

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1130, opened 2022-04-26
solution

vlad-demchuk:


iwanttosmile4u (approved):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
```
Denchepornyuk, line 8: `i` isn't a good name for the variable

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1131, opened 2022-04-26
solution

SavMary:


AnastasiyaSopotnytska (approved):
good👍

src/transformState.js
```diff
@@ -4,8 +4,20 @@
...
+      Object.assign(state, actions[key].extraData);
+    } else if (actions[key].type === 'removeProperties') {
+      for (const el of actions[key].keysToRemove) {
+        delete state[el];
+      }
+    } else {
+      for (const el in state) {
+        delete state[el];
+      }
+    }
```
Denchepornyuk, line 20: use `switch case` instead of `if, else if, else`

Denchepornyuk (changes requested):

PR #1133, opened 2022-04-27
Jstasks

MabuzaM:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const stateLink = state;
+
+  for (const action of actions) {
```
59MrRobot, line 8: You should destructure the object here

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      Object.assign(stateLink, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete stateLink[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const prop in stateLink) {
+        delete stateLink[prop];
+      }
+    }
```
59MrRobot, line 28: According to the checklist, you should use a switch statement for these conditions with a default case too

59MrRobot (changes requested):

src/transformState.js
MabuzaM: Noted thanx, Please check again.

59MrRobot (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
+
+      default:
+        throw new Error();
```
meforoma, line 27: better provide some descriptive message
e.g. 'action type is not supported'

meforoma (approved):
one tiny comment
but not blocking approval

PR #1134, opened 2022-04-27
First_commit_TransformState

igoree88:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
-  // write code here
+
+  for (const value of actions) {
+    if (value.type === 'addProperties') {
+      Object.assign(state, value.extraData);
+    }
+
+    if (value.type === 'removeProperties') {
+      for (const value2 of value.keysToRemove) {
+        if (state.hasOwnProperty(value2)) {
```
rostyslav-pyliak, line 16: It should work without this "if"

rostyslav-pyliak (approved):
Looks like everything OK

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+          delete state[value2];
+        }
+      }
+    }
+
+    if (value.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
```
Denchepornyuk, line 24: use `switch case` instead of `if`

src/transformState.js
Denchepornyuk, line 16: you can remove this `if`

Denchepornyuk (changes requested):

src/transformState.js
igoree88: Fixed.

Denchepornyuk (approved):

PR #1138, opened 2022-04-27
solution

valerii-derkach:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
+    if (object.type === 'removeProperties') {
```
iwanttosmile4u, line 9: Use switch instead of if.

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const values of object.keysToRemove) {
+        delete state[values];
+      }
+    }
+
+    if (object.type === 'addProperties') {
+      for (const pair in object.extraData) {
```
iwanttosmile4u, line 16: You can use Object.assign() instead of this loop and add obj.extraData into state.

Denchepornyuk (approved):

PR #1139, opened 2022-04-27
solution stateful object

kitaminary:


src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'removeProperties':
+        for (const key1 of obj.keysToRemove) {
+          delete state[key1];
+        }
+        break;
+
+      case 'addProperties':
+        for (const key2 in obj.extraData) {
```
iwanttosmile4u, line 18: You can use Object.assign() instead of this loop, where add obj.extraData into state object.

Denchepornyuk (approved):

PR #1141, opened 2022-04-27
Solution

PRO-GRAM-MER:
by mistake commited two files

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -6,6 +6,21 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action in actions) {
+    if (actions[action]['type'] === 'addProperties') {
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -1,11 +1,22 @@
 'use strict';
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
   // write code here
+  for (const action in actions) {
+    if (actions[action]['type'] === 'addProperties') {
```
meforoma, line 6: and same here
- de-structuring 
- `switch - case` statement

meforoma (changes requested):

VinnieJ-2k20 (approved):
Good job 👍

PR #1142, opened 2022-04-27
make stateful object

uliton:


valerii-derkach (approved):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const act of action.keysToRemove) {
+        delete state[act];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const del in state) {
+        delete state[del];
+      }
+    }
```
Denchepornyuk, line 23: use `switch case` instead of `if`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1144, opened 2022-04-27
solution made through suffering

DNK-forexample:


rostyslav-pyliak:
"solution made through suffering"

I understand you)

src/transformState.js
```diff
@@ -1,11 +1,27 @@
...
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const prop in state) {
+        delete state[prop];
+      }
```
Denchepornyuk, line 22: use `switch case` instead of `if`

Denchepornyuk (changes requested):

PR #1145, opened 2022-04-27
Solution

RomekIvantsiv:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        if (action.type === 'clear') {
```
JustDan17, line 23: Remove this condition. It's excessive. You already set such condition with case 'clear':.

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          for (const key in state) {
+            delete state[key];
+          }
+        }
+        break;
+    }
+  }
+
+  return state;
```
JustDan17, line 32: You are required to change the `state` object, but not to return anything. Delete this return statement.

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
JustDan17, line 16: Remove this condition. It's unnecessary. It does not change the result but makes code longer)

JustDan17 (changes requested):

Denchepornyuk (approved):

PR #1147, opened 2022-04-28
transform state - is done

enrikadej:


src/transformState.js
```diff
@@ -5,7 +5,22 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const object of actions) {
+    if (object.type === 'addProperties') {
```
dellachaise, line 9: Please use switch-case instead of if statements.

dellachaise (changes requested):
Good job!

Some comments:
please change if to switch-case.

Denchepornyuk (approved):

PR #1148, opened 2022-04-28
Stateful Object

dasha-chekmariova:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
JustDan17, line 9: Use `switch` construction instead of `if`. Change in all occurrences.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      const copyExtra = { ...obj.extraData };
+
+      for (const prop in copyExtra) {
+        state[prop] = copyExtra[prop];
+      }
```
JustDan17, line 14: This construction works, but use Object.assign(state, obj.extraData); instead. It's shorter, so it makes your code prettyer, more readable and more understandable.

JustDan17 (changes requested):

Denchepornyuk (approved):

PR #1149, opened 2022-04-28
Develop

WojciechCzarnowski:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+    }
```
59MrRobot, line 26: Maybe adding a default case to the switch statement as good practice?

59MrRobot (approved):
Good work, just one small suggestion though

src/transformState.js
```diff
@@ -6,6 +6,30 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const element of actions) {
```
VinnieJ-2k20, line 9:
```suggestion
  for (const action of actions) {
```

VinnieJ-2k20 (approved):
Good job 👍

PR #1151, opened 2022-04-28
Solution

rostyslav-pyliak:


DNK-forexample (approved):
everything is ok

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const removeKey of key.keysToRemove) {
+        delete state[removeKey];
+      }
+    }
+
+    if (key.type === 'clear') {
+      for (const deleteKey in state) {
+        delete state[deleteKey];
+      }
+    }
```
Denchepornyuk, line 23: use `switch case` instead of `if`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1155, opened 2022-04-28
Solved stateful object task

dellachaise:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    }
+
+    if (currentAction.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
+  return state;
```
anastasiia-nikita, line 32: In task didn`t say about return something, i think without this return code also will work

anastasiia-nikita (approved):
Good job)

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1156, opened 2022-04-28
Transform state

Alyona-Pisotska:


ArturErmolenko (changes requested):
Better use switch case

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
ArturErmolenko, line 26: add default case

ArturErmolenko (changes requested):

src/transformState.js
ArturErmolenko: oops,misclicked ,your code is correct 👍

ArturErmolenko (approved):

PR #1158, opened 2022-04-29
solution

vitaliy-kovbasinskyi:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case action.type === 'removeProperties':
+        for (const del of action.keysToRemove) {
```
Anna-Tkachenko, line 14: What is del? create more meaningful variable name

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+          delete state[del];
+        }
+        break;
+
+      case action.type === 'clear':
+        for (const del in state) {
```
Anna-Tkachenko, line 20: Same here

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+          delete state[del];
+        }
+        break;
```
Anna-Tkachenko, line 23: Don't forget about default case for switch / case statement

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1159, opened 2022-04-29
transform state

ISRLv2021:


Anna-Tkachenko (changes requested):
Rewrite your code using switch / case statement

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  if (actions.length === 0) {
+    return state;
+  }
+
+  for (const key of actions) {
```
Anna-Tkachenko, line 12: It's not key, it's action

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  if (actions.length === 0) {
+    return state;
+  }
```
Anna-Tkachenko, line 10: This check is redundant

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+  for (const key of actions) {
+    switch (key.type) {
+      case 'addProperties' :
+        Object.assign(state, key.extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const item of key.keysToRemove) {
```
Anna-Tkachenko, line 19: It's not item, it's key

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[item];
+        }
+        break;
+
+      case 'clear' :
+        for (const del in state) {
```
Anna-Tkachenko, line 25: What is del?

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[del];
+        }
+        break;
```
Anna-Tkachenko, line 28: Don't forget about default case for switch / case statement
You can add just simple break in it

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      case 'clear' :
+        for (const remove in state) {
+          delete state[remove];
+        }
+        break;
+      default :
+        Object.assign(state, action.extraData);
```
Anna-Tkachenko, line 26: Add just break here, you shouldn't do any actions for default case

Anna-Tkachenko (approved):

PR #1160, opened 2022-04-30
commit message

AndriySimyachenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const toDelete of action.keysToRemove) {
+          delete state[toDelete];
+        };
+        break;
+
+      case 'clear':
+        for (const entries in state) {
```
Anna-Tkachenko, line 21: It's not entries, it's key

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const toDelete of action.keysToRemove) {
```
Anna-Tkachenko, line 15: what is toDelete? it's also key

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[toDelete];
+        };
+        break;
+
+      case 'clear':
+        for (const entries in state) {
+          delete state[entries];
+        }
+        break;
```
Anna-Tkachenko, line 24: Don't forget about default case for switch / case statement

Anna-Tkachenko (approved):

PR #1162, opened 2022-05-01
My Solution

Yaroslav-Kononenko:


Anna-Tkachenko (changes requested):
Rewrite your code using switch / case statement

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Anna-Tkachenko, line 24: Don't forget about default case for switch / case statement
You can just add break operator here

Anna-Tkachenko (approved):
Good job!

PR #1164, opened 2022-05-01
Develop

frontend112:


meforoma (changes requested):
please apply the `switch / case` statement here - and do remember to apply `default` case to catch errors

ihor-jpeg (approved):

meforoma (approved):

PR #1166, opened 2022-05-02
solution

Uktam19980416:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+      for (const key of keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
ihor-jpeg, line 30: it seems like a good opportunity for a switch case

ihor-jpeg (changes requested):

ihor-jpeg (approved):

VinnieJ-2k20 (approved):

PR #1167, opened 2022-05-03
added first solution

dvdovynets:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
+    }
```
ArtemTopolskyi, line 21: lets use 'switch case' construction

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
```
ArtemTopolskyi, line 20: or you can just assing empty object to state

ArtemTopolskyi (changes requested):

src/transformState.js
dvdovynets: But it will be a new object? It says error  when I am trying to assign an empty object

src/transformState.js
dvdovynets: Done

ArtemTopolskyi (approved):

PR #1168, opened 2022-05-04
Solution /  First commit

Kaunin1:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const order of actions) {
+    switch (true) {
+      case order.type === 'addProperties':
```
mgrinko, line 10: why `switch (true)`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case order.type === 'removeProperties':
+        for (const toBeRemoved of order.keysToRemove) {
+          if (toBeRemoved in state) {
+            delete state[toBeRemoved];
+          }
+        }
+        break;
+
+      default:
+        for (const eraseAll in state) {
```
mgrinko, line 23: `erase` is a verb, so it is a good name for a function, not for an object key

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const order of actions) {
```
mgrinko, line 8: why does `order` is an element of `actions` array?

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+      case order.type === 'removeProperties':
+        for (const toBeRemoved of order.keysToRemove) {
+          if (toBeRemoved in state) {
+            delete state[toBeRemoved];
+          }
+        }
+        break;
+
+      default:
```
mgrinko, line 22: So if we receive an action with a type for example `addProperty` we will delete everything?!

mgrinko (changes requested):

src/transformState.js
Kaunin1: Good advice !  Taken into account, correction applied.

src/transformState.js
Kaunin1: Good advice !  There is a possibility to make switch more obvious.

src/transformState.js
Kaunin1: Excelent advice ! Have implemented addtional  swithc case to handle 'clear' action.

src/transformState.js
Kaunin1: Good advice !  Name of  the element  of actions array was change to more obvious one.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
-}
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const toBeRemoved of action.keysToRemove) {
+          if (toBeRemoved in state) {
```
mgrinko, line 16: I don't understand why don't you call it just `key`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+            delete state[toBeRemoved];
+          }
+        }
+        break;
 
+      case 'clear':
+        for (const stateKey in state) {
```
mgrinko, line 23: just `key` is enough

mgrinko (approved):

src/transformState.js
Kaunin1: Sure , lets keep the names clear and simple, code dry  and be stick to the task.

src/transformState.js
Kaunin1: It' really enough  :))

All your remarks are much appreciated ! Make me focuced on all small but critical points in development !

Olexandr6 (approved):

PR #1172, opened 2022-05-08
Develop

vinnibea:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  
+  for (let i = 0; i < actions.length; i++) {
```
illiarerun, line 9: do you really need index here?

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      for (const word in actions[i].keysToRemove) {
+        delete state[actions[i].keysToRemove[word]];
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
illiarerun, line 26: That's a great place to use `switch/case`

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+  
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const prop in actions[i].extraData) {
+        state[prop] = actions[i].extraData[prop];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const word in actions[i].keysToRemove) {
```
illiarerun, line 17: what's a word? that's a key of object or a property, so name your variables as they should be named

illiarerun: Also `keysToRemove` is an array, you can use `for ... of` here

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        delete state[actions[i].keysToRemove[word]];
```
illiarerun, line 18: and then you'd only need
```suggestion
        delete state[word];
```

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const prop in actions[i].extraData) {
+        state[prop] = actions[i].extraData[prop];
+      }
```
illiarerun, line 13: Objects already have a method to `assign` other objects to it

illiarerun (changes requested):
Nice start, now let's fix your solution:

src/transformState.js
vinnibea: Yep, I have rewrote it as you said, thanks. Btw, still not sure about my formatting, sorry for that (

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+
+        break;
+
+      default:
+
+        return state;
```
illiarerun, line 32: Do we really need this blank line?
```suggestion
      default:
        return state;
```

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
+    switch (actions[key].type) {
+
+      case 'addProperties':
```
illiarerun, line 11: and here too
```suggestion
    switch (actions[key].type) {
      case 'addProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
illiarerun, line 8: [Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) 1st item

illiarerun (changes requested):

src/transformState.js
vinnibea: > [Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) 1st item

But if I have an array of objects, why I have to iterate through for...of instead of for...in?

illiarerun (approved):

PR #1176, opened 2022-05-13
Solution v1.0

allesdmytro:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const actionsObject = action;
+    let removeValue = '';
+
+    if (actionsObject.type === 'addProperties') {
```
aliceblunt8, line 12: let's use switch case in this task

aliceblunt8 (changes requested):

ArtemTopolskyi (approved):

PR #1185, opened 2022-05-21
transformState first commit

artemfurdela:


ArturErmolenko (changes requested):
better use switch case

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
ArturErmolenko, line 27: default

ArturErmolenko (approved):
Good job ,but don't forget about default case its important

vkryvytskyy (approved):

PR #1194, opened 2022-06-01
Solution

ik-web:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const el of obj.keysToRemove) {
+        if (state.hasOwnProperty(el)) {
+          delete state[el];
+        }
+      }
+    } else if (obj.type === 'clear') {
+      for (const deletedItem in state) {
+        delete state[deletedItem];
+      }
+    }
```
illiarerun, line 21: Looks complicated, that's a great place to use switch/case

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+  }
+
+  return state;
```
illiarerun, line 24: it's a void f no need to return anything here

src/transformState.js
ik-web: Fixed

illiarerun (approved):
GJ

PR #1198, opened 2022-06-03
Solution

OlhaArama:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
ArtemTopolskyi, line 9: let's use switch-case construction

ArtemTopolskyi (changes requested):

marina-tilniak (approved):

PR #1206, opened 2022-06-06
transform

TanyaHryn:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[value];
+        }
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
vkryvytskyy, line 25: use `switch..case` instead of separate `if` statements

vkryvytskyy (changes requested):

vkryvytskyy (approved):
well done

PR #1207, opened 2022-06-06
my solution

Samaev:
my solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const array of actions) {
+    if (array.type === 'addProperties') {
+      for (const prop in array.extraData) {
+        state[prop] = array.extraData[prop];
+      }
+    }
+
+    if (array.type === 'clear') {
+      for (const ch in state) {
+        delete state[ch];
+      };
+    }
+
+    if (array.type === 'removeProperties') {
+      for (const prop of array.keysToRemove) {
+        delete state[prop];
+      }
+    }
+  }
+
+  return state;
 }
 
 module.exports = transformState;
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const prop in action.extraData) {
+          state[prop] = action.extraData[prop];
+        }
+        continue;
```
Galyopa, line 14: чого не break?

Galyopa (approved):
Гарна работа, але не зрозуміло чого continue )

src/transformState.js
Samaev: для самого уже загадка) сменил на брейк. спасибо

Samaev:
добавил "брейки"

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+      case 'removeProperties':
+        for (const prop of action.keysToRemove) {
+          delete state[prop];
+        }
+        break;
+    }
+  }
+
+  return state;
```
illiarerun, line 28: you don't need to return state in this task, all you need is to change it, so this return statement is redundant
```suggestion
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const prop in action.extraData) {
+          state[prop] = action.extraData[prop];
+        }
```
illiarerun, line 13: there is a method for that, called Object.assign, or you can use `...` spread operator. using loop ere would be bad

illiarerun (changes requested):
GJ, few fixes left:

src/transformState.js
Samaev: used it, thanks

PR #1211, opened 2022-06-07
Stateful object solution

garikr:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'removeProperties':
+        for (const item of action.keysToRemove) {
+          if (Object.keys(state).includes(item)) {
+            delete state[item];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const item in state) {
```
jstmpelowycz, line 23: as you know, when using for-of loops, we iterate over arrays, thus it's okay to give such names as `item`, but when it comes to iterating over object properties, that's better to call the current value `key`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[item];
+        }
+        break;
```
jstmpelowycz, line 26: you'd rather add default case

jstmpelowycz (changes requested):
good job! just a few things to fix

Olexandr6 (approved):

PR #1213, opened 2022-06-07
Solution

ostapfriend:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return state;
```
jstmpelowycz, line 27: I'd rather return `state` outside of the loop (as far as we don't return anything above so that we preserve the single purpose of the loop)

jstmpelowycz (approved):

src/transformState.js
jstmpelowycz: moreover, we don't need to return anything in this task

PR #1218, opened 2022-06-07
Solution

qw-collab:


LesyaSvk (commented):
Looks nice ✨

LesyaSvk (approved):

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const key of actions) {
```
```suggestion
  for (const action of actions) {
```

jstmpelowycz: here, you're iterating over the array of actions, not object
when iterating over object keys (using for-in loops) that's okay to give the current value name `key`

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+    switch (key.type) {
+      case 'addProperties':
+        Object.assign(state, key.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const remove of key.keysToRemove) {
```
jstmpelowycz, line 16:
```suggestion
        for (const keyToRemove of key.keysToRemove) {
```

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+          if (state.hasOwnProperty(remove)) {
```
jstmpelowycz, line 17: if we try to access non-existing property, we won't get any error, and deleting undefined one won't cause anything to our object

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+            delete state[remove];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const clear in state) {
```
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+          delete state[clear];
+        }
+        break;
```
jstmpelowycz, line 27: you'd rather add default case

jstmpelowycz (changes requested):
good job, just a few things to fix

jstmpelowycz (approved):

PR #1220, opened 2022-06-07
solution

khrystyna-sabat:


Dmytrit (approved):
it's OK

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const remove of action.keysToRemove) {
```
jstmpelowycz, line 15: you'd rather call it `keyToRemove` or just `key`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[remove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
jstmpelowycz, line 24: in the last case of your switch, you don't need to use break-statement

```suggestion
```

jstmpelowycz (approved):

PR #1223, opened 2022-06-07
Solution

IvanGrabovsky:


Pavlo000 (approved):
Привіт, крута робота, переглаянь мої зауваження і якщо можеш ревюшни мою задачку situs inverses

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let result = ({}, state);
```
jstmpelowycz, line 8: you don't even need to create a state copy, just modify the original

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+  for (const act of actions) {
+    if (act.type === 'addProperties') {
```
jstmpelowycz, line 11: use switch (the checklist, p.2)

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let result = ({}, state);
+
+  for (const act of actions) {
```
jstmpelowycz, line 10: don't be mean when giving your vars proper names

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        for (const elem of removeElts) {
+          if (elem === st) {
+            delete result[st];
+          }
+        }
+      }
+    }
+  }
+
+  return result;
```
jstmpelowycz, line 34: you don't need to return anything in the task

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+    if (act.type === 'removeProperties') {
+      const removeElts = act.keysToRemove;
+
+      for (const st in result) {
+        for (const elem of removeElts) {
+          if (elem === st) {
+            delete result[st];
+          }
+        }
```
jstmpelowycz, line 29: this solution is too complicated, find out how to make it simpler

jstmpelowycz (changes requested):
apply fixes to your code:

- give appropriate names
- use switch
- simplify the solution for the case with `removeProperties`

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let newState = state;
```
illiarerun, line 8: you don't need this variable at all, you already have `state`
```suggestion
```

illiarerun (changes requested):
GJ, one fix left:

illiarerun (approved):

PR #1224, opened 2022-06-07
Add stateful object

ghost:


amelentieva (approved):
Cool job!!!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const keys of action.keysToRemove) {
+          delete state[keys];
+        }
+        break;
+
+      default:
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
```
```suggestion
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
```

jstmpelowycz: why important: in this task, you're given a clear explanation of the different cases, and here's a slight amount of code, but when it comes to a larger project, other devs may get stuck understanding what's written

jstmpelowycz: also, don't forget to use the single form as you take out only one item, not a bunch

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keys of action.keysToRemove) {
```
```suggestion
        for (const key of action.keysToRemove) {
```

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1229, opened 2022-06-07
Develop

Tsymbal-Viacheslav:


AKomashchuk (approved):
👌

src/transformState.js
```diff
@@ -1,11 +1,30 @@
 'use strict';
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
jstmpelowycz, line 4: you'd rather use for-of as there is no need for indexes in the task

jstmpelowycz (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const value in action.extraData) {
+          state[value] = action.extraData[value];
+        }
```
```suggestion
        for (const key in action.extraData) {
          state[key] = action.extraData[key];
        }
```

jstmpelowycz: if you have an object, it looks this way:
```
object[key] = value
```
so `value` is inappropriate in this context

jstmpelowycz (approved):
LGTM, hope you have a look at the comment

PR #1230, opened 2022-06-07
Compleated

ant2nta:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
dimabelotskyi, line 13: U can try Object.assign instead of this

dimabelotskyi (approved):

src/transformState.js
jstmpelowycz: good point, apply changes

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return null;
```
jstmpelowycz, line 29: you don't need to return anything in this task

jstmpelowycz (changes requested):

illiarerun (approved):

PR #1231, opened 2022-06-07
Develop

Grodh06:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const types of actions) {
```
illiarerun, line 8: [Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) 3rd item, you're iterating through array of `action`, not `types`

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    switch (types.type) {
+      case 'addProperties':
+        for (const addKey in types.extraData) {
+          state[`${addKey}`] = types.extraData[`${addKey}`];
+        }
```
illiarerun, line 13: there's a special method for that - Object.assign(), or you can use `...` spread operator. Both would be better than a loop

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          delete state[`${clearKey}`];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
illiarerun, line 33:
```
/**
 * @param {Object} state
 * @param {Object[]} actions
 */
 ```
 you don't need to return anything in this task, just mutate the state
```suggestion
```

illiarerun (changes requested):
GJ, now let's work on some fixes:

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const removeKey of action.keysToRemove) {
+          delete state[`${removeKey}`];
```
illiarerun, line 16: these `string` is redundant, key value is already a string
```suggestion
          delete state[removeKey];
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const clearKey in state) {
+          delete state[`${clearKey}`];
```
illiarerun, line 22: same here

illiarerun (changes requested):

Olexandr6 (approved):

PR #1233, opened 2022-06-07
Solution

Cherednychenko:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return {};
```
LesyaSvk, line 27: we don't need to return anything in this task 😅

LesyaSvk (approved):

src/transformState.js
jstmpelowycz: yeah, good point, either remove the default case or make it like `default: break`

jstmpelowycz (approved):

jstmpelowycz (approved):

PR #1234, opened 2022-06-07
solution

LAndrikevych:


ik-web (approved):

ik-web:
Класно зроблено!👍

jstmpelowycz (approved):

src/transformState.js
```diff
@@ -6,6 +6,24 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      case 'clear':
+        for (const property in state) {
```
jstmpelowycz, line 22: that's more about key, not property, so I'd give it a name like `key` instead of `property`

PR #1235, opened 2022-06-07
Solution

Bohdan-Pyschchyk:


yanaost (approved):
it is better to use switch according to the checklist
as well as avoiding variables name without meaning like char

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+  for (const char of actions) {
+    if (char.type === 'addProperties') {
+      for (const exData in char.extraData) {
+        state[exData] = char.extraData[exData];
+      }
+    }
+
+    if (char.type === 'removeProperties') {
+      for (const keyRemove of char.keysToRemove) {
+        if (state.hasOwnProperty(keyRemove)) {
```
jstmpelowycz, line 17: if a given `keyToRemove` doesn't exist on `state`, we won't delete anything, thus there is no need to check if one contains it

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #1237, opened 2022-06-07
done

k0nstrukt-or:


DmytroSukhaniuk:
good job!! but admit some mistakes below

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {           //sort out an array with actions
```
jstmpelowycz, line 8: if code is written properly, there is no need to add comments, avoid doing so
generally, if some code is required to have descriptions, docs are created

jstmpelowycz (changes requested):

jstmpelowycz (approved):
LGTM

PR #1240, opened 2022-06-07
solution

AKomashchuk:


Tsymbal-Viacheslav (approved):

src/transformState.js
```diff
@@ -5,7 +5,20 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'clear' :
+        Object.keys(state).forEach(key => delete state[key]);
```
jstmpelowycz, line 15: I'd rather use a for-in loop here
generally, that's not a common thing to use `Object.keys(...)` for such cases

jstmpelowycz (approved):

src/transformState.js
AKomashchuk: Я думала , що не можна робити  for-in  в for-of,   (Nested loops === EVIL)

PR #1246, opened 2022-06-07
Solution

amelentieva:


juajum (approved):
хорошая работа)

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const args of actions) {
```
jstmpelowycz, line 8: you'd rather call it `action`

PR #1248, opened 2022-06-08
solution

viktorlyt:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+    }
+  }
+
+  return state;
```
Kate-Khon, line 30: На мiй погляд - тут можна не писати `return`

Kate-Khon (approved):
👍

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const property in action.extraData) {
+          state[property] = action.extraData[property];
```
jstmpelowycz, line 12: property is more like a key-value pair, so you'd rather call `key`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
```
jstmpelowycz, line 17: here, you'd rather call it `keyToRemove`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          delete state[property];
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
```
jstmpelowycz, line 23: the same way here)

src/transformState.js
jstmpelowycz: yes, in this task we don't need to return anything, so that's redundant

jstmpelowycz (approved):

PR #1252, opened 2022-06-08
solution was added

Bielichenko:


src/transformState.js
```diff
@@ -4,8 +4,32 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
illiarerun, line 11: use proper `switch` case, put your type as a value and it's values for cases, like this:
```suggestion
    switch (action.type) {
      case 'addProperties':
```

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+        break;
+
+      case action.type === 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return state;
```
illiarerun, line 28: you don't need to return anything in this func, you can use `break` here

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+    }
+  }
+
+  return state;
```
illiarerun, line 32: same here
```suggestion
```

illiarerun (changes requested):
GJ, let's work on fixes now:

illiarerun (approved):

PR #1258, opened 2022-06-08
finished transformState function

omelika:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return state;
```
illiarerun, line 27: you don't need to return anything in this function
```suggestion
        break;
```

illiarerun (approved):
GJ, left one comment:

PR #1261, opened 2022-06-09
solution v01

Taras-Kozii:


ArturErmolenko (changes requested):
Use switch case

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const elem of actions) {
+    switch (true) {
+      case elem.type === 'addProperties':
```
aliceblunt8, line 10:
```suggestion
    switch (elem.type) {
      case 'addProperties':
```

aliceblunt8 (changes requested):
incorrect usage of switch case

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const elem of actions) {
```
```suggestion
  for (const action of actions) {
```

Bohdan-Maksymiuk (approved):

ArturErmolenko (approved):

PR #1262, opened 2022-06-11
Solution

dmytro-legerda:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  const stateCopy = state;
+
+  actions.forEach(function(action) {
```
vkryvytskyy, line 11:
```suggestion
  actions.forEach((action) => {
```
Use arrow functions

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      for (const key of action.keysToRemove) {
+        delete stateCopy[key];
+      };
+    }
+
+    if (action.type === 'clear') {
+      for (const key in stateCopy) {
+        delete stateCopy[key];
+      }
+    }
```
vkryvytskyy, line 26: Use switch case

vkryvytskyy (changes requested):

aliceblunt8 (approved):

PR #1263, opened 2022-06-11
solution

DatsOlena:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
Bohdan-Maksymiuk, line 9: be aware of such destruction  - not every action have every property

Bohdan-Maksymiuk (approved):
Great job 👏

PR #1264, opened 2022-06-11
task done

Zykhor-Andrii:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const char of actions) {
```
```suggestion
  for (const action of actions) {
```
use clear naming

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    for (const key in char) {
```
Bohdan-Maksymiuk, line 9: you don't need this loop - you have preset properties - so check them as you do below (you may also use `switch`)

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1266, opened 2022-06-14
Solution

pavel-gutsal:
first commit

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -7,22 +7,22 @@
 function transformState(state, actions) {
   // write code here
   for (const object of actions) {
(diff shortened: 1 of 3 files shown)


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const object of actions) {
+    if (object.type === `addProperties`) {
```
ArtemTopolskyi, line 10: use switch-case

ArtemTopolskyi (changes requested):

pavel-gutsal:
switch statements were implemented

PR #1269, opened 2022-06-19
Develop

AntonChetverykMA:


src/transformState.js
```diff
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  const arr = Object.values(actions);
```
vkryvytskyy, line 9: actions is an array already, why do you need this?

vkryvytskyy (changes requested):

src/transformState.js
AntonChetverykMA: it was encrypted code)))

vkryvytskyy (approved):

PR #1278, opened 2022-06-30
Stateful object task solution

juliiaap:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      }
+    }
+
+    if (object.type === 'removeProperties') {
+      for (const key of object.keysToRemove) {
+        if (key in state) {
+          delete state[key];
+        }
+      }
+    }
```
Bohdan-Maksymiuk, line 27: you can use `switch case` for such checks

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (object.type === 'addProperties') {
+      for (const key in object.extraData) {
+        state[key] = object.extraData[key];
+      }
+    }
```
Bohdan-Maksymiuk, line 19: you can use `Object.assign()` and avoid loop here

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```
use clear naming

Bohdan-Maksymiuk (changes requested):

src/transformState.js
juliiaap: done

src/transformState.js
juliiaap: done

src/transformState.js
juliiaap: done

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      default:
+        for (const key of action.keysToRemove) {
+          if (key in state) {
+            delete state[key];
+          }
```
marina-tilniak, line 24: те саме тут

juliiaap: fixed

ArturErmolenko (approved):

PR #1293, opened 2022-07-04
Solution

OlgaStakhova:


butenko108 (approved):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    if (action.type === 'addProperties') {
```
VETALEK, line 9: you may use `switch case` here

VETALEK (changes requested):
Good start 👍

VETALEK (approved):
🔥

PR #1297, opened 2022-07-04
initial solution

DashaPoliazz:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+// write code here
+  actions.forEach(stage => {
+    if (stage.type === 'addProperties') {
+      for (const keyToAdd in stage.extraData) {
+        state[keyToAdd] = stage.extraData[keyToAdd];
+      }
```
Oleksandr-Pustovit, line 13: Сan be simplified  with Object.assign()

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+    } else if (stage.type === 'removeProperties') {
+      stage.keysToRemove.forEach(keyToRemove => {
```
Oleksandr-Pustovit, line 15: There is no actual need in `else` statement here. You will always have exactly one "type" and it will obly work with one if statement.

```suggestion
    }
    if (stage.type === 'removeProperties') {
      stage.keysToRemove.forEach(keyToRemove => {
```

Oleksandr-Pustovit (changes requested):
Hi! Good job.
It is mentioned in [checklist 2](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) to use switch for a limited amount of conditions. It might be a good choice if you have to write so many if else cases.

PR #1298, opened 2022-07-04
Solution

dashasm:


DashaPoliazz (approved):
Nice work.

src/transformState.js
```diff
@@ -1,11 +1,28 @@
...
+      obj.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    }
+
+    if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VETALEK, line 23: you may use `switch case` here

VETALEK (changes requested):
Nice start 💪

src/transformState.js
```diff
@@ -1,11 +1,29 @@
...
+        obj.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
 
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
VETALEK, line 23: add also `default:`, that throws an error when there is an unknown type

VETALEK (changes requested):
Gooood 🤩

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+
+      default:
+        return 'Unknown type';
```
VETALEK, line 26: `throw new Error('Unknow action type')`
cause now you are returning a string)

VETALEK (changes requested):
last charge 💪

solaryasha (approved):

PR #1299, opened 2022-07-04
task js_stateful-object

Moblat890:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (key.type === 'removeProperties') {
+      for (const keyToDel of key.keysToRemove) {
+        delete state[keyToDel];
+      }
+    } else if (key.type === 'clear') {
+      for (const keyOfState in state) {
+        delete state[keyOfState];
+      }
+    }
+  }
```
marina-tilniak, line 20: use switch case here

marina-tilniak (changes requested):

src/transformState.js
Moblat890: why can't it be

marina-tilniak (approved):

PR #1300, opened 2022-07-04
Solution

LeonidPustovit:


DashaPoliazz (approved):
Nice solution.

You are using `obj['keyName']` syntaxis to get object value. I know that you know you can do it easier by using `obj.keyName`. I heard that second variant is best practise. Your variant we must use when we need to get so type of keys like 'key Name' or another specific key.

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action['type']) {
```
solaryasha, line 9:
```suggestion
    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
solaryasha, line 11:
```suggestion
        Object.assign(state, action.extraData);
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action['keysToRemove']) {
```
solaryasha, line 15:
```suggestion
        for (const property of action.keysToRemove) {
```

solaryasha (approved):
You don't need to use square brackets for accessing `static` key

PR #1301, opened 2022-07-04
added transform state function

Oleksandr-Pustovit:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VETALEK, line 23: use `switch case` here (and don't forget to mention `default` case)

VETALEK (changes requested):
Good start 👏

solaryasha (approved):
Good job !

PR #1304, opened 2022-07-05
JS stateful object (first solution)

Oleksandr-Naumchak:
JS stateful object (first solution)

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
```
marina-tilniak, line 21: використай switch case

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
ArturErmolenko, line 25: default

ArturErmolenko (approved):
Don't forget about default case

PR #1305, opened 2022-07-05
Solution

VolodymyrDolishnii:


Oleksandr-Pustovit (approved):
Well done!

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
```
VETALEK, line 28: it'd be cool to add `default` case, and throw some error if there is an unknown action type

VETALEK (approved):
Nice 👌

PR #1315, opened 2022-07-05
Develop

illnino380:


PavloMarkov (approved):
good job! 🔥

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'Use the right action';
```
VETALEK, line 29: it must be `throw new Error(...)`, not just a string

VETALEK (changes requested):
Good start 👌

VETALEK (approved):
👍

PR #1317, opened 2022-07-05
Solution

viktor-kotik:


yuliia-dronova (approved):
Great!
I've forgotten about default, thank you for good case)

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
VETALEK, line 26: throw some error here

VETALEK (changes requested):
Good start 👏

src/transformState.js
viktor-kotik: Ок

VETALEK (approved):
👍

PR #1319, opened 2022-07-05
solution

vitmixG:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const removeProperties in state) {
+          delete state[removeProperties];
+        }
+        break;
+
+      default:
+        break;
```
VETALEK, line 29: throw some error here

VETALEK (changes requested):
Good start 🚀

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+      case 'clear':
+        for (const removeProperties in state) {
+          delete state[removeProperties];
+        }
+        break;
+
+      default:
+        return 'Enter an existing type';
```
VETALEK, line 29:
```suggestion
        throw new Error('Enter an existing type');
```

I mean smth like this 😉

VETALEK (approved):
Nice 👌

PR #1320, opened 2022-07-05
Transform-Form-State function was created

yuliia-dronova:


Danil-Hodytskyi (approved):
Good Job

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
VETALEK, line 28: better throw some error here

VETALEK (changes requested):
Good start 🚀

VETALEK (approved):
👍

PR #1321, opened 2022-07-05
solution

Danil-Hodytskyi:


opanasiukvlad (approved):
За умовою просять не забувати про дефолт значення кейсу, а так все ок )

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
VETALEK, line 26: better throw some error here

VETALEK (changes requested):
Good start 👍

src/transformState.js
Danil-Hodytskyi: I understand that I can delete default case?

src/transformState.js
```diff
@@ -22,9 +22,6 @@ function transformState(state, actions) {
           delete state[key];
         }
         break;
-
-      default:
```
VETALEK: it would be best if you had `defautl` case, where you throw some error - so user can know, that he gave you a wrong action type

VETALEK (changes requested):
Almost 💪

src/transformState.js
Danil-Hodytskyi: okay

VETALEK (approved):
👍

PR #1323, opened 2022-07-05
solution

roman-romanenko:


yuriiLevchenko2022 (approved):
Greate!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
VETALEK, line 28:
```suggestion
      default:
        throw new Error('your error mesage');
```

better do smth like that in `default`

VETALEK (changes requested):
Good start fire

VETALEK (approved):
👍

PR #1324, opened 2022-07-05
solution

chernobrovin:


tetyanabukoros (approved):
Great!

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
VETALEK, line 31: throw some error here - so user can know, that he gave you a wrong action type

VETALEK (changes requested):
Good start 🔥

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'error';
```
VETALEK, line 31:
```suggestion
        throw new Error('some error message');
```

I meant smth like this 😉

VETALEK (changes requested):
Almost 💪

src/transformState.js
chernobrovin: Okay, but we haven’t taught this yet ((I'll make it especially for you))

VETALEK (approved):
👍

PR #1325, opened 2022-07-05
Task solution

LovelyDrug:


chernobrovin (approved):
Good work!

src/transformState.js
```diff
@@ -4,8 +4,32 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const act = action.type;
```
VETALEK, line 10: why `act`? maybe `type` is better?

VETALEK (approved):
😎

PR #1326, opened 2022-07-05
Solution

rodiontseva:


vitmixG (approved):
Well done!

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        return 'this doesn\'t run';
```
VETALEK, line 27: better throw an error here
```suggestion
        throw new Error('this doesn\'t run');
```

VETALEK (changes requested):
Nice 👌 
commented

VETALEK (approved):
🚀

PR #1328, opened 2022-07-05
ruslan-petrenko / js_stateful-object

ruslan-petrenko:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+
+      case 'clear':
+        for (const stateElement in state) {
+          delete state[stateElement];
+        }
+        break;
+    }
+  }
+
+  return state;
```
VETALEK, line 29: not sure, that you need to return smth

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
VETALEK, line 10: you need `default` case here, where you can throw an error, so user can know, that he gave an unknown property

VETALEK (changes requested):
Good start 🔥

src/transformState.js
ruslan-petrenko: added default case

PR #1330, opened 2022-07-06
Solution

Lentyaeev:


Vova-Balyshev (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(e => {
```
VETALEK, line 8:
```suggestion
  actions.forEach(action => {
```

why `e`?

VETALEK (changes requested):
GJ 👍 
but you have a bit confusing naming

PR #1333, opened 2022-07-06
stateful object solution

AlexanderKolomiiets:


dellachaise:
You could ask one of students which already did this task for students review. And after fixes if needed set it for mentor review.

Githi54 (approved):
good job!👍

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const result = state;
```
dellachaise, line 8: use `spread operator`, because this way you mutate `state`.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
dellachaise, line 11: Great job 👍  But in this case better use `switch case`

dellachaise (changes requested):

dellachaise (approved):
Great job 👍

PR #1338, opened 2022-07-08
transformState

darynakravchuk:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+        delete newObj[i];
+      }
+    }
+
+    if (a.type === 'clear') {
+      for (const k in newObj) {
+        delete newObj[k];
+      }
+    }
+  }
```
marina-tilniak, line 27: use switch case

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  const newObj = state;
+
+  for (const a of actions) {
```
marina-tilniak, line 11:
```suggestion
  for (const action of actions) {
```

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          delete newObj[i];
+        }
+        break;
+
+      default:
+        for (const k in newObj) {
+          delete newObj[k];
+        }
```
valerii-derkach, line 26: it is not good practice to use default case for condition
use it in `case 'removeProperties'`
and in `default` just `break`

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          delete newObj[i];
+        }
+        break;
+
+      default:
+        for (const k in newObj) {
```
valerii-derkach, line 24: same here 
name variables semantic, it is not an index

valerii-derkach (changes requested):

PR #1344, opened 2022-07-14
Solution ver 1.0

Oleksii-Korovnichenko:
Hi. I used functions to clear and remove keys from object. I wasn't trying to use nested loops. 
Is it correct?
Please give me feedback.


ArturErmolenko (approved):
Good job 👍

PR #1347, opened 2022-07-16
Solution

mariabramnik:


src/transformState.js
```diff
@@ -6,6 +6,34 @@
...
   // write code here
+  const myState = state;
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties' :
+
+        for (const key in action.extraData) {
+          myState[key] = action.extraData[key];
+        }
```
valerii-derkach, line 17: it is better to use Object.assign here, less code

src/transformState.js
```diff
@@ -6,6 +6,34 @@
...
+        break;
+
+      case 'removeProperties':
+        const keysArr = action.keysToRemove;
+
+        for (const keyRemove of keysArr) {
+          delete myState[keyRemove];
+        }
+        break;
+    }
```
valerii-derkach, line 33: add default case

valerii-derkach (changes requested):

PR #1350, opened 2022-07-19
Solution

Serhii-Vovk:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete state[remove];
+      }
+    }
+
+    if (data.type === 'addProperties') {
+      Object.assign(state, data.extraData);
+    }
+  }
+
+  return state;
```
valerii-derkach, line 26: please use switch case instead if here

PR #1355, opened 2022-07-22
Develop

KostiukPavlo:


src/transformState.js
```diff
@@ -62,7 +62,29 @@
  * @param {Object[]} transforms
  */
 function transformState(state, transforms) {
-  // write code here
+  const result = state;
+
+  for (const transformsItems of transforms) {
```
Bohdan-Maksymiuk, line 67:
```suggestion
  for (const transformsItem of transforms) {
```

Bohdan-Maksymiuk: or better `transform` - just singular

src/transformState.js
```diff
@@ -62,7 +62,29 @@
...
+      for (const key in transformsItems.properties) {
+        result[key] = transformsItems.properties[key];
+      }
+    }
+
+    if (transformsItems.operation === 'removeProperties') {
+      for (const value of transformsItems.properties) {
+        delete state[value];
+      }
+    }
```
Bohdan-Maksymiuk, line 84: use `switch case`

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1360, opened 2022-07-26
solution

AndreyJigijalkin:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    if (actions[i].type === 'clear') {
+      for (const key in result) {
+        delete result[key];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const keys in result) {
+        if (actions[i].keysToRemove.toString().includes(keys)) {
+          delete result[keys];
```
AlexanderKolomiiets, line 24: In loop you could iterate only keysToRemove instead of whole result.

AlexanderKolomiiets (approved):
looks fine )

MaksimusCRS (changes requested):
Use switch/case

MaksimusCRS (approved):

PR #1373, opened 2022-08-02
Solution

Dmytro-Matsehora:


dellachaise:
You could ask one of students which already did this task for students review. And after fixes if needed set it for mentor review.

sergii-nosachenko (changes requested):
Hi, Dima!
There are some minor fixes I've found so far. Feel free to discuss. ✌

src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let key = 0; key < actions.length; key++) {
```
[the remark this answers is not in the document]
Dmytro-Matsehora: Thanks for your tips and comments, i used them and now my code looks really cleaner 👍

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i in actions) {
```
dellachaise, line 8: Create more meaningful variable instead of `i`.

dellachaise (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[`${key}`];
```
sergii-nosachenko, line 15:
```suggestion
        delete state[key];
```
You don't need to convert `key` to string, cause inside `[]` the value of variable will be autonatically converted to string.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
sergii-nosachenko, line 9:
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        // case code
        break;
```

You can reformat your conditions as switch...case statement. But it's up to you.

sergii-nosachenko (changes requested):
One little fix and optional improvement (switch...case) for your decision.

src/transformState.js
Dmytro-Matsehora: fixed

src/transformState.js
Dmytro-Matsehora: Generally yes, but in this case I like the `if ` construct more, since there are only 3 options, but if there were more of them, I would use switch case =)

sergii-nosachenko (approved):
Good work!

danylo-onishchuk (approved):

PR #1374, opened 2022-08-02
solution

Kulchevych:


sergii-nosachenko (approved):
Solid as rock! 🦾
Only one optional recommendation.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
```
dellachaise, line 16: Better create a copy of state and then operate with it. Change functions arguments - bad practice.

dellachaise (changes requested):
Everything works great 👍 , but better rewrite to `switch case` instead of `if`.

PR #1375, opened 2022-08-02
Solution transformState

AnastasiaTitarenko:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const keyOfState in state) {
+        delete state[keyOfState];
+      }
+    }
+  }
 }
```
marina-tilniak, line 27: use switch case here

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
```
alexandra-protyanova, line 11: u can use Object.assign here

alexandra-protyanova (approved):

PR #1376, opened 2022-08-02
solution

Anton-Liada:


src/transformState.js
```diff
@@ -1,11 +1,28 @@
+/* eslint-disable no-unused-vars */
```
sergii-nosachenko, line 1:
```suggestion
```

You don't need this rule to be disabled.

src/transformState.js
```diff
@@ -1,11 +1,28 @@
...
 'use strict';
 
 /**
  * @param {Object} state
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
sergii-nosachenko, line 9:
```suggestion
  for (const action of actions) {
```

Maybe here is better to give a conscious name to variable.

sergii-nosachenko (changes requested):
Hi, Anton! Good job.
A few fixes and it will be perfect)

sergii-nosachenko (approved):
Good job! 👌

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
```
dellachaise, line 15: Better create a copy of  `state` and then operate with it. Change functions arguments - bad practice.

dellachaise (changes requested):
Everything works great 👍 , but better rewrite to `switch case`.

dellachaise (approved):
Well done 👍

PR #1378, opened 2022-08-03
solution

tiserett:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
sergii-nosachenko, line 9: Better use `switch-case` construction. Or use just `if`, cause there are only unique cases here. It will be more clearer. It's a mentor-preferred solution 😬

sergii-nosachenko (changes requested):
Hi! Some code reformat needed. Be sure, it will speed up mentor's review 😎

src/transformState.js
tiserett: code looks much cleaner with switch statement, thanks!

sergii-nosachenko (approved):
Perfect! 🔥

AndriiRut (approved):
🔥

PR #1381, opened 2022-08-04
Solution

YuliaDzisiak:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act in actions) {
+    const actObj = actions[act];
```
Bohdan-Maksymiuk, line 9: do not shortcut naming of variables

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const key in actObj.keysToRemove) {
+        const prop = actObj.keysToRemove[key];
+
+        delete state[prop];
+      }
+    } else {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Bohdan-Maksymiuk, line 23: use `switch` for such predefined cases

Bohdan-Maksymiuk (changes requested):

Virass (approved):

PR #1382, opened 2022-08-05
Solution

nikitaovch99:


artem-musii (approved):
I think that it's better to use switch here, but it's up to you

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i]['type'] === 'addProperties') {
+      Object.assign(state, actions[i]['extraData']);
+    } else if (actions[i]['type'] === 'removeProperties') {
+      const remove = Object.keys(state);
+
+      for (let g = 0; g < remove.length; g++) {
+        for (let k = 0; k < remove.length; k++) {
```
naziamoff, line 15: You should avoid nested loops when it is possible

naziamoff: if you have limited amount of conditions - you should use switch/case

naziamoff (changes requested):

naziamoff (approved):
Nice!

PR #1385, opened 2022-08-06
solution

andredolgoleo:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const char of actions) {
```
```suggestion
  for (const action of actions) {
```
what does char mean?)

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1390, opened 2022-08-07
Solution transformState

wfurs0w:


Andrii-Stiebalov (approved):
Very good, but better use switch.)
image

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const char of actions) {
+    switch (char.type) {
```
RomanYakm, line 9: switch/case should always have default case for error handling.

RomanYakm (changes requested):
checklist 4

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
RomanYakm, line 9: switch/case should always have default case for error handling.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[prop];
+        }
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
+    }
```
RomanYakm, line 25: I think you need to add default : throw new Error(Unknown action type: ${action.type});

RomanYakm (approved):
GJ. Just few suggestion

MaksimusCRS (approved):

PR #1391, opened 2022-08-07
solutionTransformState

YuriiPronin:


plemiannik (approved):
wow, looks great, mb not sure about return in the end, but generally i like it, awesome job 👍

AndriiRut (approved):
🔥

PR #1396, opened 2022-08-08
Solution

smiyka:


alexandra-protyanova (changes requested):
let us switch for this task

alexandra-protyanova (approved):
Good job!

PR #1397, opened 2022-08-08
Solution

andrii-dmitruk:


src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        const propsToAdd = Object.entries(extraData);
+
+        for (const [key, value] of propsToAdd) {
+          state[key] = value;
+        }
```
VolodymyrChuyko, line 17: Your solution is ok, but it could be done easier with Object.assign

VolodymyrChuyko (approved):

MaksimusCRS (approved):

PR #1398, opened 2022-08-08
transformState solution

BipBobBoop:


alexandra-protyanova (changes requested):
let us switch case here

alexandra-protyanova (approved):
Well done 👍

PR #1399, opened 2022-08-08
solution

artem-kodesnikov:


Andrii-Stiebalov (approved):
image
Very good. But better use switch)

danylo-onishchuk (approved):

PR #1401, opened 2022-08-08
done

Maxim-Gumeniuk:


src/transformState.js
```diff
@@ -5,14 +5,14 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  for (const obj of actions) {
-    if (obj.type === 'addProperties') {
-      Object.assign(state, obj.extraData);
-    } else if (obj.type === 'removeProperties') {
-      for (const key of obj.keysToRemove) {
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
anyakrainiuk, line 9: Instead of 'if' better use switch case(according to checklist)

anyakrainiuk: And don't forget about 'default' case

anyakrainiuk (changes requested):

anyakrainiuk (approved):

MaksimusCRS (approved):
GJ

PR #1410, opened 2022-08-08
Solution

anna-kolmakova:


Alzay007 (approved):
nice job) it would be more convenient probably through a cycle 'for of'

Alzay007 (approved):

danylo-onishchuk (approved):

PR #1416, opened 2022-08-09
implemented the function transformState

ShmelyovOleg:


max-kravchenko (approved):
Great!! 

p.s. Maybe 'default' case should have smth like throw new Error('bla bla')?

ShmelyovOleg:
> Great!!
> 
> p.s. Maybe 'default' case should have smth like throw new Error('bla bla')?

done)

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear' :
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      default :
+      throw new Error(`Invalid ${action.type}`);
```
AndriiRut, line 29:
```suggestion
      default:
        throw new Error(`Invalid ${action.type}`);
```

AndriiRut (approved):
🔥

PR #1422, opened 2022-08-09
my solution of stateful object

OleksiiPustovit:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
```
ibondrk, line 10: Use switch statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
ibondrk, line 9: Use switch statement if you have limited amount of conditions. The same below

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (state.hasOwnProperty(key)) {
```
ibondrk, line 15: You can not check whether such a key exists, becouse if not - u'll can't delete it from your object. So i think that it's  futile action

ibondrk (changes requested):

src/transformState.js
OleksiiPustovit: Thanks for your feedback, I've made changes))

ibondrk (approved):
Good job!🔥

AndriiRut (approved):
🔥

PR #1431, opened 2022-08-09
Solution

Anatolii-Bukoros:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        for (const prop of keysToRemove) {
+          delete state[prop];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
```
RomanYakm, line 25: I think you need to add "throw new Error(`Unknown action type: ${action.type}`);"
for default

RomanYakm (approved):
GJ. Just one suggestion.

src/transformState.js
MaksimusCRS, line 25: At least break required

MaksimusCRS (approved):

PR #1433, opened 2022-08-09
the task is done

Anastasiya145:


alexandra-protyanova (changes requested):
It would be better to use switch case in this task

ArtemTopolskyi (approved):

PR #1434, opened 2022-08-09
Stateful object

maks-golovatyuk:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
```
alexandra-protyanova, line 8: use meaningful names for variables not just `i`

alexandra-protyanova (approved):

PR #1437, opened 2022-08-10
solution

vadym-zahazei:


src/transformState.js
```diff
@@ -5,7 +5,49 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
AndriiRut, line 9: look how the `switch case` is writing

AndriiRut (changes requested):

src/transformState.js
vadym-zahazei: > look how the `switch case` is writing
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

I used curly brackets because I saw that this is how Yuriy writes `switch case`

vadym-zahazei: and thought that it would be better that way

vadym-zahazei: > look how the `switch case` is writing
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

done

AndriiRut (approved):
🚀 

using separate function for actions which you do only once it's overkill but here it's ok)

PR #1439, opened 2022-08-10
Solution

SvyatoslavYatsenko:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+    if (type === 'removeProperties') {
+      for (const a in keysToRemove) {
+        delete state[keysToRemove[a]];
+      }
+    }
+
+    if (type === 'clear') {
+      for (const b in state) {
+        delete state[b];
```
Alzay007, line 23: chekclist #2 : use switch

Alzay007 (changes requested):
add default - checklist #4

Alzay007 (approved):
good 👍

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'Error';
```
AndriiRut, line 27: what if you write just `return;` at the end of the function, will it work with it?)

AndriiRut (approved):
🔥

PR #1443, opened 2022-08-10
1st version

Grechana:


alexandra-protyanova (changes requested):
Let us use switch case in this task

valerii-derkach (approved):

PR #1444, opened 2022-08-11
Solution

Andrii-Dunets:


vladgusakovskiy (approved):
Not bad at all! But also, you can use **switch** statement if you have a limited amount of conditions.

MaksimusCRS (changes requested):
Use switch case

AlinaMatuschak (approved):
4 point from the checklist 🥲

PR #1449, opened 2022-08-11
solved

PaulVoron:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        Object.assign(state, actions[i].extraData);
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          for (const key in state) {
```
alexandra-protyanova, line 16: it is not necessary to use loop here, u can just check this property with `hasOwnProperty`

alexandra-protyanova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
marina-tilniak, line 8: you can use for of here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    switch (actions[i].type) {
+      case 'addProperties':
+        Object.assign(state, actions[i].extraData);
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          if (state.hasOwnProperty(actions[i].keysToRemove[j])) {
+            delete state[actions[i].keysToRemove[j]];
+          }
```
marina-tilniak, line 18: and here

marina-tilniak (changes requested):

marina-tilniak (approved):

PR #1453, opened 2022-08-14
Transform state function was created

ArmPlay:


ArturErmolenko (changes requested):
avoid if else construction ,better use switch case

stanislav-topikha (approved):

PR #1456, opened 2022-08-15
Solution

TheVovchik:


src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+      toClear();
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      toRemove(actions[i].keysToRemove);
+    }
+
+    if (actions[i].type === 'addProperties') {
+      toAdd(actions[i].extraData);
+    }
```
grafit-off, line 19: You should use the `switch/case` statement as written in checklist
image

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
```
grafit-off, line 9: If you used the `for` loop, it's gonna be better if you set a name for actions[i], for better readability

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+    for (const key in state) {
+      if (x.includes(key)) {
+        delete state[key];
+      }
+    }
+  }
+
+  function toAdd(x) {
+    Object.assign(state, x);
+  }
```
grafit-off, line 38: Looks interesting

grafit-off (approved):
Keep going, I left some comments below 🔥

src/transformState.js
```diff
@@ -5,7 +5,46 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear': {
+        toClear();
```
naziamoff, line 11: I don't think you really need to create separate functions here. It looks to me that it makes your code more difficult to read. Try it without additional functions

src/transformState.js
```diff
@@ -5,7 +5,46 @@
...
+    }
+  }
+
+  function toClear() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  function toRemove(x) {
```
naziamoff, line 37: you should never name your function with just one letter. Variable should describe what is stored inside

naziamoff (changes requested):
Well done, but there are some comments

src/transformState.js
TheVovchik: maybe I will use them in the future - this functions, so I create them. for me when I write this was more understandable to work with separate function

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in state) {
+          if (action.keysToRemove.includes(key)) {
+            delete state[key];
+          }
```
naziamoff, line 20: You may iterate not the state but `keysToRemove`
```suggestion
        for (const keyToRemove of action.keysToRemove) {
            delete state[keyToRemove];
        }

naziamoff (changes requested):
Sorry for your functions man 😢 
But sometimes we have to remove our loved ones for making a code less complicated

naziamoff (approved):
Nice 🔥

PR #1461, opened 2022-08-17
Solution

miamarichka:


src/transformState.js
```diff
@@ -4,8 +4,41 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        if (state[key]) {
+          state[key] = action.extraData[key];
+        }
+
+        if (!state[key]) {
+          state[key] = action.extraData[key];
+        }
```
marina-tilniak, line 20: use Object.assign here

marina-tilniak (changes requested):
let's use switch case in this task

alexandra-protyanova (approved):
GJ!

PR #1462, opened 2022-08-17
Solution

offsset:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const key of result.keysToRemove) {
+        delete state[key];
+      }
+    } else if (result.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
```
marina-tilniak, line 25: let's use switch case here

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+        break;
+      case (result.type === 'clear'):
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
```
SanyaBratashchuk, line 28: switch(true) is useless) Use switch(result.type) so you will be able to use cases in proper way `case 'clear': 
some code`

SanyaBratashchuk (changes requested):

marina-tilniak (approved):

PR #1465, opened 2022-08-18
Solution

Taras-1:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        }
+      }
+
+      if (currentObject[key] === 'clear') {
+        for (const keyObject in state) {
+          delete state[keyObject];
+        }
+      }
+    }
+  }
```
marina-tilniak, line 32: better use switch case here

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+            delete state[ch];
+          }
+          break;
+
+        case currentObject[key] === 'clear':
+          for (const keyObject in state) {
+            delete state[keyObject];
+          }
+          break;
+      }
```
alexandra-protyanova, line 30:
```suggestion
      switch (currentObject[key] ) {
        case 'addProperties':
          Object.assign(state, currentObject.extraData);
          break;

        case 'removeProperties':
          const deleteKeys = currentObject.keysToRemove;

          for (const ch of deleteKeys) {
            delete state[ch];
          }
          break;

        case 'clear':
          for (const keyObject in state) {
            delete state[keyObject];
          }
          break;
      }
```

alexandra-protyanova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
SanyaBratashchuk, line 8: usu `for...of` loop

SanyaBratashchuk (changes requested):

SanyaBratashchuk (approved):

PR #1466, opened 2022-08-19
js_stateful-object done

EhorSaprykin:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
```suggestion
  for (const action of actions) {
```
why do you need `for in` for iterating array?

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      Object.assign(state, extraData);
+    } else if (actionType === 'removeProperties') {
+      for (const removeProp of removeArr) {
+        delete state[removeProp];
+      }
+    } else if (actionType === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Bohdan-Maksymiuk, line 23: avoid such nesting `else if` conditions
Better to use `switch` here and switch by action type

Bohdan-Maksymiuk (changes requested):

src/transformState.js
EhorSaprykin: @Bohdan-Maksymiuk  done

src/transformState.js
EhorSaprykin: @Bohdan-Maksymiuk done

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const [removeArr, actionType, extraData] = [
+      action.keysToRemove, action.type, action.extraData];
```
stanislav-topikha, line 10:
```suggestion
    const { keysToRemove, type, extraData } = action;
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+    switch (actionType) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const removeProp of removeArr) {
+          delete state[removeProp];
```
stanislav-topikha, line 19:
```suggestion
          delete state[key];
```

stanislav-topikha (changes requested):
comments ⬇️

src/transformState.js
EhorSaprykin: @stanislav-topikha done

src/transformState.js
EhorSaprykin: @stanislav-topikha done

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
SanyaBratashchuk, line 29: remove default if you do nothing

SanyaBratashchuk (changes requested):

src/transformState.js
EhorSaprykin: @SanyaBratashchuk done

alexandra-protyanova (approved):
GJ!

PR #1468, opened 2022-08-20
Solution

IhorChaikovskyi:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+    } else if (oneAction.type === 'removeProperties') {
+      for (const oneArray of oneAction.keysToRemove) {
+        delete state[oneArray];
+      }
+    } else if (oneAction.type === 'clear') {
+      for (const oneState in state) {
+        delete state[oneState];
+      }
+    }
+  }
```
Bohdan-Maksymiuk, line 20: avoid such nesting `else if` conditions
and better use `switch` to manage such conditions

Bohdan-Maksymiuk (changes requested):

valerii-derkach (approved):

PR #1469, opened 2022-08-22
Solution

novak-k:


src/transformState.js
```diff
@@ -4,8 +4,25 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (let i = 0; i < action.keysToRemove.length; i++) {
+        delete state[action.keysToRemove[i]];
+      }
+    }
+
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
```
liliia-livandovskaya, line 24: let's use  switch case instead of if

liliia-livandovskaya (changes requested):
GJ! Let`s make some changes

src/transformState.js
novak-k: done, switch case added instead of 'if'

valerii-derkach (approved):

PR #1474, opened 2022-08-25
Solution

AMahalias:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          for (const name of obj['keysToRemove']) {
+            if (n === name) {
+              delete state[n];
+            }
+          }
+        }
+      }
+
+      if (obj[key] === 'clear') {
+        for (const k in state) {
```
alexandra-protyanova, line 25: use meaningful names for variables

alexandra-protyanova (changes requested):
Let us use `switch case` in this task

marina-tilniak (approved):

PR #1479, opened 2022-08-29
final solution

Quelaan1:
final solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const { type } = actions[i];
+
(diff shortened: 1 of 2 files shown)


kamal-tufekcic (approved):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          }
+        }
+        break;
+
+      default:
+        if (state) {
+          for (const key of Object.keys(state)) {
+            delete state[key];
+          }
+        }
```
VinnieJ-2k20, line 31: It's not a default case, it has its own condition.
Default case can return error message about broken inputs.

VinnieJ-2k20 (changes requested):
Looks good, but one thing needs to be changed,

VinnieJ-2k20 (approved):
Good job 👍

PR #1482, opened 2022-08-30
done

Dmytro-Yankovskyi:


valerii-derkach (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const act in actions) {
```
SanyaBratashchuk, line 8: use `for...of` loop to iterate through array

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
-}
+  for (const operation of actions) {
+    switch (operation.type) {
+      case 'addProperties':
+        Object.assign(state, operation.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property in operation.keysToRemove) {
```
alexandra-protyanova, line 15: don't iterate through array with `for in` loop

alexandra-protyanova (changes requested):

ArturErmolenko (approved):

PR #1485, opened 2022-08-31
Solution

LiubomyrShyshak:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
+    if (act.type === 'addProperties') {
+      for (const [key, value] of Object.entries(act.extraData)) {
+        state[key] = value;
+      }
```
marina-tilniak, line 12: just use object.assign here

marina-tilniak (changes requested):
let's use switch case here

alexandra-protyanova (approved):
Well done!

PR #1486, opened 2022-08-31
solution

olena-hapon:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        break;
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
stanislav-topikha, line 23: image

stanislav-topikha (changes requested):

src/transformState.js
stanislav-topikha: https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md

SanyaBratashchuk (approved):

PR #1491, opened 2022-09-02
Task is finished

YevaDrach:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```
use clear naming for variables

Bohdan-Maksymiuk (approved):

PR #1495, opened 2022-09-05
created

dmitriyn3679:
why i can call arrow func before init in this task?

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
+    switch (object.type) {
+      case 'addProperties':
+        addPropertiesFunc(state, object.extraData);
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+  for (const value of toRemove) {
+    delete state[value];
+  }
+};
+
+const clearFunc = state => {
+  for (const key in state) {
+    delete state[key];
+  }
+};
```
Bohdan-Maksymiuk, line 39: strange to me, mb because of exporting and running in another place

Bohdan-Maksymiuk (approved):

PR #1496, opened 2022-09-05
Solution

MaxManis:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      // For the type of 'removeProperties':
+    } else if (action['type'] === 'removeProperties') {
+      for (const delItem of action['keysToRemove']) {
+        delete state[delItem];
+      }
+      // For the type of 'clear':
+    } else if (action['type'] === 'clear') {
+      for (const delAllItem in state) {
+        delete state[delAllItem];
+      }
```
Anna-Tkachenko, line 24: if you have some logic, that you need to run depending on the value of some variable (action type in this case), will be better to use switch / case

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
 function transformState(state, actions) {
-  // write code here
+  // Main Loop:
+  for (const action of actions) {
+    switch (action['type']) {
+      // For the type of 'addProperties':
+      case 'addProperties':
+        for (const extraItem in action['extraData']) {
+          state[extraItem] = action['extraData'][extraItem];
+        }
```
Denchepornyuk, line 15: use Object.assign for adding new values to the object

Denchepornyuk (changes requested):

PR #1499, opened 2022-09-06
made function transformState

Hlazman:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (iterator.type === 'removeProperties') {
+      for (let i = 0; i <= iterator.keysToRemove.length - 1; i++) {
+        delete state[iterator.keysToRemove[i]];
+      }
+    }
+
+    if (iterator.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
liliia-livandovskaya, line 24: Let's rewrite it using switch case

liliia-livandovskaya (changes requested):
👍  GJ, lets work on changes

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    switch (iterator.type) {
+      case 'addProperties': {
+        for (const key in iterator.extraData) {
+          state[key] = iterator.extraData[key];
+        }
+        break;
+      }
+
+      case 'removeProperties': {
+        for (let i = 0; i <= iterator.keysToRemove.length - 1; i++) {
```
alexandra-protyanova, line 18: it is better to use `for of` loop here

alexandra-protyanova (approved):
Well done!

src/transformState.js
Hlazman: > it is better to use `for of` loop here

Ok. Thanks

PR #1501, opened 2022-09-06
Solution

SergIats:


alexandra-protyanova (changes requested):
Good job, but it is a perfect task to use switch case, let us do it!)

SergIats:
Alexandra, you're right. Switch is cool idea) Simply "if/else" is already always in brain)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    for (const k in actions[i]) {
```
alexandra-protyanova, line 9: `k` isn't really good name for variable

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      switch (actions[i][k]) {
+        case 'addProperties':
+          Object.assign(state, actions[i].extraData);
+          break;
+
+        case 'removeProperties':
+          for (let j = 0; j < actions[i].keysToRemove.length; j++) {
```
alexandra-protyanova, line 16: use for of loop here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+            delete state[actions[i].keysToRemove[j]];
+          };
+          break;
+
+        case 'clear':
+          for (const l in state) {
```
alexandra-protyanova, line 22: use meaningful names for variables

alexandra-protyanova (changes requested):

SergIats:
Thanks for recommendations! I have done)

liliia-livandovskaya (approved):
Great job 🔥

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          break;
+
+        case 'clear':
+          for (const keyName in state) {
+            delete state[keyName];
+          }
+          break;
+      }
+    }
+  }
```
liliia-livandovskaya, line 28: you don't need both for in and for of loops. Use just for of loop

liliia-livandovskaya (changes requested):
GJ, pay attention to the comment

PR #1502, opened 2022-09-08
commit

devWizHub:


amitpatiljc (changes requested):
All good just try to get rid of 1 loop

src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          delete state[actions[i].keysToRemove[j]];
+        }
+        break;
+
+      case 'clear':
+        for (const m in state) {
+          delete state[m];
+        }
+    }
```
amitpatiljc, line 26: add default statement

amitpatiljc, line 26: do not use variable names like `i, j, m ` use meaningful names

src/transformState.js
```diff
@@ -4,8 +4,26 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
   // write code here
-}
+  for (const i in actions) {
```
amitpatiljc, line 10: use `for of `loop here , it works well here

amitpatiljc (changes requested):

amitpatiljc (changes requested):

amitpatiljc (changes requested):
I think still changes are not made

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+          delete state[action.keysToRemove[names]];
+        }
+        break;
+
+      case 'clear':
+        for (const states in state) {
+          delete state[states];
+        }
+        break;
+      default:;
```
amitpatiljc, line 27: don't leave default statement blank
```suggestion
       default:
        break;
```

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
-}
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
 
+      case 'removeProperties':
+        for (const names in action.keysToRemove) {
+          delete state[action.keysToRemove[names]];
```
amitpatiljc, line 18: If you use `for of `loop it will simply look like 

```suggestion
        for (const names of action.keysToRemove) {
          delete state[names];
```

amitpatiljc (changes requested):

amitpatiljc (approved):
Good Job 👍

PR #1503, opened 2022-09-08
js_stateful-object solution

anton-martyniuk1:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
alexandra-protyanova, line 8: it is better to use `for of` loop here

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
+    } else if (actions[i].type === 'removeProperties') {
+      for (const propertie in actions[i].keysToRemove) {
```
alexandra-protyanova, line 12: don't iterate through array with `for in` loop

alexandra-protyanova (changes requested):
Let us use `switch case` here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
SanyaBratashchuk, line 9:
```suggestion
    switch (action.type) {
```

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case ('addProperties'):
+        Object.assign(state, action.extraData);
+        break;
+
+      case ('removeProperties'):
+        for (let i = 0; i < action.keysToRemove.length; i++) {
+          const deletePropertie = action.keysToRemove[i];
```
stanislav-topikha, line 16:
```suggestion
          const propertyToDelete = action.keysToRemove[i];
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+          delete state[deletePropertie];
+        };
+        break;
+
+      case ('clear'):
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
```
stanislav-topikha, line 26: image

stanislav-topikha (changes requested):
Comments ⬇️

valerii-derkach (approved):

PR #1506, opened 2022-09-11
solution

vmamchur:


volodymyrklymkiv (approved):
awesome!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const correctState = state;
+
+  for (const action of actions) {
```
naziamoff, line 10: You should try destructuring of `action` here

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          delete correctState[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in correctState) {
+          delete correctState[key];
+        }
+        break;
+    }
```
naziamoff, line 27: add `default` case for `switch/case`

naziamoff (changes requested):

src/transformState.js
vmamchur: Denys removed this point from the checklist :)
image

naziamoff (approved):

PR #1507, opened 2022-09-11
solution

artem-musii:


nikitaovch99 (approved):
Good job using switch

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        for (const key in actions[i].extraData) {
+          state[key] = actions[i].extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in actions[i].keysToRemove) {
+          delete state[actions[i].keysToRemove[key]];
+        }
+    }
```
naziamoff, line 28: `switch/case` should have `default` case

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const result = state;
+
+  for (let i = 0; i < actions.length; i++) {
```
naziamoff, line 10: if you don't need to work with indexes - you should use `for...of` loop instead

naziamoff (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
naziamoff, line 13: Try some desctructuring, it is really great thing to use
```suggestion
  for (const action of actions) {
  const { type, extraData, keysToRemove } = action;
    switch (type) {
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
```

naziamoff (changes requested):

naziamoff (approved):

PR #1508, opened 2022-09-12
Solution

vladimir-dublya:


myroslav-diiak (approved):

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+      for (const key in state) {
+        if (actions[i].keysToRemove.includes(key)) {
+          delete state[key];
+        }
+      }
+    }
+
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
+    }
```
Denchepornyuk, line 27: it's a good place for switch case, use switch case instead of ifs

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in state) {
+        if (actions[i].keysToRemove.includes(key)) {
```
Denchepornyuk, line 19: you don't need `if` here, if key doesn't exist in object delete will ignore this key

Denchepornyuk (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in state) {
+          if (actions[i].keysToRemove.includes(key)) {
+            delete state[key];
+          }
```
Denchepornyuk, line 22: you don't need this if here, if key doesn't exist in the object then delete will ignore this key

Denchepornyuk (approved):

PR #1510, opened 2022-09-12
solution

myroslav-diiak:


vladimir-dublya (approved):

Denchepornyuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+            delete state[property];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
Denchepornyuk, line 28: you don't need hasOwnProperty here, if key doesn't exist in the object then delete will ignore this key

Denchepornyuk (approved):

PR #1512, opened 2022-09-12
solution

oleksii-bieliaiev:


ArturErmolenko (changes requested):
[CODE STYLE]: switch/case should always have default case for error handling.

oleksii-bieliaiev:
default has been added

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default: return 'Error';
```
SanyaBratashchuk, line 24: Throw errors

SanyaBratashchuk (changes requested):

src/transformState.js
oleksii-bieliaiev: done

valerii-derkach (approved):

PR #1513, opened 2022-09-12
solution

volodymyrklymkiv:


src/transformState.js
```diff
@@ -6,6 +6,26 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  const obj = state;
```
vmamchur, line 9: i don't know if that's a good name for the variable

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+      for (const remove of key.keysToRemove) {
+        delete obj[remove];
+      }
+    }
 
+    if (key.type === 'clear') {
+      for (const properties in state) {
+        delete obj[properties];
+      }
+    }
```
vmamchur, line 26: it'll be better to use `switch / case` instead of `if` here =)
image

vmamchur (changes requested):

vmamchur (approved):
gj

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+        for (const remove of action.keysToRemove) {
+          delete copyState[remove];
+        }
+        break;
 
+      case 'clear':
+        for (const properties in state) {
+          delete copyState[properties];
+        }
+    }
```
naziamoff, line 27: add `default` case for switch/case

naziamoff (approved):

PR #1515, opened 2022-09-12
Solution

Andrew-Skoryk:
Solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
anastasiia-nahaiska, line 9: You can write switch (action.type) and then in cases, you can write only 'removeProperties', etc.

anastasiia-nahaiska (commented):
GJ!

src/transformState.js
Andr1yK, line 9: Why u use the switch with `true` and in every case, u write `action.type`?

Andr1yK (changes requested):

PR #1519, opened 2022-09-12
Solution

Apollonia5:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action['type'] === 'addProperties') {
```
Yevheniya-Deryka, line 9: You can just use `action.type`

Yevheniya-Deryka (approved):
GJ!
Here is the same suggestion as in Stateful clones

src/transformState.js
Andr1yK: and use switch case 🤩

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      const additionalObj = action['extraData'];
+
+      for (const objKey in additionalObj) {
+        state[objKey] = additionalObj[objKey];
+      }
+    }
+
+    if (action['type'] === 'removeProperties') {
+      const arr = action['keysToRemove'];
```
Andr1yK, line 18: What is arr?)

Andr1yK (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        const additionalObj = action['extraData'];
+
+        for (const objKey in additionalObj) {
+          state[objKey] = additionalObj[objKey];
+        }
```
Andr1yK, line 15: just use Object.assign

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+        break;
+
+      case 'removeProperties':
+        const arr = action['keysToRemove'];
+
+        for (const arrElem of arr) {
```
Andr1yK, line 22:
```suggestion
        for (const key of arr) {
```

Andr1yK (approved):

PR #1520, opened 2022-09-12
Solution

glugovsky:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    switch (actions[action].type) {
```
Anna-Borodaienko, line 9: Why do you use 'in' in this loop? Actions is not an object, it's an array. If you use action OF actions, you can use later  'action', not 'actions[action]'

glugovsky: Good point, thank you)

Vladosik1337 (commented):
GJ)

Denchepornyuk (approved):

PR #1521, opened 2022-09-12
Solution

Yevheniya-Deryka:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const el of actions) {
+    switch (el.type) {
```
Anna-Borodaienko, line 9: Maybe it would be better not to use names like 'el' or 'element' - these names are not very semantic

Yevheniya-Deryka: Agree

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => {
```
Denchepornyuk, line 19: use `for in` loop here instead of `Object.keys`

Denchepornyuk (approved):

PR #1523, opened 2022-09-12
solution

ihorandrianov:


korniienkokarina (commented):
good job! 👏

src/transformState.js
```diff
@@ -1,11 +1,63 @@
...
+          return p;
+        }, _state);
+
+        return updatedState;
+      case 'clear':
+        return {};
+      default:
+        break;
+    }
+  };
```
Denchepornyuk, line 55: remove comments

src/transformState.js
```diff
@@ -1,11 +1,63 @@
+/* eslint-disable spaced-comment */
 'use strict';
 
 /**
  * @param {Object} state
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const reducer = (_state, action) => {
```
Denchepornyuk, line 9: why do you use `_` for `state`?
It's a normal argument for the faction, don't use `_`

src/transformState.js
```diff
@@ -1,11 +1,63 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete _state[key];
+        }
+        break;
+      case 'clear':
+        const objectProps = Object.getOwnPropertyNames(_state);
+
+        objectProps.forEach((key) => delete _state[key]);
+        break;
```
Denchepornyuk, line 25: you can use `for in` instead of getOwnPropertyNames and forEach

denys-cheporniuk (changes requested):

src/transformState.js
ihorandrianov: @Denchepornyuk Its a "no-shadowing" rule in ESlinter

src/transformState.js
```diff
@@ -1,11 +1,36 @@
...
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const reducer = (reducerState, action) => {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          reducerState[key] = action.extraData[key];
+        }
```
Denchepornyuk, line 14: use Object.assign for adding new values to the object

denys-cheporniuk (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,34 @@
...
+        break;
+      case 'clear':
+        for (const param in reducerState) {
+          delete reducerState[param];
+        }
+        break;
+      default:
+        break;
+    }
+  };
```
OleksiiBlezniuk, line 27: This helper function also can be moved outside `transformState` function in order to make it reusable

OleksiiBlezniuk (approved):
Good job!

PR #1525, opened 2022-09-12
Solution

imaksweb:


src/transformState.js
```diff
@@ -5,7 +5,43 @@
...
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
```
Anna-Borodaienko, line 21: 2. [CODE STYLE]: Use `switch` statement if you have limited amount of conditions.
Checklist )

Anna-Borodaienko (approved):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
```
naziamoff, line 8: You should try destructuring of `action` in `for..of` loop

naziamoff (changes requested):

src/transformState.js
imaksweb: > You should try destructuring of `action` in `for..of` loop

Yeah, but depending on `type` we need to use a different property. If type is 'addProperties' we use `extraData` and if 'removeProperties' - `keysToRemove`. So if we use destructuring we always have extra variable.

naziamoff: agreed!

naziamoff (approved):

PR #1526, opened 2022-09-12
Transforming state

Vladosik1337:


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+          delete state[deleteValue];
+        }
+        break;
+
+      case 'clear' :
+        for (const deleteAll in state) {
+          delete state[deleteAll];
+        }
+        break;
+    }
```
glugovsky, line 28: switch/case should always have default case for error handling
```suggestion
    default:
        throw new Error('Wrong type!');
    }
```

glu3s (approved):
GJ!

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+  for (const objects of actions) {
+    const { type, extraData, keysToRemove } = objects;
+
+    switch (type) {
+      case 'addProperties' :
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const deleteValue of keysToRemove) {
```
naziamoff, line 18: This would make your code clearer
```suggestion
        for (const keyToRemove of keysToRemove) {
```

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+          delete state[deleteValue];
+        }
+        break;
+
+      case 'clear' :
+        for (const deleteAll in state) {
```
naziamoff, line 24: delete all would be a good name for a function, but not for a variable.
It has to be a noun and it should describe what is stored inside

naziamoff (changes requested):

naziamoff (approved):
GJ!

PR #1527, opened 2022-09-12
solution

artem-nasadchuk:


anton-iskryk (approved):
gj!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
naziamoff, line 25: you should add default case for switch/case

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
naziamoff, line 8: try destructuring `action`

naziamoff (changes requested):

src/transformState.js
artem-nasadchuk: I don't agree with you, it is not a good case here, we discussed it with Denis in the morning lesson today, and he removed this point from a checklist

naziamoff: You are right, approved!

naziamoff (approved):

PR #1528, opened 2022-09-12
Solution

iPanda177:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action['type']) {
```
iam-datsenko, line 9: U could use just action.type

iam-datsenko (approved):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'removeProperties':
+        for (const delKey of action['keysToRemove']) {
+          delete state[delKey];
+        }
+        break;
+
+      case 'clear':
+        for (const delKeys in state) {
+          delete state[delKeys];
+        }
```
Denchepornyuk, line 23:
```suggestion
        for (const key of action['keysToRemove']) {
          delete state[delKey];
        }
        break;

      case 'clear':
        for (const keys in state) {
          delete state[delKeys];
        }
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action['type']) {
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
Denchepornyuk, line 11: agree with @iam-datsenko, use action.extraData, fix in all places

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1529, opened 2022-09-12
Solution

anastasiia-nahaiska:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
naziamoff, line 8: try destructuring `action` on the next line

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        const arrKeysToRemove = action.keysToRemove;
```
naziamoff, line 21: arrKeysToRemove is unnecessary here, feel free to remove it and use `action.keysToRemove` instead
P.S. if you find a way to desctructure `action` - it will be just `keysToRemove`

naziamoff (changes requested):

src/transformState.js
naziamoff: I have just discussed it with Denys, you are right 👍

naziamoff (approved):
🔥

PR #1530, opened 2022-09-12
Solution v01

anton-iskryk:


artem-nasadchuk (approved):

naziamoff (changes requested):
Good job 🔥 
Add a `default` statement for switch/case

naziamoff (approved):

PR #1531, opened 2022-09-12
Solution

ArturNagaycev:


jojnajbude (approved):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
naziamoff, line 8: You should practice destructuring here

naziamoff (changes requested):
Well done, but with destructuring it wil be even better

src/transformState.js
naziamoff: discussed with Denys, agreed that destructuring should not be used here.
PR approved!

naziamoff (approved):

PR #1533, opened 2022-09-12
Solution

Rina07:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Oleksandra-Shevchenko, line 10: you need to use switch  because of checklist :)

Oleksandra-Shevchenko (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,30 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw Error('unknown action type');
+    }
```
OleksiiBlezniuk, line 29: This logic can be moved to separate function

OleksiiBlezniuk (approved):
Good job!

PR #1537, opened 2022-09-12
Solution

Dikey945:


illia-sosnytskyi (commented):
GJ !

src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const obj of actions) {
```
naziamoff, line 9: You should take a look at the checklist again
We need to use proper naming in `for...of` loop

src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+    if (obj['type'] === 'addProperties') {
```
naziamoff, line 10: it is better to use switch/case if you have limited amount of conditions.
It was on the checklist too

naziamoff (changes requested):

src/transformState.js
Dikey945: Done

src/transformState.js
```diff
@@ -6,6 +6,28 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
naziamoff, line 12: if you know the exact name of the variable you may use `.` to access the property
```suggestion
        Object.assign(state, action.extraData);
```

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action['keysToRemove']) {
```
naziamoff, line 16: same here

naziamoff (changes requested):

src/transformState.js
Dikey945: done

src/transformState.js
Dikey945: done

naziamoff (approved):
Gj

PR #1538, opened 2022-09-13
State transformation

Yuriy-Komolov:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
DenysKleimenov, line 27: switch/case should always have a default case for error handling.
So you should add something to let a user know about an error.

DenysKleimenov (changes requested):

naziamoff (approved):
Well done 🔥

PR #1541, opened 2022-09-13
Solution

OleksandrGlulhovskiy:


Andrew-Skoryk (changes requested):
I feel like we have to use switch statements here
2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const k in state) {
```
Denchepornyuk, line 11: `k` isn't a good name for variable

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[k];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in state) {
+          if (action.keysToRemove.includes(key)) {
+            delete state[key];
+          }
```
Denchepornyuk, line 20: you don't need this if here

Denchepornyuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+    }
```
OleksiiBlezniuk, line 25: switch/case should have default case, you can use it to handle cases when action type is different from these

OleksiiBlezniuk (changes requested):

Andr1yK (approved):

PR #1544, opened 2022-09-13
solution

mayuces:


Quelaan1 (commented):
if statement should not be in default.

Quelaan1 (changes requested):
if statement should not be in default.

Quelaan1 (approved):
Good job.

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
```
VinnieJ-2k20, line 19: Remove the check, it can work incorrectly if value is falsy.
Delete works fine with non-existent properties.

VinnieJ-2k20 (changes requested):
Looks good, one thing needs to be fixed.

VinnieJ-2k20 (approved):

ihor-jpeg (approved):

PR #1545, opened 2022-09-13
add js code who decides task

SerhiiRomaniukWD:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          if (Object.keys(state).filter(elem => elem === property)) {
+            delete state[property];
+          }
```
Anna-Borodaienko, line 18: You don't need to check if this property exists, method delete do it by itself
```suggestion
        for (const property of action.keysToRemove) {
                    delete state[property];
        
```

You don't need to check if this property exists, method delete do it by itself

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
```
Anna-Borodaienko, line 25:
```suggestion
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
```

SerhiiRomaniukWD: checklist has a rule 'don't use for in loop for iterating over array'

src/transformState.js
SerhiiRomaniukWD: 100%!You are right, thanks

Andr1yK (approved):

PR #1546, opened 2022-09-13
js_stateful-object

xiimiik:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties': {
+        Object.assign(state, action.extraData);
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
```
avkovlit, line 16: also you can use more information in variable name keys,  will be better to name deleteKey for example

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          delete state[key];
+        }
+        break;
+      }
+
+      case 'clear': {
+        for (const key in state) {
```
avkovlit, line 23: and here too

avkovlit (approved):
Well done! I left some comments, i think it will better to fix it

denys-cheporniuk (approved):

PR #1548, opened 2022-09-13
Solution

avkovlit:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action['type']) {
```
xiimiik, line 9: you just should use dot here and in the code below
`switch (action.type)`

xiimiik (changes requested):
GJ, left some comments 👍

src/transformState.js
avkovlit: done!

xiimiik (approved):
Great! GJ 🔥

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const delProp of action.keysToRemove) {
```
naziamoff, line 15: you should try to make your code maximally clear
```suggestion
        for (const propToDelete of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[delProp];
+        }
+        break;
+
+      case 'clear':
+        for (const delAllProp in state) {
```
naziamoff, line 21: same here, you should use proper names for variables. The name of the variable should tell what is stored inside

naziamoff (changes requested):
Well done!
There are just some suggestions on improving naming

naziamoff (approved):
Good job, but check my comment pls. Naming is really important

PR #1549, opened 2022-09-13
Solution

SanyaKovalenko:


MAKSOti (changes requested):
Well done!)
Try to make switch method solution (watch №2 in checklist)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        };
```
Anna-Tkachenko, line 13: you can also use Object.assign here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
```
Anna-Tkachenko, line 26: add default option (check point 4 in checklist)

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):

PR #1552, opened 2022-09-13
solution

NickBrodyuk:


src/transformState.js
```diff
@@ -8,20 +8,26 @@ function transformState(state, actions) {
...
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
     }
   }
```
Ihor-1993, line 33: GJ!!

Ihor-1993 (approved):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const item of actions) {
+    let itemType;
+
+    switch (true) {
+      case item.type === 'addProperties':
```
Denchepornyuk, line 12: add item.type to the switch instead of true and remove === from cases

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
Denchepornyuk, line 31: add clear case, this code shouldn't be in the default

denys-cheporniuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const item of actions) {
+    let itemType;
```
OleksiiBlezniuk, line 9: Why do you need this variable? It looks strange

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+    switch (item.type) {
+      case 'addProperties':
+        itemType = item.extraData;
+
+        Object.assign(state, itemType);
+        break;
+
+      case 'removeProperties':
+        itemType = item.keysToRemove;
```
OleksiiBlezniuk, line 19: Better to use this object property instead of variable
(you didn't use it outside switch/case, you store different values in it)
It's harder to read your code with it

OleksiiBlezniuk (changes requested):

nazarmatsevych (approved):

PR #1554, opened 2022-09-14
transformState

Arsen-Shpak:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
```
AndriyZhabynets, line 25: You can add a default value after this case for errors

AndriyZhabynets (approved):
Good job!

Andr1yK (changes requested):
[point 4](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)

Andr1yK (approved):

PR #1556, opened 2022-09-14
result

AndriyZhabynets:


Arsen-Shpak (approved):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const states = state;
```
Andr1yK, line 8: Why do u need this?

Andr1yK (changes requested):

src/transformState.js
AndriyZhabynets: You're right, it's unnecessary

Andr1yK (approved):

PR #1562, opened 2022-09-14
Solution stateful object

Ihor-1993:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const item of actions) {
+    let itemType;
+
+    switch (true) {
+      case item.type === 'addProperties':
```
Denchepornyuk, line 12: add `item.type` to the switch instead of `true` and remove `===` from cases

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        for (const key of itemType) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Denchepornyuk, line 30: add `clear` case, this code shouldn't be in the default

Denchepornyuk (changes requested):

Denchepornyuk (approved):

Anna-Tkachenko (approved):

PR #1563, opened 2022-09-14
State__solution

VovaDzhuryk:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
+    }
```
OleksiiBlezniuk, line 25: What if action.type will have other value?

switch/case construction should have `default` case (you can just use `break` there)

OleksiiBlezniuk (changes requested):

OleksiiBlezniuk (approved):
Create more meaningful commit messages, it’s hard to understand what was done in this commit from `StatefulObject_fixed` message.

PR #1564, opened 2022-09-14
function transformState was added

Olena-Shevchenko:


ArturErmolenko (changes requested):
Use switch case here is [Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)

liliia-livandovskaya (approved):
Great job!

PR #1571, opened 2022-09-15
Solution

AleksanderPl:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      for (const property of keysToRemove) {
+        delete state[property];
+      }
+    }
+
+    if (type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
marina-tilniak, line 29: let's use switch case here also

marina-tilniak (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+      case 'removeProperties':
+        // delete only several properties what`s in type: "removeProperties"
+        for (const key of act.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        // delete all
+        for (const key of Object.keys(state)) {
```
ivanchuk-yurii, line 24: `for in` was better, bc it's faster

ivanchuk-yurii (approved):

PR #1572, opened 2022-09-15
Solution

GizemEce:


Quelaan1 (approved):
A switch case can also be used.

ihor-jpeg (changes requested):
use switch case for `type`

ihor-jpeg (approved):

PR #1575, opened 2022-09-15
Solution

MAKSOti:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[clearedKey];
+        }
+        break;
+
+      default:
+        throw Error('Unknown action type');
+    }
+  }
+
+  return state;
```
DenysKleimenov, line 31: I think it should work without this return, but it's ok!

DenysKleimenov (approved):
GJ!

Anna-Tkachenko (approved):
good job! ⭐️

PR #1577, opened 2022-09-16
Solition

Fispil:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const value in state) {
+          delete state[value];
+        }
+        break;
+    }
+  }
```
SanyaBratashchuk, line 26: destructure action object

SanyaBratashchuk (changes requested):

src/transformState.js
Bohdan-Maksymiuk: you do not use destructuring
for example
```
  const {
    type,
    extraData,
    keysToRemove,
  } = action;
```

Bohdan-Maksymiuk (changes requested):
GJ, but fix comment

Bohdan-Maksymiuk (approved):
Good job 👍

PR #1579, opened 2022-09-16
solution for Stateful object

sapnachoudhary06:


src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+        }
+        break;
+
+      case actionRemoveProperties:
+        for (const property of action.keysToRemove) {
+          delete transformed[property];
+        }
+        break;
+
+      default:
```
kamal-tufekcic, line 32: Checklist item # 4

kamal-tufekcic (changes requested):

src/transformState.js
sapnachoudhary06: Done.

ihor-jpeg (approved):

PR #1582, opened 2022-09-17
solution

La-La-Op:


MAKSOti (changes requested):
GJ
But you need to read the checklist
[CODE STYLE]: Use switch statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const obj of actions) {
+    const { type, extraData, keysToRemove } = obj;
```
OleksiiBlezniuk, line 9: 👍

OleksiiBlezniuk (approved):

PR #1583, opened 2022-09-17
Solution for stateful object

petrovskyi-yevhen:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```

Bohdan-Maksymiuk: use meaningful names for variables

Bohdan-Maksymiuk (approved):

src/transformState.js
petrovskyi-yevhen: done

src/transformState.js
```diff
@@ -1,11 +1,28 @@
 'use strict';
 
 /**
- * @param {Object} state
- * @param {Object[]} actions
+ * @param {action} state
+ * @param {action[]} actions
```
Bohdan-Maksymiuk, line 5: you shouldn't change smth there - it is just described types of arguments

Bohdan-Maksymiuk (approved):

PR #1584, opened 2022-09-18
Solution

tatyana22:


MAKSOti (approved):
GJ

src/transformState.js
```diff
@@ -6,6 +6,33 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  const nextActions = [];
+
+  for (const key1 of actions) {
```
imaksweb, line 11:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -6,6 +6,33 @@
...
+    switch (key1.type) {
+      case 'addProperties':
+        Object.assign(state, key1.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key2 of key1.keysToRemove) {
```
imaksweb, line 18:
```suggestion
        for (const property of key1.keysToRemove) {
```

src/transformState.js
```diff
@@ -6,6 +6,33 @@
...
+          delete state[key2];
+        }
+        break;
+
+      case 'clear':
+        for (const key3 in state) {
```
imaksweb, line 24:
```suggestion
        for (const property in state) {
```

src/transformState.js
```diff
@@ -6,6 +6,33 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  const nextActions = [];
+
+  for (const key1 of actions) {
+    switch (key1.type) {
+      case 'addProperties':
+        Object.assign(state, key1.extraData);
```
imaksweb: Don't mutate object or arrays - it will cause unexpected results later on. You should make copy using `Object.assign` or `spread operator`. You mutate your `state`, you should make a copy of `state` and mutate a copy. All input parameters shouldn't be changed.

imaksweb (approved):
GJ 👍

silchyn (approved):
Cool!

PR #1588, opened 2022-09-19
solved

Deniso4ka23:


src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
+      for (let k = 0; k < action[i].keysToRemove.length; k++) {
+        delete state[action[i].keysToRemove[k]];
+      }
+    }
+
+    if (action[i].type === 'clear') {
+      for (const keys in state) {
+        delete state[keys];
+      }
+    }
```
Bohdan-Maksymiuk, line 23: use switch for such defined cases

src/transformState.js
```diff
@@ -4,8 +4,26 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function transformState(state, action) {
+  for (let i = 0; i < action.length; i++) {
```
Bohdan-Maksymiuk, line 8: better to use `for of` loop - you don't need indexes here

Bohdan-Maksymiuk (changes requested):

PR #1589, opened 2022-09-19
commit1

Gintako666:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (const key in actions) {
```
alexandra-protyanova, line 10: do not use for in loop to iterate through array

alexandra-protyanova (changes requested):
Let us switch in this task

alexandra-protyanova (changes requested):
U can use for of loop here & I have asked about switch case)

alexandra-protyanova (approved):

PR #1590, opened 2022-09-19
js_stateful-object

KalmykovMikhail2:


alexandra-protyanova (changes requested):
let us use switch here

alexandra-protyanova (approved):

PR #1591, opened 2022-09-19
Implement transformState function

anastasiia-tilikina:


src/transformState.js
```diff
@@ -5,7 +5,40 @@
...
+        addProperties(state, action.extraData);
+        break;
+
+      case action.type === 'removeProperties':
+        removeProperties(state, action.keysToRemove);
+        break;
+
+      default:
+        clearObj(state);
+    }
```
alexandra-protyanova, line 20:
```suggestion
    switch (action.type ) {
      case 'addProperties':
        addProperties(state, action.extraData);
        break;

      case 'removeProperties':
        removeProperties(state, action.keysToRemove);
        break;

      default:
        clearObj(state);
    }
```

src/transformState.js
```diff
@@ -5,7 +5,40 @@
...
+  }
+}
+
+function addProperties(obj, data) {
+  Object.assign(obj, data);
+}
+
+function removeProperties(obj, properies) {
+  for (const prop of properies) {
```
alexandra-protyanova, line 29: in my opinion it is a bit overkilling to create separate functions in this case

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):

PR #1592, opened 2022-09-19
Solution for transformState

Andrii-Medintsev:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      }
+    } else if (obj.type === 'removeProperties') {
+      for (const elem of obj.keysToRemove) {
+        delete state[elem];
+      }
+    } else if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
liliia-livandovskaya, line 21: lets use switch case instead if if else there

liliia-livandovskaya (changes requested):
Great job! Lets work on changes

PR #1596, opened 2022-09-21
Solution

vovchuk-nazar:


silchyn (changes requested):
Read the checklist

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.getOwnPropertyNames(state)) {
+          delete state[key];
+        }
```
naziamoff, line 23: `for...in` loop will do better here. And you won't need getOwnPropertyNames here if you use it

naziamoff (changes requested):

PR #1604, opened 2022-09-26
Develop

Dmytro-Kravchenko:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          }
+        }
+        break;
+
+      default:
+        const objKeys = Object.keys(state);
+
+        objKeys.forEach(key => {
+          delete state[key];
+        });
```
stanislav-topikha, line 27: It should not be a 'default' case
Simple loop will do it in one iteration, in your case its two

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
```
stanislav-topikha, line 15:
```suggestion
        for (const key of action.keysToRemove) {
```

```
name : 5 => property(key/value pair)
name => key or property name
5 => value or property value
```

stanislav-topikha (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        } break;
```
SanyaBratashchuk, line 23:
```suggestion
        for (const key in state) {
          delete state[key];
        } 
        break;
```

SanyaBratashchuk (approved):

PR #1618, opened 2022-10-04
Solution

Julia2063:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const property of action.keysToRemove) {
+        delete state[property];
+      }
+    } else {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
liliia-livandovskaya, line 19: lets use switch case there

liliia-livandovskaya (changes requested):
Let's work on changes

liliia-livandovskaya (approved):
Great job! 🔥

PR #1625, opened 2022-10-11
solution

mykola8426:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+        state[key] = action.extraData[key];
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    } else {
+      for (const key in state) {
+        delete state[key];
+      }
```
ArtemTopolskyi, line 20: switch-case will be better, then if in this task

ArtemTopolskyi (approved):

PR #1628, opened 2022-10-14
Develop

haduigon:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+
+  Object.assign(state, state);
```
TarasHoliuk, line 9: What does this line do?

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+  for (const action in actions) {
+    switch (actions[action].type) {
```
AlisaVlasova, line 12: use destructuring

AlisaVlasova (changes requested):

src/transformState.js
TarasHoliuk, line 12: would be nice to create a variable for `actions[action]` - code will be more readable in this case

TarasHoliuk (changes requested):

src/transformState.js
haduigon: I cannot understand how to use destruction here properly

src/transformState.js
haduigon: deleted )

src/transformState.js
haduigon: done

AlisaVlasova (approved):

PR #1630, opened 2022-10-15
solution

Roman-Malysh:


loralevitska (changes requested):
try to complete this task avoiding nested code.
Use helper functions, variables or objects where the key can be a variable and the value a function (careful, you should use the default value), also use google to find better solution for your case. For example https://www.linkedin.com/pulse/avoid-if-else-hell-javascript-afsal-m-h/

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const i in action.keysToRemove) {
+          delete state[action.keysToRemove[i]];
+        }
```
tetiana-ivanchenko, line 19:
```suggestion
        for (const key of action.keysToRemove) {
          delete state[key];
        }
```

tetiana-ivanchenko: this is not a mistake, but "for of" looks more concise

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      case 'clear':
+        for (const u in state) {
+          delete state[u];
```
tetiana-ivanchenko, line 24: please, use more explicit naming for variables, instead of "u"

src/transformState.js
tetiana-ivanchenko: and please, use more explicit naming for variables, instead of "i"

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #1631, opened 2022-10-16
transformState function impl.

divertitoS:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case action.type === 'removeProperties':
+        removeProperties(state, action.keysToRemove);
+        break;
+
+      case action.type === 'clear':
+        removeProperties(state, Object.keys(state));
+        break;
+    }
```
Bohdan-Maksymiuk, line 21:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case action.type === 'removeProperties':
        removeProperties(state, action.keysToRemove);
        break;

      case action.type === 'clear':
        removeProperties(state, Object.keys(state));
        break;
    }
```
check appropriate usage of switch and fix all cases as I show

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (approved):

PR #1632, opened 2022-10-17
Solution

natalia-maryshko:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+
+    if (obj.type === 'removeProperties') {
+      for (const del of obj.keysToRemove) {
+        delete state[del];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const el in state) {
+        delete state[el];
```
polosanya, line 21: It is better to use [switch...case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) to go through all possible types

polosanya (changes requested):
Nice start! Please check my comments, I think they will help you to improve your code

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const del of obj.keysToRemove) {
+          delete state[del];
+        }
+        break;
+
+      case 'clear':
+        for (const el in state) {
+          delete state[el];
+        }
```
polosanya, line 23: Don't forget to add default case for your switch

polosanya (approved):
Good job! I'm approving it, but please check my comment

PR #1633, opened 2022-10-17
Task was done

Sergey-Mironenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      Object.assign(state, actions[i].extraData);
+    } else if (actions[i].type === 'removeProperties') {
+      for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+        for (let k = 0; k < Object.keys(state).length; k++) {
+          if (actions[i].keysToRemove[j] === Object.keys(state)[k]) {
+            delete state[Object.keys(state)[k]];
+          }
+        }
+      }
+    } else {
```
tetiana-ivanchenko, line 19: please try to avoid large nesting of loops, the more nesting, the more complex the code.
You should create inside "else if" only one loop and that will be enough.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
tetiana-ivanchenko, line 9: try to use switch case block inslead of if, else if, else

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):

PR #1634, opened 2022-10-18
solution

MaxFedak:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    }
+
+    if (action.type === 'addProperties') {
+      for (const [key, value] of Object.entries(action.extraData)) {
+        state[key] = value;
+      }
+    }
```
Kry-Gin, line 26: Check list point 2 > [CODE STYLE]: Use switch statement if you have limited amount of conditions.

Kry-Gin (approved):

src/transformState.js
MaxFedak: oh, didn't noticed there is checklist - thx

src/transformState.js
```diff
@@ -7,22 +7,23 @@
...
 
-    if (action.type === 'addProperties') {
-      for (const [key, value] of Object.entries(action.extraData)) {
-        state[key] = value;
-      }
+      case 'addProperties':
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
     }
```
Kry-Gin, line 27: Check list point 4 > [CODE STYLE]: switch/case should always have default case for error handling.

src/transformState.js
Kry-Gin: Just like me at the beginning ^^*

src/transformState.js
```diff
@@ -6,6 +6,32 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
```
Olexandr6, line 14: 🔥

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      case 'addProperties':
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
```
Olexandr6, line 26: Let's find a more straightforward way to add props to an existing object without using any loops

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+        break;
+
+      default:
+        return 'Something went wrong';
```
Olexandr6, line 30: Let's throw an error here instead.

Olexandr6 (changes requested):
Your solution is good but not great🙂

src/transformState.js
```diff
@@ -6,6 +6,30 @@
...
+          delete state[key];
+        });
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      default:
+        throw new Error('Smth went wrong');
```
Olexandr6, line 28: This error isn't informative

Olexandr6 (approved):

PR #1635, opened 2022-10-18
Solution

Kry-Gin:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const deleteKey of action.keysToRemove) {
```
Olexandr6, line 16:
```suggestion
        for (const keyToDelete of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case 'clear':
+        for (const clear in state) {
+          delete state[clear];
+        }
+
+        break;
+
+      default:
+        break;
```
Olexandr6, line 30: Let's throw an error in case of `default` case

Olexandr6: Read more - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

Olexandr6 (changes requested):
Your solution is good, but not great)

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case 'clear':
+        for (const clear in state) {
+          delete state[clear];
+        }
+
+        break;
+
+      default:
+        throw Error;
```
Olexandr6, line 30: This error isn't informative

Olexandr6, line 30:
```suggestion
        throw new Error(`Unknow type of action: ${action.type}`);
```

Olexandr6 (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case 'removeProperties':
+        for (const keyToDelete of action.keysToRemove) {
+          delete state[keyToDelete];
+        }
+
+        break;
+
+      case 'clear':
+        for (const clear in state) {
```
denys-danyliuk, line 23:
```suggestion
        for (const key in state) {
```

PR #1636, opened 2022-10-18
Solution

Illia-Skladnik:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          if (state.hasOwnProperty(property)) {
+            delete state[property];
+          }
+        }
```
iryna-lelyk, line 19: I think here is too much nesting, but I'm not might be wrong. Checklist, point 5.

iryna-lelyk (approved):
Great solution!

mykola-koval (approved):

PR #1637, opened 2022-10-18
solution Masha Girya

masha-girya:


src/transformState.js
```diff
@@ -6,6 +6,34 @@
...
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
+
+      default:
+        break;
```
mitattt, line 34:
```suggestion
      default:
        return '...';
```
i think you should add something in this case

src/transformState.js
```diff
@@ -6,6 +6,34 @@
...
+        Object.assign(state, actions[i].extraData);
+        break;
+
+      case 'removeProperties':
+        const removeValues = actions[i].keysToRemove;
+
+        for (const toRemove of removeValues) {
+          for (const keys in state) {
+            if (toRemove === keys) {
+              delete state[keys];
```
mitattt, line 21: 5.[CODE STYLE]: Nested loops === EVIL ☠️☠️☠️

src/transformState.js
```diff
@@ -6,6 +6,34 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
```
mitattt, line 9:
```suggestion
 for (const action of actions)  {
```  
imho this will be better

mitattt (changes requested):
GJ! 
just a few mistakes
👇

src/transformState.js
masha-girya: I like them)))))))) i really did not know what to add instead, but when i was reviewing your code i understood

src/transformState.js
masha-girya: yes, thank you!

mitattt (changes requested):
👍

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
+
+      default:
+        return 'Something went wrong';
```
Olexandr6, line 28: It's not an informative error)

Olexandr6: Let's throw an error here instead.

Olexandr6 (changes requested):
Almost done👍 Let's figure out what is `new Error` in JavaScript

Olexandr6 (approved):

PR #1638, opened 2022-10-18
Solution

FallenMAD:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        for (const key of actionObj.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const element in state) {
+          if (state.hasOwnProperty(element)) {
+            delete state[element];
+          }
```
loralevitska, line 24: you sort out all keys of the object, they are in this object

```suggestion
          delete state[element];
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+    }
+  }
+
+  return state;
```
loralevitska, line 29: unexpected behavior, the method should only modify external state and all

```suggestion
```

loralevitska (approved):
great!

PR #1640, opened 2022-10-18
Solution

yaro-bey:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const removeKey of action.keysToRemove) {
+          delete state[removeKey];
+        }
+        break;
+
+      case 'clear':
+        for (const clear in state) {
```
Anna-Shapovalova, line 22: It will be better to rename the variable here, I think

Anna-Shapovalova (approved):
Well done))

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error();
```
mykola-koval, line 28:
```suggestion
        throw new Error('error description');
```

mykola-koval (approved):

PR #1646, opened 2022-10-19
Solution for stateful object

iryna-lelyk:


Illia-Skladnik (approved):
Great work!

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete state[property];
+        };
+        break;
```
Illia-Skladnik, line 18: A question for **mentor.** Should we use '};' or '}' in cases like this?

mykola-koval, line 18:
```suggestion
      case 'removeProperties':
        for (const property of action.keysToRemove) {
          delete state[property];
        };
        
        break;
```

mykola-koval (approved):

PR #1647, opened 2022-10-19
Solution

andrudenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
DarkMistyRoom, line 8: You don't need index, so better use for of

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+        const removeKey = actions[i].keysToRemove[j];
+
+        delete state[removeKey];
+      }
```
DarkMistyRoom, line 24: Better use forEach here. It will be shorter and more readable

DarkMistyRoom (changes requested):
image

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const prop of action.keysToRemove) {
+        delete state[prop];
+      }
+    }
```
SanyaBratashchuk, line 23: use `switch case`

SanyaBratashchuk (changes requested):

stanislav-topikha (approved):
GJ 👍🏻

PR #1649, opened 2022-10-19
solution

mitattt:


masha-girya (approved):
Great job!!

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const keyRemove of action.keysToRemove) {
```
Olexandr6, line 23:
```suggestion
        for (const keyToRemove of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[keyRemove];
+        };
+
+        break;
+
+      default:
+        return 'Error';
+    }
```
Olexandr6, line 31: Let's throw an error here instead.

G-YaSha (changes requested):
Your solution is good but not great🙂

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        };
+
+        break;
+
+      default:
+        throw new Error('Error');
```
Olexandr6, line 30: This error isn't informative

G-YaSha (approved):

PR #1650, opened 2022-10-19
stateful object solution

Svitlana-Yudina:
I couldn't figure out how to solve the 5th item of the checklist.

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    for (const prop in obj) {
+      if (obj[prop] === 'addProperties') {
+        Object.assign(state, obj.extraData);
+      }
+
+      if (obj[prop] === 'removeProperties') {
+        for (const ch of obj.keysToRemove) {
+          if (ch in state) {
+            delete state[ch];
+          }
+        }
+      }
+
+      if (obj[prop] === 'clear') {
+        for (const del in state) {
+          delete state[del];
+        }
+      }
+    }
+  }
+
+  return state;
 }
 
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
-  // write code here
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        Object.assign(state, obj.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const ch of obj.keysToRemove) {
+          if (ch in state) {
```
andreyolin, line 16: i think you don`t need 'if'' here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        Object.assign(state, obj.extraData);
```
andreyolin, line 11: maybe you can write it in one row?

andreyolin (approved):
gj! look at some comments maybe it will help you to make code better)

src/transformState.js
Svitlana-Yudina: I think the line will be too long and hard to read

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
mykola-koval, line 8:
```suggestion
  for (const action of actions) {
```

src/transformState.js
mykola-koval: no need for `if` here

mykola-koval (approved):

PR #1656, opened 2022-10-20
Solution

andrii-bassarab:


nazarko94 (approved):
Well done! looks pretty clear)

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
Olexandr6, line 13: Let's find a more straightforward way to add new properties to an existing object without using  a `loop`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+    }
+  }
```
Olexandr6, line 28: Where is the `default` block? It's required in the `switch case` statement.

[Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) the 4th point

Olexandr6 (changes requested):
Your solution is good, but not great)

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
```
denys-danyliuk, line 21:
```suggestion
        for (const key in state) {
```

denys-danyliuk (approved):

PR #1658, opened 2022-10-20
solution

nazarko94:


andrii-bassarab (approved):
Good job! 
But in my opinion commented code should be removed

nazarko94:
Because cut off electricity I didn't noticed that I lost inter-connection and I couldn't push edited file with deleted commented code so I deleted the main js file and now I can't recover it the right solution for this task saves in Update transformState.js commit

mykola-koval (changes requested):
can you make another commit which will restore the file?

nazarko94:
> can you make another commit which will restore the file?
image

PR #1659, opened 2022-10-20
solution

Oleksandr-Stryzhavchuk:


Stepico (approved):
Good and clear. The only thing I can think of is that you wrote a semicolon character after for

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+
+        break;
+
+      default:
+        return 'Oops';
```
mykola-koval, line 30: use `throw new Error('error description');` for error handling

mykola-koval (approved):

PR #1661, opened 2022-10-20
Solution

Anna-Shapovalova:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+
+        break;
+
+      default:
+        break;
```
yaro-bey, line 29: use throw Error here (checklist 4)

yaro-bey (approved):
Great job!

src/transformState.js
mykola-koval: +

mykola-koval (approved):

src/transformState.js
Anna-Shapovalova: I`ve already fixed) TNX ;)

PR #1662, opened 2022-10-20
Solution with question

bohdan-stepanets:
I have 2 small questions in this task

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -4,8 +4,34 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  // Is it OK to create a function inside function or better move it above?
+  // Is it OK to create function in this task or better use cycle?
+  function deleteAllProperties() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        continue;
+      case `removeProperties`:
+        for (let i = 0; i < action.keysToRemove.length; i++) {
+          delete state[action.keysToRemove[i]];
+        }
+        continue;
+      case `clear`:
+        deleteAllProperties(state);
+        continue;
+      default:
+    }
+  }
+
+  return state;
 }
 
 module.exports = transformState;
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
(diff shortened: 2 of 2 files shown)


src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
 function transformState(state, actions) {
-  // write code here
+  // Is it OK to create a function inside function or better move it above?
+  // Is it OK to create function in this task or better use cycle?
+  function deleteAllProperties() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
```
Svitlana-Yudina, line 16: I think you should move this inside your function

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+        }
+        continue;
+
+      case `clear`:
+        deleteAllProperties(state);
+        continue;
+
+      default:
+    }
+  }
```
Svitlana-Yudina, line 35: Use code from your external function here, and use break instead continue))

Svitlana-Yudina (changes requested):
I think there is something to fix

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+      case `clear`:
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+    }
+  }
```
Svitlana-Yudina, line 31: add to `default` some action. Like `return 'error'`, or `throw new Error('actions is not correct');`

Svitlana-Yudina (changes requested):

Svitlana-Yudina (approved):
I think it's ok)

src/transformState.js
```diff
@@ -4,8 +4,43 @@
...
+      case `clear`:
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+        // Is it ok to leave default empty?
+        throw new Error('Error');
```
illiarerun, line 39: No, it's not, error text should be meaningful

src/transformState.js
```diff
@@ -4,8 +4,43 @@
...
+
 function transformState(state, actions) {
-  // write code here
+  // At first I've created such function for case `clear`.
+  // Is it OK for this task or better use only cycle?
+  // function deleteAllProperties(state) {
+  //   for (const key in state) {
+  //     delete state[key];
+  //   }
+  // }
```
illiarerun, line 15: You can do it, for sure

src/transformState.js
```diff
@@ -4,8 +4,43 @@
...
+
+        break;
+
+      default:
+        // Is it ok to leave default empty?
+        throw new Error('Error');
+    }
+  }
+
+  return state;
```
illiarerun, line 43: you don't need to return `state` here, you only change existing object
```suggestion
```

illiarerun (approved):

PR #1663, opened 2022-10-20
Solution

webdevw:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const item of rem) {
+        delete state[item];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Bohdan-Maksymiuk, line 25: use `switch case` for such predefined cases

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    const { extraData, keysToRemove: rem } = obj;
```
Bohdan-Maksymiuk, line 9: `keysToRemove: rem` - do not shorten variable names like this

Bohdan-Maksymiuk (changes requested):
Good job - review comments

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const item of keysToRemove) {
+          delete state[item];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 26: don't forget to add default case

polosanya (approved):
Great! I'm approving your work, but please pay attention to my comment

PR #1665, opened 2022-10-21
Solution

Esceype:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
Stepico, line 27: It's better to throw an error while dealing with unexpected input

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete state[property];
+        };
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
```
Stepico, line 21: Maybe, but why? Why not key in state?

Stepico (changes requested):
Good job, just some minor fixes left

Stepico (approved):
GJ, but you still have this Object.keys(state), extra calculations

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
```
denys-danyliuk, line 15:
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformState.js
denys-danyliuk: it's one of the way to do it

denys-danyliuk (approved):

PR #1668, opened 2022-10-21
Task solution

vadymbaranov:


bohdan-stepanets (approved):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const part of actions) {
+    const { type } = part;
+    const { keysToRemove } = part;
+    const { extraData } = part;
```
denys-danyliuk, line 11: same as in stateful clones

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const removeKeys of keysToRemove) {
+          delete state[removeKeys];
+        }
+        break;
+
+      case 'clear':
+        for (const keys in state) {
```
denys-danyliuk, line 25:
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    const { keysToRemove } = part;
+    const { extraData } = part;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const removeKeys of keysToRemove) {
```
denys-danyliuk, line 19:
```suggestion
        for (const key of keysToRemove) {
```

denys-danyliuk (approved):

src/transformState.js
Largoxoxo: > 

Is it really ok to use the same variable name in loop twice?

PR #1674, opened 2022-10-22
js_stateful-object

MXMUK:


Anastasiia-Smirnova (changes requested):
Good job! But the checklist says you should use switch in this task

src/transformState.js
```diff
@@ -4,8 +4,37 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
   // write code here
+
+  for (const action of actions) {
+    switch (true) {
```
denys-danyliuk, line 12:
```suggestion
    switch (action.type) {
```

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+      case (action.type === 'addProperties') :
```
denys-danyliuk, line 13:
```suggestion
      case 'addProperties' :
```

PR #1678, opened 2022-10-23
Solution

Oleksii-Siviolov:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
```
Largoxoxo, line 16: You'r using same variable name twice. Better to change one instance to other name. E.g 'removeKey'.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
Largoxoxo, line 27: Mentors recommend using some error handling here, with 'throw' statement.

Largoxoxo (changes requested):
Gj overall!

src/transformState.js
Oleksii-Siviolov: > You'r using same variable name twice. Better to change one instance to other name. E.g 'removeKey'.

Ah, ok, that's a good point, thank you.

src/transformState.js
Oleksii-Siviolov: > Mentors recommend using some error handling here, with 'throw' statement.

Need to check, how to do it... :)

Largoxoxo (approved):

illiarerun (approved):

PR #1684, opened 2022-10-24
Solution

bonchuk:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      const keysToRemove = actions[key].keysToRemove;
+
+      for (const val in keysToRemove) {
+        delete state[keysToRemove[val]];
+      }
+    } else if (actions[key].type === 'clear') {
+      for (const key2 in state) {
+        delete state[key2];
+      }
+    }
```
tetiana-ivanchenko, line 23: try to use switch case

tetiana-ivanchenko (changes requested):

src/transformState.js
bonchuk: done

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
marina-tilniak, line 8:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (actions[key].type) {
+      case 'addProperties':
+        const extraData = Object.assign({}, actions[key].extraData);
+
+        Object.assign(state, extraData);
```
marina-tilniak, line 13:
```suggestion
        Object.assign(state, action.extraData);
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+      case 'removeProperties':
```
marina-tilniak, line 15:
```suggestion
        break;
        
      case 'removeProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        const keysToRemove = actions[key].keysToRemove;
+
+        for (const val in keysToRemove) {
+          delete state[keysToRemove[val]];
+        }
+        break;
+      case 'clear':
```
marina-tilniak, line 22:
```suggestion
        break;
        
      case 'clear':
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        const extraData = Object.assign({}, actions[key].extraData);
+
+        Object.assign(state, extraData);
+        break;
+      case 'removeProperties':
+        const keysToRemove = actions[key].keysToRemove;
+
+        for (const val in keysToRemove) {
+          delete state[keysToRemove[val]];
+        }
```
marina-tilniak, line 20:
```suggestion
        for (const key in keysToRemove) {
          delete state[keysToRemove[val]];
        }
```

marina-tilniak: let's use for of here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+      case 'clear':
+        for (const key2 in state) {
```
marina-tilniak, line 23:
```suggestion
        for (const key in state) {
```

marina-tilniak (changes requested):

src/transformState.js
bonchuk: I'm sorry I've tried to replace "in" for "of" but its not working correctly((

src/transformState.js
bonchuk: also in this case code can pass the test only with my variant

AlisaVlasova (approved):

PR #1686, opened 2022-10-24
Develop

nazarko94:


src/transformState.js
```diff
@@ -4,8 +4,28 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
-}
+const transformState = (state, actions) => {
+  for (const action of actions) {
+    switch (true) {
```
mykola-koval, line 9: `switch(true)` is a bad idea, I think you can put a different condition here

mykola-koval (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
-}
+const transformState = (state, actions) => {
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const item of Object.keys(state)) {
```
denys-danyliuk, line 11: 3 rule in checklist
```suggestion
        for (const key of Object.keys(state)) {
```

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+          delete state[item];
+        }
+        break;
+
+      case 'removeProperties':
+        for (let i = 0; i < action.keysToRemove.length; i++) {
```
denys-danyliuk, line 17: use for of instead

denys-danyliuk (changes requested):

denys-danyliuk (approved):

PR #1689, opened 2022-10-27
Solution

Oleksandr-Filo:
We can also use 'switch case' instead 'if'.

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -6,16 +6,25 @@
  */
 function transformState(state, actions) {
   for (const action of actions) {
-    if (action.type === 'addProperties') {
-      Object.assign(state, action.extraData);
-    } else if (action.type === 'removeProperties') {
-      for (let i = 0; i < action.keysToRemove.length; i++) {
-        delete state[action.keysToRemove[i]];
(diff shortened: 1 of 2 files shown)


Fynkzia (changes requested):
You right, it's better use switch in this case and checklist required it(pt.2)

Fynkzia (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (let i = 0; i < action.keysToRemove.length; i++) {
```
valerii-derkach, line 15: use` for of` loop here

valerii-derkach (changes requested):

src/transformState.js
Oleksandr-Filo: done

valerii-derkach (approved):

PR #1690, opened 2022-10-27
Solution.

ShkarupaDenys:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    for (const key in obj) {
+      if (obj[key] === 'addProperties') {
```
AlisaVlasova, line 10: it'll be better to use switch case here

AlisaVlasova (changes requested):

AlisaVlasova (approved):

PR #1693, opened 2022-10-29
Complete

SalehAkaJim:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      Object.assign(res, eachObj.extraData);
+    } else if (eachObj.type === 'removeProperties') {
+      for (const key of eachObj.keysToRemove) {
+        delete res[key];
+      }
+    } else if (eachObj.type === 'clear') {
+      for (const each of Object.getOwnPropertyNames(res)) {
+        delete res[each];
+      }
+    } else {}
```
IoghiStudio, line 21: You should use 'if' instead of 'if else' and also delete the last else

IoghiStudio (commented):
You can change the code if you want to change the if else to just if , but delete the last else for sure

IoghiStudio (approved):
Awesome!!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const res = state;
```
meforoma, line 8: i'm pretty sure if you `console.log(res, state)` they will be same
that's because res is a reference to state, but not the standalone variable
you better spread the state:
```suggestion
  const res = [...state];//this will create an individual variable instead of a reference
```

meforoma (approved):
approved.
i've left you an important comment for the future.

PR #1697, opened 2022-10-31
initial

kostasburnazaki:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    switch (true) {
+      case actions[action]['type'] === 'addProperties':
```
DarkMistyRoom, line 10:
```suggestion
    switch (action.type) {
      case 'addProperties':
```

DarkMistyRoom: Why so complicated?

DarkMistyRoom: Make it work everywhere

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        Object.assign(state, actions[action]['extraData']);
+        break;
+
+      case actions[action]['type'] === 'removeProperties':
+        for (const keyToRemove in actions[action]['keysToRemove']) {
+          const propToRemove = actions[action]['keysToRemove'][keyToRemove];
+
+          if (state.hasOwnProperty(propToRemove)) {
+            delete state[propToRemove];
+          }
```
DarkMistyRoom, line 20: Just use .forEach

DarkMistyRoom (changes requested):

Bohdan-Maksymiuk (approved):

PR #1698, opened 2022-10-31
solution

euFarafonov:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
```suggestion
  for (const action of actions) {
```

DarkMistyRoom: Make it work. And don't use for in with arrays

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'addProperties':
+        Object.assign(state, actions[key]['extraData']);
+        break;
+
+      case 'removeProperties':
+        const val = Object.values(actions[key]['keysToRemove']);
+
+        for (let i = 0; i < val.length; i++) {
+          delete state[val[i]];
+        }
```
DarkMistyRoom, line 19: Use .forEach here. it is simpler

DarkMistyRoom (changes requested):

euFarafonov:
fixed loops

SanyaBratashchuk (approved):

PR #1699, opened 2022-11-01
Solution

OlggaJul:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, extraData: toAdd, keysToRemove: toDelete } of actions) {
```
shyptia, line 8: Great usage of destructuring)

shyptia (approved):

src/transformState.js
solaryasha: I agree, but let's use better naming for these properties. You don't have a restriction on variable names length, so let's use descriptive namings
```suggestion
  for (const { type, extraData: propertiesToAdd , keysToRemove } of actions) {
```

solaryasha (approved):

PR #1700, opened 2022-11-01
Solution transformState

vadiimvooo:


4ebupel (changes requested):
I think it's better to change names of your variebles so your code will look better
<sub>Володя може карать за таке, а може й ні, як пощастить))</sub>

4ebupel (approved):
Good Job!

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
solaryasha, line 9: Why ? Action type has restricted values (addProperties, clear,removeProperties), so let's use switch for them

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+      case (action.type === 'addProperties'):
+        Object.assign(state, action.extraData);
+        break;
+
+      case (action.type === 'removeProperties'):
+        action.keysToRemove.map(currentKey => {
```
solaryasha, line 15: It's a bad usage of map method. Let's use forEach instead

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case (action.type === 'clear'):
+        Object.getOwnPropertyNames(state).forEach((stateProperty) => {
+          delete state[stateProperty];
+        });
+        break;
+
+      default:
+        return Error;
```
solaryasha, line 31:
```suggestion
        throw Error;
```

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+          if (state.hasOwnProperty(currentKey)) {
+            delete state[currentKey];
+          }
+
+          return state;
+        });
+        break;
+
+      case (action.type === 'clear'):
+        Object.getOwnPropertyNames(state).forEach((stateProperty) => {
```
solaryasha, line 25: Cool, it's a great usage of that method!

src/transformState.js
solaryasha: And let's some descriptive message for the error

solaryasha (changes requested):

ivan-kadykalo (approved):

PR #1701, opened 2022-11-02
Solution

GabrielMaksimovich:


shyptia (approved):
Looks good to me

src/transformState.js
```diff
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function transformState(state, [...actions]) {
```
solaryasha, line 7: Why do we need a copy of the actions?

solaryasha: Anyway, you will get an array of objects as the second param.

solaryasha (approved):

PR #1706, opened 2022-11-05
Added the solution

zbyn3k:


amitpatiljc (changes requested):
In case of limited condition its always better to use `switch` case

zbyn3k:
Hi Amit, I have rewritten the function to use switch-case. Let me know if it looks good. Thank you

zbyn3k:
Done!

amitpatiljc (approved):
Good job 👍

PR #1709, opened 2022-11-08
Solution

petrolozynskyi:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+      for (const keyToRemove of action.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const stat in state) {
+        delete state[stat];
+      }
+    }
```
liliia-livandovskaya, line 26: Lets use `switch case` instead of `if`

PR #1714, opened 2022-11-09
Solution 1.0

aliliolek:


src/transformState.js
```diff
@@ -1,11 +1,31 @@
 'use strict';
 
 /**
- * @param {Object} state
- * @param {Object[]} actions
- */
+   * @param {Object} state'
+   * @param {Object[]} actions
+   */
```
loralevitska, line 6: why are you changing the dock to a function?

src/transformState.js
```diff
@@ -1,11 +1,31 @@
...
 function transformState(state, actions) {
   // write code here
+
+  for (let i = 0; i < actions.length; i++) {
+    const oneElement = actions[i];
+
+    if (oneElement.type === 'addProperties') {
```
loralevitska, line 13: in each block `if` you compare one value, there is better suited here `switch/case`

loralevitska (changes requested):

src/transformState.js
aliliolek: hmm, sory, I don`t know, I took it from the last task

src/transformState.js
```diff
@@ -1,11 +1,35 @@
...
+
+        for (let m = 0; m < keysRemoving.length; m++) {
+          delete state[`${keysRemoving[m]}`];
+        };
+        break;
+      case ('clear'):
+        for (const keyA in state) {
+          delete state[keyA];
+        };
+        break;
```
polosanya, line 28: Don't forget to add default case

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment

PR #1716, opened 2022-11-10
Solution

vovakov123:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+      for (let m = 0; m < keysRemoving.length; m++) {
+        delete state[`${keysRemoving[m]}`];
+      }
+    } else if (oneElement.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
qoa11a, line 24: Your solution works fine, good job!
But I really recommend to try to solve it with switch. For now you have 3-deep indentation. It makes code hard to read.

qoa11a (changes requested):

liliia-livandovskaya (approved):
Great job! 🔥

PR #1719, opened 2022-11-13
solution

olenitut:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    } else {
+      Object.keys(state).forEach(key => {
+        delete state[key];
+      });
+    }
+  });
```
Bohdan-Maksymiuk, line 20: avoid such nesting `else if` conditions

Bohdan-Maksymiuk (changes requested):
try to use `switch case` for such predefined cases

src/transformState.js
```diff
@@ -5,7 +5,22 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+      default :
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+    }
```
Bohdan-Maksymiuk, line 22: believe that it should be default behaviour
what if someone will pass invalid data - state will be cleared by default
[usefulLink](https://www.quora.com/What-is-the-purpose-of-default-in-a-switch-case)
[usefulLink](https://softwareengineering.stackexchange.com/questions/181459/is-it-necessary-to-add-the-default-case-while-using-switch-cases)

Bohdan-Maksymiuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,22 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+    }
```
Viktor-Kost, line 22: You would better use `default` for switch statement. Cause what will happen if we pass a case we don't have?
For example for default you can throw an error `throw new Error('invalid type of action')`

Viktor-Kost (changes requested):
One more little stuff.
Have given a recommendation

Viktor-Kost (approved):
Well done!

PR #1725, opened 2022-11-14
Solution

panianchuk:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      for (const i of action.keysToRemove) {
+        delete state[i];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    }
```
Bohdan-Maksymiuk, line 21: avoid such nesting `else if` conditions
you can use `switch case` here

Bohdan-Maksymiuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { keysToRemove, extraData, type } = action;
+
+    switch (true) {
+      case type === 'addProperties':
```
Bohdan-Maksymiuk, line 12: same comment
[usefulLink](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)
```suggestion
    switch (type) {
      case 'addProperties':
```
fix everywhere

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        Object.assign(state, extraData);
+        break;
+
+      case type === 'removeProperties':
+        for (const i of keysToRemove) {
```
Bohdan-Maksymiuk, line 17:
```suggestion
        for (const key of keysToRemove) {
```
use meaningful names

Bohdan-Maksymiuk (changes requested):

src/transformState.js
loralevitska: not fixed. If your have some problems with deploy fill free to ask in fe_chat

loralevitska (changes requested):
hi! did you push your changes to github? Previous comments didn't fix.

Viktor-Kost (approved):

PR #1726, opened 2022-11-14
solution

kateryzhen:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    } else if (actions[key].type === 'removeProperties') {
+      const keysToRemove = actions[key].keysToRemove;
+
+      for (const value in keysToRemove) {
+        delete state[keysToRemove[value]];
+      }
+    } else if (actions[key].type === 'clear') {
+      for (const key1 in state) {
+        delete state[key1];
+      }
```
nazarmatsevych, line 22: Avoid such nesting else if conditions
This code is hard to read and maintain.

In each block "if" and "else if" you compare one value, there is better suited here switch/case

nazarmatsevych (changes requested):

nazarmatsevych (approved):
GJ!

PR #1727, opened 2022-11-14
Solution

IVxVI:


src/transformState.js
```diff
@@ -4,8 +4,24 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
Bohdan-Maksymiuk, line 10: try to use [destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) from action object

src/transformState.js
```diff
@@ -4,8 +4,24 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.map((key) => delete state[key]);
+        break;
+
+      default:
+        Object.keys(state).map((key) => delete state[key]);
```
Bohdan-Maksymiuk, line 20: what if someone pass invalid action type?
all properties will be deleted?)
pass this to `clear` case

src/transformState.js
```diff
@@ -4,8 +4,26 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
Bohdan-Maksymiuk, line 10:
```suggestion
    const { type, extraData, keysToRemove } = action;
```
why didn't you destructure another keys

src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete state[key]);
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
+    }
```
Bohdan-Maksymiuk, line 23: also read about `default` case in switch and why do we need it

Bohdan-Maksymiuk (approved):

PR #1729, opened 2022-11-15
solution

Dmytro-Litinskyi:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
tetiana-ivanchenko, line 9: let's use switch case

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      case 'removeProperties':
+        for (const char of actions[i].keysToRemove) {
+          delete state[char];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
tetiana-ivanchenko, line 23: it isn't default case it is case for 'clear'

use default for error handling, like
throw new Error('error description');

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
```
tetiana-ivanchenko, line 9:
```suggestion
  for (const action of actions) {
    switch (action.type) {
```

tetiana-ivanchenko (changes requested):

src/transformState.js
Dmytro-Litinskyi: Whats the diference?

src/transformState.js
tetiana-ivanchenko: fix please this comment

src/transformState.js
tetiana-ivanchenko: just more consize

tetiana-ivanchenko (changes requested):

nazarmatsevych (approved):
👍

PR #1731, opened 2022-11-16
Solution

IrynaK-a:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    // type contains a string: 'addProperties', 'removeProperties' or 'clear';
```
polosanya, line 9: Don't forget to delete comments

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+    if (obj.type === 'addProperties') {
```
polosanya, line 11: Use switch...case instead

polosanya (changes requested):
Good job! Please check my comments to improve your code

loralevitska (approved):
great!

PR #1734, opened 2022-11-18
Task's solution. All test are passed.

Chupiak:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+        }
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
SanyaBratashchuk, line 29: use `for..of or forEach` and switch case inside

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const i in actions) {
+    switch (actions[i].type) {
+      case 'addProperties' : {
+        for (const key in actions[i].extraData) {
+          state[key] = actions[i].extraData[key];
+        }
```
AlisaVlasova, line 14:
```suggestion
  for (const key in actions) {
    switch (key.type) {
      case 'addProperties' : {
        Object.assign(state, key. extraData);
```

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+        break;
+      }
+
+      case 'removeProperties': {
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
```
AlisaVlasova, line 19: use for...of or forEach here to

AlisaVlasova (changes requested):

Chupiak:
Please check

AlisaVlasova (approved):

PR #1736, opened 2022-11-19
Solution

llllewvvaa:


mr-valentos (approved):
GJ

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[property];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
ivan-kadykalo, line 31: There is no requirement to return something. This is mutable function

ivan-kadykalo (approved):

PR #1741, opened 2022-11-21
Solution

Deniskakaka:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    } else if (actions[i].type === 'addProperties') {
```
Viktor-Kost, line 13: Try to avoid `if else` nesting. It's a harder to maintain and read. 
So, for this situation you can use `switch` statement. And don't forget about returning a value by default (it can be even an error)
Some tips to write better conditions [useful_link](https://www.digitalocean.com/community/posts/5-tips-to-write-better-conditionals-in-javascript)

liliia-livandovskaya (approved):
🔥

PR #1745, opened 2022-11-21
Solution

romanblyashuk94:


superpooperxxx (approved):
Good job. I like putting everything into separate function.

src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+    keysToDelete.forEach(key => {
+      delete object[key];
+    });
+  }
+
+  function clearAll(obj) {
+    for (const key in obj) {
+      delete obj[key];
+    }
+  }
```
ivan-kadykalo, line 43: Interesting solution, but why do you need these functions? You can write this logic in your switch cases.

ivan-kadykalo (changes requested):

solaryasha (approved):
Nice usage of the `forEach` method!

PR #1746, opened 2022-11-21
Solution

nurdanokcu:


ysrckr (approved):
Looks good. Just a suggestion it is all checks equality so you could use switch statement.

mgrinko (approved):

PR #1747, opened 2022-11-21
Solution

vlad-boikov:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    if (obj.type === 'addProperties') {
+      for (const key in obj.extraData) {
+        state[key] = obj.extraData[key];
+      }
+    } else if (obj.type === 'removeProperties') {
+      obj.keysToRemove.forEach(el => delete state[el]);
+    } else if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
SanyaBratashchuk, line 18: use `switch case`

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        };
+        break;
+      case 'removeProperties':
+        obj.keysToRemove.forEach(el => delete state[el]);
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
stanislav-topikha, line 22: checklist
image

stanislav-topikha (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    switch (obj.type) {
+      case 'addProperties':
+        for (const key in obj.extraData) {
+          state[key] = obj.extraData[key];
+        };
+        break;
+      case 'removeProperties':
+        obj.keysToRemove.forEach(el => delete state[el]);
+        break;
+      default:
```
DarkMistyRoom, line 18: This case should work only when the action is 'clear', but if it is the default it will work in all cases except listed above.
You should leave 'clear' here and add a default case to handle "wrong" inputs

DarkMistyRoom (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        for (const key in obj.extraData) {
+          state[key] = obj.extraData[key];
+        };
```
DarkMistyRoom, line 13: It is ok, but you should know that you can do the same this way:
`Object.assign(state, action.extraData);`

DarkMistyRoom (approved):
Good

PR #1749, opened 2022-11-22
Solution

UmizDemud:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        const dataToAdd = Object.entries(action.extraData);
+
+        for (const data of dataToAdd) {
+          state[data[0]] = data[1];
+        }
+        break;
```
lefteris-L, line 16: I like the `Object.entries` array approach

lefteris-L (approved):
Great solution 👍

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+            delete state[keyToRemove];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
lefteris-L, line 30: It would be better to have a default clause

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        const dataToAdd = Object.entries(action.extraData);
+
+        for (const data of dataToAdd) {
+          state[data[0]] = data[1];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          if (keyToRemove in state) {
```
mgrinko, line 20: you don't need this check

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        const dataToAdd = Object.entries(action.extraData);
+
+        for (const data of dataToAdd) {
+          state[data[0]] = data[1];
+        }
```
mgrinko, line 15:
```suggestion
        const [key, value] = Object.entries(action.extraData);

        for (const data of dataToAdd) {
          state[key] = value;
        }
```

mgrinko (approved):

PR #1750, opened 2022-11-22
solution

KrisPlanka:


superpooperxxx (changes requested):
Good Job, but checklist 2

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
ivan-kadykalo, line 31: There is no need to return something, it is a mutable function.

ivan-kadykalo (approved):

src/transformState.js
KrisPlanka: Ok, thank you, done

PR #1751, opened 2022-11-22
solution

lefteris-L:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
CharAnagn, line 13: You could use Object.assign() for less code

CharAnagn (approved):
LGTM 👍 💯

src/transformState.js
lefteris-L: Didn't think of that. How about this? `state = { ...state, ...action.extraData };`

CharAnagn: I also did it this way at first, but i believe there was an error that said we shouldn't modify the originial state like that. The same when at the case 'clear' I tried state = { }, instead of deleting a key one by one, but it wasn't correct.

lefteris-L: Yeah, even when state is `let ` it doesn't seem to work

mgrinko (approved):

PR #1753, opened 2022-11-23
solution

Lun4trick:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    switch (actions[action].type) {
```
CharAnagn, line 9: You basically get the action from the loop, so you can access the type with only action.type and all the other instances that you used it.

CharAnagn (approved):
GG

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const removable in action.keysToRemove) {
+          delete state[action.keysToRemove[removable]];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
```
mgrinko, line 21:
```suggestion
        Object.keys(state)
          .forEach(key => delete state[key]);
```

mgrinko (approved):

PR #1755, opened 2022-11-24
Finished task

JzSol:


Alireza-Mohammadhossein (commented):
Nice job!

meforoma (changes requested):
please use the `switch..case` construct for this task
the algorithm is fine, just move it inside `switch..case`

src/transformState.js
```diff
@@ -1,11 +1,33 @@
...
+"use strict";
 
 /**
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
naziamoff, line 9:
```suggestion
  for (const action of actions) {
```

naziamoff (approved):
🔥

meforoma (approved):

PR #1756, opened 2022-11-24
solution

grslckr:


meforoma (changes requested):
your entire algorithm is reasonable,
just one important thing - it's a common approach TO NOT mutate function inputs - state in this case.

Introduce a function variable that is an exact copy of the state and work with this function variable instead of mutating function input
`const stateCopy = { ...state }`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
+    }
+  }
```
meforoma, line 28: i believe you have to return a state?

meforoma (changes requested):
i think you need to return a state?
might be wrong.

meforoma (approved):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
meforoma, line 26: why leaving an empty `default:` case?
it would be reasonable to put here, say `clear` path:
```suggestion
      default:
        for (const key in state) {
          delete state[key];
        }
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
meforoma, line 14: if you include this in `default:` case, you won't need it here
```suggestion
```

meforoma (changes requested):
see the comment about the default case

grslckr:
> your entire algorithm is reasonable, just one important thing - it's a common approach TO NOT mutate function inputs - state in this case.
> 
> Introduce a function variable that is an exact copy of the state and work with this function variable instead of mutating function input `const stateCopy = { ...state }`

i changed default as you say but
do you mean a link of state? because tests are checking the state, not a value we return if i'm understanding right this code:
  expect(state)
    .toEqual({
      foo: 'bar', bar: 'foo', name: 'Jim', hello: 'world',
    });

meforoma (approved):
@grslckr you're right, this one expects mutating behaviour.
👍

PR #1758, opened 2022-11-24
solution

Alireza-Mohammadhossein:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
```
CharAnagn, line 20: You can set it as the default case

CharAnagn (approved):
🔥

meforoma (approved):

PR #1759, opened 2022-11-24
Solution

Rijo-J:


mmahfuzi (changes requested):
Great job! Just you need to return state at the end of your function :)

Rijo-J:
> Great job! Just you need to return state at the end of your function :)

Thank you!
It was on purpose, I was thinking the actual value is not needed for further functionalities imo, just to change  an exterior reference (which it does) but will do! 😁 👍

mmahfuzi:
> > Great job! Just you need to return state at the end of your function :)
> 
> Thank you! It was on purpose, I was thinking the actual value is not needed for further functionalities imo, just to change an exterior reference (which it does) but will do! 😁 👍

Ahaha I see :D

mmahfuzi (approved):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        Object.assign(state, obj.extraData);
+        break;
+      case 'removeProperties':
+        obj.keysToRemove.forEach(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+      default:
+        break;
```
meforoma, line 20: don't leave `default:` case blank
at least `throw new Error('not supported action')` or something like that

meforoma (approved):

PR #1761, opened 2022-11-24
solution

mmahfuzi:


sercansimsek (approved):
Nicely done!👍 🚀

src/transformState.js
```diff
@@ -6,6 +6,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
meforoma, line 28: no sense of keeping an empty `default:` case, but you can make it behave as `clear` i suppose
```suggestion
      default:
        for (const key in state) {
          delete state[key];
        }
```

meforoma (changes requested):
see the comment about the `default:` case

src/transformState.js
meforoma: just thought, you could actually keep the default
but don't keep it empty - at least console.log some info or even better `throw new Error('action is not supported')`

meforoma (approved):

src/transformState.js
mmahfuzi: Sure, thanks

PR #1763, opened 2022-11-24
Solution

sercansimsek:


Rijo-J (approved):
🔥 🚀 👍

Rijo-J (approved):
Looks good to me! 🚀

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
meforoma, line 27: you could just make `clear` path a default behaviour:
```suggestion
      default:
        for (const key in state) {
          delete state[key];
        }
```

meforoma (changes requested):
see the comment about `default:` case

meforoma (approved):
👍

PR #1764, opened 2022-11-25
Solution

Mohameddahi:


mmahfuzi (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      case 'removeProperties':
+        for (const key of action['keysToRemove']) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
meforoma, line 23: it can be like this for this task, yet if you would call the function like this:
`transformState({}, {type: 'unknown type'})` - it will trigger the default case of your switch.

So it's better to explicitly cover all the known cases and for default - provide an explanatory fallback, like
```
```

meforoma (approved):
good 👍 
consider one comment for the future

PR #1766, opened 2022-11-25
Develop

Grayfox147:


mmahfuzi (changes requested):
Good job! Just at the end of your function you need to return state and also it's better to not keep the default empty.  you can add:   throw new Error('action is not supported')  to default.

mmahfuzi (approved):

meforoma (approved):
nice 👍

PR #1767, opened 2022-11-25
Solution

RomeoKs:


mmahfuzi (approved):
Good job! Just one small comment, it's better not to keep the default empty - you can add: throw new Error('action is not supported')

RomeoKs:
> Good job! Just one small comment, it's better not to keep the default empty - you can add: throw new Error('action is not supported')

I can't pat myself on the back for a solution learned during the lesson :(

I thank you for the help. It really means a lot to me!

mmahfuzi:
> > Good job! Just one small comment, it's better not to keep the default empty - you can add: throw new Error('action is not supported')
> 
> I can't pat myself on the back for a solution learned during the lesson :(
> You have already proved how great you are in the previous tasks, definitely if you had time you could have done it yourself  so pat yourself on the back for all including this one ;) 

> I thank you for the help. It really means a lot to me!
you are most welcome, though it was nothing :D

meforoma (approved):
sweet 🏅

PR #1770, opened 2022-11-25
solution

oleksandrnikora:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
 
+      case 'default':
+        break;
```
oleh-yablunovskyi, line 27: Looks like it should be just default, according to the switch documentation.
```suggestion
      default:
        break;
```

oleh-yablunovskyi (approved):
Everything is okay, except for one minor moment.

ivan-kadykalo (approved):
🔥

PR #1771, opened 2022-11-25
Solution

NgineZZ:


Kavster92 (commented):
Nice code Fez! Appreciate the line breaks for clarity, I did the same.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+
```
meforoma, line 15:
```suggestion
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
+    }
```
meforoma, line 29: don't leave `default:` blank - provide an explanatory fallback, like
```
```

meforoma (approved):
good.
just one comment

PR #1772, opened 2022-11-25
solution

kenvastolord:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
NgineZZ, line 26: The whole code looks good. One recommandation as per guideline "switch/case should always have default case for error handling."

kenvastolord: ok thanks for the suggestion. i will add it.

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
meforoma, line 28: don't leave `default:` blank - provide an explanatory fallback, like
```
```

meforoma (approved):

PR #1773, opened 2022-11-25
Retrieved State of Object, Transformed State Based on Rules Given

Kavster92:


Lun4trick (approved):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const order of actions) {
+    switch (order.type) {
+      case 'addProperties':
+        Object.assign(state, order.extraData);
+
+        break;
+
+      case 'removeProperties':
```
kenvastolord, line 15: good job working on this task, just to make sure , is it good idea to leave a blank line after and before the break statement?

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+        break;
```
meforoma, line 30: don't leave `default:` blank - provide an explanatory fallback, like
```
```

meforoma (approved):
just one comment for the future 👍

PR #1775, opened 2022-11-25
solution

victoria-kovalenko:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case 'removeProperties':
+
+        for (const key of action.keysToRemove) {
+          if (key in state) {
+            delete state[key];
+          }
+        }
+        break;
```
viktoriia-kravchenko, line 27: I'd recommend removing an empty line between case and for.
We can omit if (key in state).

viktoriia-kravchenko (approved):
Good job!

valerii-derkach (changes requested):
fix comments

valerii-derkach (approved):

PR #1785, opened 2022-11-25
Solution

Sergii-Osedchuk:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
llllewvvaa, line 29: It is a mutable function, no need to return something
```suggestion
```

llllewvvaa (changes requested):

src/transformState.js
Sergii-Osedchuk: fixed

Ihor-Karpyn (approved):
LGTM

PR #1790, opened 2022-11-26
Solution

miftaha:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
mmahfuzi, line 28: Its better not to keep default empty. You can add : throw new Error('action is not supported')

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+    }
+  }
 }
```
mmahfuzi, line 31: Your function should return state.

mmahfuzi (changes requested):

mmahfuzi:
Well done! just two small comments :)

mmahfuzi (approved):

naziamoff (changes requested):
You should not use the `for` loop for iterating over an array. There's a better one
And when you use switch/case you should have the default case

naziamoff (approved):
Well done

PR #1791, opened 2022-11-26
Solution for transform state

katerinalex:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
+    if (i.type === 'addProperties') {
+      Object.assign(state, i.extraData);
+    } else if (i.type === 'removeProperties') {
```
polosanya, line 11: Avoid nesting else...if. Try switch...case instead
https://www.w3schools.com/js/js_switch.asp

polosanya (changes requested):
Good job! Please provide some fixes

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
```
```suggestion
  for (const action of actions) {
```
Let's use more informative names

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (i.type) {
+      case 'addProperties':
+        Object.assign(state, i.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const j of i.keysToRemove) {
```
polosanya, line 15: Same here. What is j?

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+            delete state[j];
+          }
+        }
+        break;
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 26: It would be great to add a default case to avoid potential errors in a situation where the type does not match any of the cases.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#syntax

polosanya (changes requested):
Same comments as in the stateful-clones task 😉

Anastasiia-Svintsova (approved):

PR #1797, opened 2022-11-27
Solution

novytskyi-d:


src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+          delete state[properties];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
llllewvvaa, line 32: There is no requirement to return something. This is a mutable function

llllewvvaa (changes requested):

llllewvvaa (approved):

Ihor-Karpyn (approved):
good job

PR #1802, opened 2022-11-27
solution

anastasiia-roiek:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
Mykhailyk-Vadym, line 31: no need to return something. this is mutable function.

Mykhailyk-Vadym (changes requested):
good job. Just delete return.

src/transformState.js
anastasiia-roiek: done!

Mykhailyk-Vadym (approved):
nice

Ihor-Karpyn (approved):
good job

PR #1812, opened 2022-11-28
Solution

vladyslav-krainyk:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+      Object.assign(state, obj.extraData);
+    } else if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    } else if (obj.type === 'removeProperties') {
+      for (const keyToRemove of obj.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    }
```
qoa11a, line 19: Let's use switch statement instead of "if".

qoa11a (changes requested):

AlisaVlasova (approved):

PR #1815, opened 2022-11-29
Solution

Anastasia4sik:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
```
VladimirEdelmann, line 21: Keep one way of naming. Here should be `keyToRemove` too

VladimirEdelmann (approved):

src/transformState.js
Anastasia4sik: do I really need it here? I'm using keyToRemove with for...of loop action.keysToRemove

VladimirEdelmann: same meaning

valerii-derkach (approved):

PR #1819, opened 2022-11-30
state

Vadym-Prydatok:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+      }
+
+      if (action[key] === 'clear') {
+        for (const property in state) {
+          if (property in state) {
+            delete state[property];
+          }
+        }
+      }
+    }
```
Bohdan-Maksymiuk, line 33: use `switch case` instead of this internal loop

Bohdan-Maksymiuk (changes requested):

src/transformState.js
Vadym-Prydatok: тримайте switch

AlisaVlasova (approved):

PR #1825, opened 2022-12-03
Solution

Ukrainiane-panda:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Viktor-Kost, line 24: It's a good practice to add default case for switch statement. If you don't have any default behavior - just throw an Error at default state

Viktor-Kost (changes requested):
Well done! 

Just make some improvements

Viktor-Kost (approved):
GJ!

PR #1828, opened 2022-12-05
Task Guideline

anastasia-radchencko:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return state;
```
Viktor-Kost, line 27: You should not return state by default
Just throw an error for default behavior with a message "invalid type of action"

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
```
Viktor-Kost, line 21:
```suggestion
        for (const key of state) {
```

Viktor-Kost (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'invalid type of action';
```
tetiana-ivanchenko, line 27: it's just string, check prev comments
image

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

tetiana-ivanchenko (changes requested):

stanislav-topikha (approved):

PR #1831, opened 2022-12-06
Task solution

Dynamizx:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  const trans = state;
```
polosanya, line 9: Do you really need it? In this task you should change the original state.
image
Btw, if you wanted to create a copy of object, it is not working like that 
https://www.javascripttutorial.net/object/3-ways-to-copy-objects-in-javascript/

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+
+  for (const a in actions) {
```
polosanya, line 11:
```suggestion
  for (const action in actions) {
```
Let's provide full, descriptive names

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+    switch (actions[a].type) {
+      case 'addProperties':
+        Object.assign(trans, actions[a].extraData);
+        break;
+      case 'removeProperties':
+        for (const k of actions[a].keysToRemove) {
```
polosanya, line 17: Same here and below. You can also make a [destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) to get all the properties you need from actions[action]

polosanya (changes requested):
Almost perfect!

qoa11a (approved):

PR #1832, opened 2022-12-07
first success

spencerspe:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const changes = Object.values(actions);
+
+  const toRemove = [];
+
+  for (let i = 0; i < changes.length; i++) {
```
polosanya, line 12: You can just do like that:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+  // console.log(changes[i])
```
polosanya, line 13: Don't forget to delete the comments

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    if (changes[i]['type'] === 'addProperties') {
+      Object.assign(state, changes[i]['extraData']);
+    }
+
+    if (changes[i]['type'] === 'removeProperties') {
```
polosanya, line 18: image

polosanya (changes requested):
Almost done! Let's provide some fixes

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'addProperties':
+        Object.assign(state, elemment['extraData']);
+        break;
+      case 'removeProperties':
+        Object.assign(toRemove, elemment['keysToRemove']);
+        break;
+      case 'clear':
+        for (const entries in state) {
+          delete state[entries];
+        }
```
qoa11a, line 21:
```suggestion
        for (const prop in state) {
          delete state[prop];
        }
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
-  // write code here
+  const toRemove = [];
+
+  for (const elemment of actions) {
+    switch (elemment['type']) {
+      case 'addProperties':
+        Object.assign(state, elemment['extraData']);
+        break;
+      case 'removeProperties':
+        Object.assign(toRemove, elemment['keysToRemove']);
```
qoa11a, line 16: It's very hard to read solution and it's also bad in terms of performance. Delete required props right away.

qoa11a (changes requested):

src/transformState.js
spencerspe: I'm not sure what you mean here?

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        for (const value of action.keysToRemove) {
+          delete state[value];
+        }
+        break;
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
```
polosanya, line 22: Let's add a default case to avoid potential errors if some type doesn't match any of cases
image

polosanya (changes requested):
Almost perfect! Just small fix left

spencerspe:
Okay, thanks. Would console.log("No match found") make sense as a default?

polosanya (approved):
Good job!

PR #1836, opened 2022-12-09
solution

vasyliev-anton:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
AlisaVlasova, line 9: use switch case

AlisaVlasova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          if (state[i]) {
+            delete state[i];
+          }
+        }
+        break;
+
+      default:
```
agniya-i, line 22:
```suggestion
      case 'clear'
``` 
it is better to specify the сase

agniya-i (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        const objKeys = Object.keys(state);
+
+        objKeys.forEach(key => {
+          delete state[key];
+        });
+    }
```
witflash, line 28: `switch / case` needs to have `default` case (for error handling for example)

witflash (changes requested):
GJ!
But check my last comment, please

witflash (approved):
Good work 👍🏻

PR #1841, opened 2022-12-12
Solution stateful object

anna-akhalmosulishvili:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
```
Bohdan-Maksymiuk, line 9: why do you need index in this loop? use `for of`
```suggestion
  for (const action of actions) {
    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        for (const keys of actions[i].keysToRemove) {
+          delete state[keys];
+        }
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+    }
```
Bohdan-Maksymiuk, line 23: also is a good point to add default case to switch
it's crucial for handling errors

Bohdan-Maksymiuk: https://stackoverflow.com/questions/4649423/should-switch-statements-always-contain-a-default-clause

Bohdan-Maksymiuk (approved):

tetiana-ivanchenko (changes requested):
fix comments

tetiana-ivanchenko (changes requested):
please, check and fix ALL comments (default case)

PR #1842, opened 2022-12-14
Solution

Vadivel-Babu:


src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
```
polosanya, line 24: Don't forget to add the default case to avoid possible errors when action.type does not match any of the cases

polosanya (changes requested):
Almost perfect!

src/transformState.js
jstmpelowycz: fix this comment, please

jstmpelowycz (changes requested):
what do you put inside `default` case? reach me out in Slack

polosanya (approved):
Well done!

PR #1843, opened 2022-12-15
Function transformState

ilona-romaniuk:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
+    for (const key in object) {
+      if (object[key] === 'addProperties') {
```
loralevitska, line 10: use `switch case` instead of nested `if`

loralevitska (changes requested):

Anastasiia-Svintsova (approved):

PR #1848, opened 2022-12-18
solution of transformState

RomanTsvindin:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete transformedState[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in transformedState) {
+        delete transformedState[key];
+      }
+    }
+  }
```
Anastasiia-Svintsova, line 22: Use switch/case instead of if/else

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #1851, opened 2022-12-19
Solution

zvir91:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
```

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
polosanya, line 9: You can also make a destructuring to get all the needed properties of action to make your code cleaner

polosanya (changes requested):
Almost perfect! Please check the comments)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (true) {
+      case type === 'addProperties' :
```
polosanya, line 12:
```suggestion
    switch (type) {
      case 'addProperties' :
```
In all cases you check the type, so why do you need to switch true?

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case type === 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
polosanya, line 25: Don't forget to add a default case to handle potential errors in case type doesn't match any of cases

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
```suggestion
    const { 
      type, 
      extraData, 
      keysToRemove, 
    } = action;
```

polosanya (changes requested):
Almost perfect!

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default: let noType = {};
+
+        noType = state;
+
+        return noType;
```
Anastasiia-Svintsova, line 34:
```suggestion
        break;

      default: 
        break;
```

Anastasiia-Svintsova (approved):

PR #1853, opened 2022-12-19
Solution

tarasavage:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
Viktor-Kost, line 24: Try to avoid `else if` nesting
Use `switch `statements instead. And don't forget about default case when you use switch

Viktor-Kost (changes requested):

PR #1855, opened 2022-12-22
Solution

kuzhelek:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
loralevitska, line 23: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

loralevitska (changes requested):

src/transformState.js
kuzhelek: Switch-case construction has been applied instead of simple compare one value with others

kuzhelek:
Switch-case construction has been applied instead of simple compare one value with others

kuzhelek:
Can you check this changes too please?

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
```suggestion
    const { 
      type, 
      extraData, 
      keysToRemove, 
    } = action;
```

polosanya (approved):
Well done!

PR #1856, opened 2022-12-22
Solution

KseniyaGonchar:


vlad-demchuk (changes requested):
Good, but it will be clearer if you use switch case here🙂

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+
+      case 'removeProperties':
+        for (let n = 0; n < actions[i].keysToRemove.length; n++) {
+          delete state[actions[i].keysToRemove[n]];
+        };
+        break;
+
+      default:
+        Object.keys(state).forEach(n => delete state[n]);
```
tetiana-ivanchenko, line 21: it's not "default", it's "clear" case, default shoul hadling errors

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
GJ just add, please, default case that will handling errors

KseniyaGonchar:
sorry, i don`t understand what is GJ

AlisaVlasova (approved):
GJ means good job)

KseniyaGonchar:
i thought it is abbreviation in js, broke my brain, it is time to rest, thanks)))

PR #1857, opened 2022-12-23
Solution

ruslan-yarosh:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
```
ViktorHavryliuk, line 12: Please use switch statement instead of multiple ifs

ViktorHavryliuk (changes requested):

TarasHoliuk (approved):

PR #1858, opened 2022-12-23
Solution

OlhaYevstifieieva:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      default:
+        break;
+    }
+  }
+
+  return state;
```
andrieieva-yuliia, line 28: You don't need to return anything here 🙂

andrieieva-yuliia (approved):
Looks great 😎

src/transformState.js
alexandra-protyanova: agree

alexandra-protyanova (changes requested):

alexandra-protyanova (approved):
Well done 👍

PR #1863, opened 2022-12-25
Develop

jaxxxon78:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties' :
+
+        Object.assign(state, action.extraData);
+
+        break;
```
witflash, line 14:
```suggestion
      case 'addProperties' :
        Object.assign(state, action.extraData);
        break;
```

As for me, there is no need to add these empty lines.

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      case 'removeProperties' :
+        for (const remove of action.keysToRemove) {
```
witflash, line 16: It's better to name it `key` instead of `remove` because you iterate `KEYStoRemove`.

witflash (approved):
Good job!
I left some comments for minor fixes.

PR #1868, opened 2022-12-25
Solution

Kardash-Oleksandr:


src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    const check = Object.values(action);
 
+    switch (check[0]) {
```
andrieieva-yuliia, line 12: slightly unclear variable name for me 🙂

andrieieva-yuliia (approved):

src/transformState.js
Kardash-Oleksandr: fixed

src/transformState.js
```diff
@@ -4,8 +4,24 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear' :
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+      case 'addProperties' :
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties' :
+        for (const property of action.keysToRemove) {
```
alexandra-protyanova, line 18: You can use `Object.keys & forEach` here as well

alexandra-protyanova (approved):
GJ!

PR #1870, opened 2022-12-26
Solution

AntonKabanets:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        delete state[i];
+      };
+    }
+
+    if (action.type === 'addProperties') {
+      for (const [key, value] of Object.entries(action.extraData)) {
+        state[key] = value;
+      }
+    }
+  }
```
tetiana-ivanchenko, line 26: 1. use switch...sace instead of if blocks
2. o and i - bad naming, use more meanfull names

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
please, make sure that you fixed all coments 
image
 also you should add default case to error handlings

vlad-demchuk (approved):

PR #1871, opened 2022-12-26
Solution

sv-dubov:


nzrii-kysel (approved):
well done👍
could use a `switch` statement for a cleaner look

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
Denchepornyuk, line 22: instead of three `if` you can use switch case

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1873, opened 2022-12-26
Develop: Finished function transformState

sonyaxent:


src/transformState.js
```diff
@@ -4,8 +4,21 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
agniya-i, line 10: it's better to change to `switch` statement instead of using `if-else`

agniya-i (changes requested):

Virass (approved):

Virass (approved):

PR #1877, opened 2022-12-27
Solution

metalikaqq:


ArturErmolenko (changes requested):
checklist point 3,4 
use `for of` loop

ArturErmolenko (approved):

PR #1879, opened 2022-12-27
solution

Dashaverz:


src/transformState.js
```diff
@@ -6,6 +6,23 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
ann-yurkevych, line 10: It's better to use here switch case :
`case 'addProperties':

case 'removeProperties':
case 'clear':
`

ann-yurkevych (approved):
You've done an amazing job))

alexandra-protyanova (changes requested):
Checklist point 2

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+          delete state[key];
+        }
+        break;
 
+      default:
+        throw new Error('unknown type');
+    }
+  }
+
+  return state;
```
alexandra-protyanova, line 32: You don't need to return anything from function

alexandra-protyanova (approved):
Well done 👍

PR #1881, opened 2022-12-27
Solution

BrandonSunrise:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
anna-inozemets, line 9: Maybe it is better to use switch instead of if

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
```
anna-inozemets, line 14: You have loop in another loop (cheklist | point 5)

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        delete state[keyToRemove];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
```
anna-inozemets, line 20: The same, checklist | point 5 = Nested loops === EVIL

anna-inozemets (approved):
Good job, just small fixes🐇

src/transformState.js
BrandonSunrise: Got it! i will fix it !

src/transformState.js
BrandonSunrise: Got it! i will fix it !

src/transformState.js
BrandonSunrise: Got it! i will fix it !

alexandra-protyanova (changes requested):
Fix previous comments

alexandra-protyanova (approved):
Well done 👍

PR #1883, opened 2022-12-27
Test was solved

ihor-soloviov:


src/transformState.js
```diff
@@ -6,6 +6,26 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
```
Avichez, line 13: Checklist 2, Use `switch` statement if you have a limited amount of conditions.

Avichez (changes requested):
Great solution 👍 
Just remain to change `if` statement to `switch`.

src/transformState.js
ihor-soloviov: allright. forgot about it

sv-dubov (approved):
Now looks better with switch :)

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
ArturErmolenko, line 9: don't use `true` in switch case, 
```suggestion
    switch (action.type) {
```

ArturErmolenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Oups! Incorrect type!');
+    }
+  }
+
+  return state;
```
alexandra-protyanova, line 31: you don't need to return anything

alexandra-protyanova (approved):

PR #1890, opened 2022-12-28
v-01

AndreyBerezyanskiy:


tania-boiar (approved):
Well done! 🔥

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        for (const keysRemove of el.keysToRemove) {
+          delete state[keysRemove];
+        };
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty([key])) {
+            delete state[key];
+          }
```
Denchepornyuk, line 24: you don't need `hasOwnProperty` here, delete will ignore keys which doesn't exist

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const el of actions) {
```
Denchepornyuk, line 8: `el` isn't a good name for variable

Denchepornyuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        break;
```
Denchepornyuk, line 27: add some logic to default block

Denchepornyuk (approved):

PR #1895, opened 2022-12-28
Finished task)

andrii-kurash:
Thanks for review 😉🔥

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let res = {};
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      res = Object.assign(state, action.extraData);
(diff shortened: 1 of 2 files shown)


OlhaYevstifieieva (changes requested):
Here you also need to use switch case 🤗

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.getOwnPropertyNames(state).forEach(key => delete state[key]);
```
alexandra-protyanova, line 17: A little overkill, `Object.keys` will work fine as well)

alexandra-protyanova (approved):

PR #1897, opened 2022-12-28
Solution

MFrolova-art:


chermak-oleh (changes requested):
Well done! According to checklist point 2, it is better to use switch statement here.

Denchepornyuk (approved):

PR #1899, opened 2022-12-28
solution

OlehSharha:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
victorionov, line 9: In Checklist:
#2 [CODE STYLE]: Use switch statement if you have limited amount of conditions.
Remake it with "swith  case" instead "if".

victorionov (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const removleKey of action.keysToRemove) {
```
alexandra-protyanova, line 15:
```suggestion
        for (const removleKey of action.keysToRemove) {
```
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        throw new Error('Something wrong');
+    }
+  }
+
+  return state;
```
alexandra-protyanova, line 30: You don't need to return anything from function

alexandra-protyanova (approved):
Well done 👍

PR #1900, opened 2022-12-28
Solution

victorionov:


andrii-kurash (changes requested):
You can add default value in the end of  your switch-case. But it isn't mandatory)

Well done!) 👍

andrii-kurash (approved):

src/transformState.js
```diff
@@ -5,7 +5,20 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const transform of actions) {
+    switch (transform.type) {
```
Denchepornyuk, line 9: where is default block?

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #1903, opened 2022-12-28
Stateful object solution

PolinaSukhorukova:


nzrii-kysel (changes requested):
good solution, but checklist advises you to use `switch` statement

alexandra-protyanova (approved):
Well done 🚀

PR #1904, opened 2022-12-28
Solution

karaedd:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
 }
```
marina-tilniak, line 25: let's use switch case in this task

marina-tilniak (changes requested):

Viktor-Kost (approved):
Nice!

For future - it's a good practice to use `default` case at switch statement. If you don't have default behavior - just throw an error

PR #1907, opened 2022-12-28
Solution

andrii-andriushchenko:


andrieieva-yuliia (approved):
Nice work)

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case ('addProperties'):
```
Denchepornyuk, line 10: you don't need brackets for cases
```suggestion
      case 'addProperties':
```

Denchepornyuk (approved):

src/transformState.js
andrii-andriushchenko: > you don't need brackets for cases

Thanks, removed

PR #1911, opened 2022-12-29
solution

Anton-Zhytonbaiev:


OlhaYevstifieieva (changes requested):
The same, you must use switch statement 🙃

alexandra-protyanova (approved):

PR #1914, opened 2022-12-29
solution

AlexPro100pro:


nzrii-kysel (approved):

src/transformState.js
```diff
@@ -6,6 +6,30 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('action not found')
+    }
+  }
+
+  return state;
```
alexandra-protyanova, line 32: You don't need to return anything

alexandra-protyanova (approved):

PR #1915, opened 2022-12-29
solution_1

Borys-Andrew:


sv-dubov (approved):
Good job!

src/transformState.js
```diff
@@ -6,6 +6,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error(Error);
```
Anna-Tkachenko, line 28: will be better to specify understandable message for error

Anna-Tkachenko (approved):
Good job!

PR #1916, opened 2022-12-29
solution

popeye88:


src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error(Error);
```
polosanya, line 28:
```suggestion
        throw new Error('Error text');
```
Specify the error

src/transformState.js
```diff
@@ -4,8 +4,32 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 10: You can make destructuring to get all needed properties of action and make your code cleaner

polosanya (changes requested):
Almost perfect! Let's improve your code a bit

src/transformState.js
popeye88: Pushed. Believ inderstood you well

SanyaBratashchuk (approved):

PR #1917, opened 2022-12-30
transform state solution

serbinm:
I know I probably could've done in a more optimized way, like: 
- use switch instead of iterating array of actions
- may be use ForEach instead of making iteration every time

but I like training like this, so don't judge me hard :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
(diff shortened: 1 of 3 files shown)


sv-dubov (changes requested):
Yes, its better to use switch in this task. I also made like you at first. But than had to remade it with switch :)

ArturErmolenko (changes requested):
Checklist point 2 & 3

ArturErmolenko (approved):

PR #1919, opened 2022-12-31
Stateful object solution

AV1788:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+  for (const elem of actions) {
+    switch (elem.type) {
+      case 'addProperties':
+        Object.assign(state, elem.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key in state) {
+          for (const elem1 of elem.keysToRemove) {
+            if (key === elem1) {
```
tetiana-ivanchenko, line 17: avoid nesting loop and if - it's redundant

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const elem of actions) {
```
```suggestion
  for (const action of actions) {
```

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```

loralevitska (approved):

PR #1921, opened 2023-01-01
Solution

VadymTsyndra:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case (action.type === 'clear'):
```
tetiana-ivanchenko, line 10:
```suggestion
    switch (action.type) {
      case 'clear':
```

tetiana-ivanchenko (changes requested):
also add default with error handling

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let action of actions) {
+    switch (action.type) {
+      case ('clear'):
```
agniya-i, line 10:
```suggestion
      case 'clear':
```
you can remove brackets

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case ('removeProperties'):
+        for (const prop of action.keysToRemove) {
+          delete state[prop];
+        }
+
+        break;
+
+      default:
+        action = null;
```
agniya-i, line 30:
```suggestion
       throw new Error('Unexpected state');
```
it's better to throw error like this

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+    }
+  }
+
+  return state;
```
agniya-i, line 34: do not need to return anything from the function

SanyaBratashchuk (approved):

PR #1922, opened 2023-01-03
add solution

NikitaNevv:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
agniya-i, line 9: use `switch-case` instead of `if-else`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
agniya-i, line 8: you can use `for...of` to keep it simple
```suggestion
  for (const action in actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const index in actions[i].keysToRemove) {
```
agniya-i, line 14:
```suggestion
      for (const key in actions[i].keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete state[actions[i].keysToRemove[index]];
```
agniya-i, line 15:
```suggestion
        delete state[key];
```

agniya-i (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
SanyaBratashchuk, line 8: Do not iterate through array via `for in`  loop use `for of` instead

SanyaBratashchuk (changes requested):

agniya-i (approved):

PR #1923, opened 2023-01-03
done

Alessandranitti15:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
+    if (actions[key].type === 'clear') {
```
jstmpelowycz, line 9: let's use switch-case statements

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
jstmpelowycz, line 8: you don't have to iterate over `actions` with for-in loop, use `forEach`

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    }
+
+    if (actions[key].type === 'addProperties') {
+      Object.assign(state, actions[key].extraData);
+    }
+
+    if (actions[key].type === 'removeProperties') {
+      for (const azione in actions[key].keysToRemove) {
+        delete state[actions[key].keysToRemove[azione]];
```
jstmpelowycz, line 21: split this single-line action into multi-line actions so that it becomes more readable

jstmpelowycz (changes requested):
awesome! almost done, a couple of things to fix

src/transformState.js
Alessandranitti15: Hi! I have corrected the task js_stateful-objects according to your suggestions, but I won't use "forEach" as you suggested. I tried but I cannot. That is in a future module (Array Iteration Method), but I am still at the "Object Advanced" module. I hope the "for in" will work for now. Thank you for your understanding.

qoa11a (changes requested):
Good job! 👍 
Please see my comments to this task you've done and apply same changes here.
https://github.com/mate-academy/js_stateful-clones/pull/1885#pullrequestreview-1236995781

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      default:
+        throw new Error('unknown action');
```
jstmpelowycz, line 27: 🔥

jstmpelowycz (approved):

PR #1924, opened 2023-01-03
Solution

ihor-halkin:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const arr of actions) {
+    if (arr['type'] === 'clear') {
```
Anastasiia-Svintsova, line 9: Use switch/case instead of `if` statements

Anastasiia-Svintsova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const arr of actions) {
+    switch (true) {
+      case arr['type'] === 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        };
+        break;
```
Anastasiia-Svintsova, line 14:
```suggestion
    switch (arr['type']) {
      case 'clear':
        for (const prop in state) {
          delete state[prop];
        };
        break;
```

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+
+      case arr['type'] === 'addProperties':
+        Object.assign(state, arr['extraData']);
+        break;
+
+      case arr['type'] === 'removeProperties':
```
Anastasiia-Svintsova, line 20:
```suggestion
      case 'removeProperties':
      default:
```

Anastasiia-Svintsova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const arr of actions) {
+    switch (arr['type']) {
```
marina-tilniak, line 9:
```suggestion
    switch (arr.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const arr of actions) {
```
marina-tilniak, line 8:
```suggestion
  for (const action of actions) {
```

src/transformState.js
marina-tilniak: you need square brackets only when you don't know name of property(when you use variables for that)

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+    switch (arr['type']) {
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        };
+        break;
+
+      case 'addProperties':
+        Object.assign(state, arr['extraData']);
```
marina-tilniak, line 17: same here

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const prop of arr['keysToRemove']) {
```
marina-tilniak, line 21: same hare

marina-tilniak (changes requested):
It's good practice write switch case with default

ihor-halkin:
> It's good practice write switch case with default

Hello
I don't understand what I should to do in default case.
Thanks for you comments

loralevitska:
> > It's good practice write switch case with default
> 
> Hello I don't understand what I should to do in default case. Thanks for you comments

You should add `default` property 
image

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```

loralevitska (changes requested):

ihor-halkin:
> > > It's good practice write switch case with default
> > 
> > 
> > Hello I don't understand what I should to do in default case. Thanks for you comments
> 
> You should add `default` property, so that the code executes with the default settings if no value is suitable image

I understand what mean 'default case'. If i got right without default case, loop will be continiue, so i don't understand what speciefic  i should to write in default. Thanks

Anastasiia-Svintsova (approved):

PR #1926, opened 2023-01-05
Finish task

U-Ra-All:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
mykhalenych, line 10: switch case will be better

mykhalenych (changes requested):

src/transformState.js
U-Ra-All: Thanks, done

Anastasiia-Svintsova (approved):

PR #1930, opened 2023-01-07
added solution

constantweb:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const prop of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (prop.type) {
+      case 'addProperties':
+        for (const key in prop.extraData) {
+          state[key] = prop.extraData[key];
+        }
```
tetiana-ivanchenko, line 13: you can use Object.assign to make copy

tetiana-ivanchenko (changes requested):
also

tetiana-ivanchenko (approved):

PR #1931, opened 2023-01-07
Solution

ghost:


src/transformState.js
```diff
@@ -6,6 +6,21 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const obj of actions) {
```
mykhalenych, line 9: I think will be better `for (const action of actions)`

src/transformState.js
```diff
@@ -6,6 +6,21 @@
...
+    if (obj.type === 'addProperties') {
```
mykhalenych, line 10: change to `switch (action.type) {`

mykhalenych (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const keyword of action.keysToRemove) {
+          delete state[keyword];
+        };
+        break;
+      case 'clear':
+        for (const properties of Object.keys(state)) {
```
tetiana-ivanchenko, line 20:
```suggestion
        for (const key in state) {
```

tetiana-ivanchenko (changes requested):

PR #1934, opened 2023-01-10
solution transform state

Spec0s:


src/transformState.js
```diff
@@ -6,6 +6,38 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  function addProps(add) {
```
```suggestion
  function addProps(propsToAdd) {
```
Let's use more descriptive names for function props

src/transformState.js
```diff
@@ -6,6 +6,38 @@
...
+    for (const key in add) {
+      state[key] = add[key];
+    }
+  }
+
+  function removeProps(remove) {
```
polosanya, line 15: Same here

src/transformState.js
```diff
@@ -6,6 +6,38 @@
...
+  }
 
+  function clearAllProps() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 28: You can make a destructuring to get all needed props of action (type, extraData etc.)

polosanya (changes requested):
Almost perfect!

src/transformState.js
Spec0s: i am not sure if i got it right but you tell me :)

qoa11a (approved):
Good job! 👍

PR #1936, opened 2023-01-13
Solution

Pallaci0:


nazarmatsevych (changes requested):
Good work,
just add default case for error handling
image

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+        break;
+
+      case 'clear':
+
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
```
tetiana-ivanchenko, line 29:
```suggestion
        break;
        
      default:
```

tetiana-ivanchenko (approved):
GJ 🔥

PR #1938, opened 2023-01-14
solution

44shaggy:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i in actions) {
```
```suggestion
  for (const action of actions) {
```
image
image

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    switch (actions[i].type) {
```
```suggestion
    switch (action.type) {
```
Also, you can make destructuring to get all the needed properties of action.

polosanya (changes requested):
Almost done!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const j of keysToRemove) {
+          delete state[j];
```
polosanya, line 18:
```suggestion
        for (const key of keysToRemove) {
          delete state[key];
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
polosanya, line 27: Don't forget to add a default case
image

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
```suggestion
    const { 
      type, 
      extraData, 
      keysToRemove 
    } = action;
```

polosanya (changes requested):
Let's polish it up

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        console.log('Unexpected action type'); // eslint-disable-line no-console
```
polosanya, line 33: It is better to throw an error or just break the code

polosanya (approved):
Well done!

PR #1939, opened 2023-01-14
transformState

Oleksandr-Maslo:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    const type = obj.type;
```
nazarmatsevych, line 9: looks redundant but up to you

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[keysRemove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
nazarmatsevych, line 27: Add default case for error handling

nazarmatsevych (changes requested):
Good work,
pls check my comments

TarasHoliuk (approved):

PR #1940, opened 2023-01-15
solution added

ghost:


VadimMentorQA (approved):
Goof job.
A little advice - it can be more readable with switch case usage.

PR #1941, opened 2023-01-16
Solution

OlyaNikora:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
 
+      default:
```
mykhalenych, line 26: default does nothing

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    switch (action.type) {
```
mykhalenych, line 9: you can destruct values from action
`const {type, extraData ...} = action `

mykhalenych (changes requested):

mykhalenych (approved):

PR #1942, opened 2023-01-16
task done

tecHamster:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```

loralevitska (changes requested):
Almost done)

qoa11a (approved):

PR #1943, opened 2023-01-16
js_stateful-object

RomanHrytsiuk:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const item of property.keysToRemove) {
+        delete state[item];
+      }
+    }
+
+    if (property.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
marina-tilniak, line 23: let's use switch case here

marina-tilniak (approved):

marina-tilniak (changes requested):
sorry, wrong approval

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const property of actions) {
```
loralevitska, line 8: Destructure `property`

loralevitska (approved):

PR #1945, opened 2023-01-17
Develop

Peti51:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
+    if (actions[key].type === 'addProperties') {
```
polosanya, line 9: image

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
polosanya, line 8: image

polosanya (changes requested):
Let's improve your code a bit

qoa11a (approved):

PR #1946, opened 2023-01-17
the task was solved

batiukova-nataliia:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    } else if (objects.type === 'removeProperties') {
+      for (const remove of objects.keysToRemove) {
+        delete state[remove];
+      }
+    } else {
+      for (const deleteAll in state) {
+        delete state[deleteAll];
+      }
+    }
```
OlEzhA0, line 21: try to use switch case here

OlEzhA0 (changes requested):

IvanFesenko (approved):

PR #1947, opened 2023-01-17
Solution

kirill-vypirovskyi:


kuznetsova-anastasiia (approved):
GJ :)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+  const addProps = (props) => {
+    Object.assign(state, props);
+  };
+
+  const removeProps = (keys) => {
+    for (const key of keys) {
+      delete state[key];
+    }
+  };
```
Andr1yK, line 22: let's move it to outside of this function (now u create those functions on each call of transformState)

Andr1yK (approved):

PR #1948, opened 2023-01-18
Solution transformState

NazariiAlieksieiev:
I think solution can be more shortly and with less code, or no?

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -6,6 +6,35 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  const result = state;
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i]['type'] === 'addProperties') {
+      const extraData = actions[i]['extraData'];
+
+      for (const key in extraData) {
+        result[key] = extraData[key];
+      };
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -6,6 +6,35 @@
...
+      for (let j = 0; j < keysToRemove.length; j++) {
+        for (const key in result) {
+          if (keysToRemove[j] === key) {
+            delete result[key];
+          };
+        };
+      };
+    };
+
+    if (actions[i]['type'] === 'clear') {
```
witflash, line 32:
```suggestion
    if (actions[i].type === 'clear') {
```

src/transformState.js
```diff
@@ -6,6 +6,35 @@
  */
 function transformState(state, actions) {
   // write code here
-}
+  const result = state;
```
witflash, line 9: There is no need to create new `result` constant. 
You can mutate `state` directly.

witflash (changes requested):
Pay attention to my comments, please.

Viktor-Kost (approved):

PR #1949, opened 2023-01-18
Solution

nadia-hope:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (obj.type === 'addProperties') {
+      Object.assign(state, obj.extraData);
+    }
```
Also, you can make a destructuring to get all the needed props of action(obj)

polosanya (changes requested):
Good job! Let's improve your code a bit

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Anastasiia-Svintsova, line 28: You should add a default case to the switch/case constructions. For example
```suggestion
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
        default: 
        break;
```

Anastasiia-Svintsova (approved):

PR #1950, opened 2023-01-19
Solution

annperetiatko:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
mykhalenych, line 8:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    for (const element in obj) {
```
mykhalenych, line 9: it is redundant

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+      const value = obj[element];
+      const extraData = obj.extraData;
+      const keysToRemove = obj.keysToRemove;
+
+      if (value === 'addProperties') {
```
mykhalenych, line 14: change to switch, and use Object.assign(state, action.extraData);

mykhalenych (changes requested):

Virass (approved):

PR #1952, opened 2023-01-19
Solution

G37ua:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
agniya-i, line 9: it would be better to use `switch-case` instead of `if-else` statements

agniya-i (changes requested):

G37ua:
corrected

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can make a destructuring to get all the needed props of action (type, extraData, keysToRemove).
It will make your code cleaner

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+    }
```
polosanya, line 24: image

polosanya (changes requested):
Almost done!

G37ua:
corrected

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
+        break;
+
+      default:
+        break;
```
nazarmatsevych, line 29: It's better to throw new Error in default case with text for example: 'Unknown action type'

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```

nazarmatsevych (approved):
Good work,
just small suggestions

PR #1953, opened 2023-01-19
Solution

pushkarskiyrodion:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
DarkMistyRoom, line 24: switch/case should always have default case for error handling

DarkMistyRoom (approved):

PR #1954, opened 2023-01-20
My solution

andriy-litkovets:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

nazarmatsevych (approved):
Good work,
just small suggestion

src/transformState.js
andriy-litkovets: thanks

PR #1957, opened 2023-01-21
Solution

dianaavem:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const rem in action.keysToRemove) {
+        delete state[action.keysToRemove[rem]];
+      }
+    } else {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
SanyaBratashchuk, line 21: use `switch case`

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+      case 'removeProperties':
+        for (const rem in action.keysToRemove) {
+          delete state[action.keysToRemove[rem]];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
tetiana-ivanchenko, line 26: it's not default case, its 'clear' case, default should hadling errors

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'ERROR';
```
agniya-i, line 30: it's better to use `throw new Error()`

agniya-i (changes requested):

ViktorHavryliuk (approved):

PR #1958, opened 2023-01-21
Develop

oleksandr-plias:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        Object.getOwnPropertyNames(state).forEach(function(prop) {
```
Volodymyr-Melnyk-Fron10der, line 22: It's better to use arrow function as callback, if you don't need to use `this`
`forEach((prop) => ...)`

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
Volodymyr-Melnyk-Fron10der, line 16: Use forEach method as well, to be consistent with line 22

Volodymyr-Melnyk-Fron10der (approved):
Good job! There are only few not critical items to improve. PR is Approved.

PR #1960, opened 2023-01-22
The tests were passed successfully

read-my-code:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const item of actions) {
+    for (const key in item) {
```
Volodymyr-Melnyk-Fron10der, line 10: Sorry, I don`t understand why are you mapping all keys in action.. You have to catch every type of action and make changes with object. Example:
```
for (const { type, extraData, ...otherPropsYouNeed } of actions) {

switch (type) {
   case "extraData":
      ......your changes
      break;
   case "keysToRemove"
   ....... so on
}
}   
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default :
+        return;
```
Volodymyr-Melnyk-Fron10der, line 27: You can throw error if action is unknown 
```suggestion
      default :
        throw new Error(`Action ${type} is unknown!`);;
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: Use destructurization 

```suggestion
  for (const { type, extraData, keysToRemove } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+    switch (action.type) {
+      case 'addProperties' :
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
```
Volodymyr-Melnyk-Fron10der, line 17: Use array method forEach, it will be just one line of code

```suggestion
        keysToRemove.forEach(key => delete state[key))
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
```
Volodymyr-Melnyk-Fron10der, line 23: Same story)

```suggestion
        Object.keys(state).forEach(key => delete state[key])
```

Volodymyr-Melnyk-Fron10der (changes requested):
Cool! Much better! There are just a few improvements left to make your code perfect)

SanyaBratashchuk (approved):

PR #1962, opened 2023-01-23
Solution

georgiy-nezbritskiy:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+      for (const keyToRemoveProperties of action.keysToRemove) {
+        delete state[keyToRemoveProperties];
+      }
+    }
+
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        state[key] = action.extraData[key];
+      }
+    }
```
vkryvytskyy, line 26: It's better to use `switch` in this case

vkryvytskyy (approved):

PR #1963, opened 2023-01-23
function was created

YevheniiHura:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete copiedState[arr[k]];
+      }
+    } else if (actions[i].type === 'clear') {
+      for (const key in copiedState) {
+        if (copiedState.hasOwnProperty(key)) {
+          delete copiedState[key];
+        }
+      }
+    }
+  }
```
Viktor-Kost, line 26: try to avoid else if nesting
You can apply switch. It will be much better

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const copiedState = state;
+
+  for (let i = 0; i < actions.length; i++) {
```
Viktor-Kost, line 10:
```suggestion
  for (const action of actions) {
```

Viktor-Kost (changes requested):

src/transformState.js
YevheniiHura: done

src/transformState.js
YevheniiHura: done

Viktor-Kost (approved):
Nice!

PR #1965, opened 2023-01-23
Solution

AnastasiiaEm:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (let x = 0; x < actions[i].keysToRemove.length; x++) {
+          delete state[actions[i].keysToRemove[x]];
+        };
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        };
+        break;
+    }
```
Viktor-Kost, line 25: it is a good practice if you add default behavior (if you don't have such one - throw an error)

Viktor-Kost (approved):
Nice!

PR #1966, opened 2023-01-24
Solution

valeriaoleynik:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
agniya-i, line 24: add a default case that throws an error in the case of an unknown type

agniya-i (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const caseDefault = 'An error because of an unknown type';
+
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 11: You can also make a destructuring to get all the needed props of action (type, extraData, keysToRemove)

polosanya (approved):
Good job!

PR #1967, opened 2023-01-24
First solution

Vasyl-Pavlenko:
Which solution better for case clear? From my solution or if use loop?

        / for (const key in state) {
        // if (state.hasOwnProperty(key)) {
        //   delete state[key];
        // }
        // }

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          delete state[i];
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+
+      case 'removeProperties':
+        for (const i of action.keysToRemove) {
+          delete state[i];
+        };
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
```
witflash, line 22: Look. `switch / case` should have `default` statement always. You can handle some errors inside it, for example

witflash:
> Which solution better for case clear? From my solution or if use loop?

For this case, your solution is good because we have guaranteed correct key names. 
For the future, it's best practice to check if the key exists.

witflash (changes requested):
Good job!
Check my comment, please.

Viktor-Kost (approved):
GJ!

PR #1969, opened 2023-01-25
Develop

oksanatytanych:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const part in actions) {
+    const action = actions[part];
+
+    if (action.type === 'addProperties') {
```
OlEzhA0, line 11: try to use switch case

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      Object.assign(state, action['extraData']);
+    };
+
+    if (action.type === 'removeProperties') {
+      for (const removeElement in action.keysToRemove) {
+        for (const key in state) {
+          if (key === action.keysToRemove[removeElement]) {
+            delete state[key];
+          }
+        }
```
OlEzhA0, line 21: as I see, here is should be just
```suggestion
      for (const removeElement in action.keysToRemove) {
        delete state[removeElement];
```

OlEzhA0 (changes requested):

PR #1971, opened 2023-01-25
Develop

silcki:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const i in actions) {
```
Bohdan-Maksymiuk, line 9: [CODE STYLE]: don't use for in loop for iterating over array

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+    const { type, extraData = {} } = actions[i];
+    let { keysToRemove = [] } = actions[i];
+
+    if (type === 'addProperties') {
```
Bohdan-Maksymiuk, line 13: [CODE STYLE]: Use switch statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+      Object.assign(state, extraData);
+
+      continue;
+    }
+
+    if (type === 'clear') {
+      keysToRemove = Object.keys(state);
```
Bohdan-Maksymiuk, line 20: you can just remove all values
you don't need assign keysToRemove for that

Bohdan-Maksymiuk (changes requested):

Bohdan-Maksymiuk (commented):
check you code with [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) one more time

PR #1974, opened 2023-01-26
task was solved

vanyachyzh:


src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const item in actions[i].extraData) {
+        state[item] = actions[i].extraData[item];
+      }
+
+      resultArray.push({
+        ...state,
+      });
```
IvanFesenko, line 18: Do not repeat your self, do this action once at the end

IvanFesenko (changes requested):
Good job. Please fix code duplicating

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
```
AlisaVlasova, line 11: use switch/case

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      for (const item in actions[i].extraData) {
```
AlisaVlasova, line 12: use Object.assign

AlisaVlasova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (true) {
+      case actions[i].type === 'addProperties':
+        Object.assign(state, actions[i].extraData);
```
OlEzhA0, line 13: don't mutate input state, add some constant

OlEzhA0 (changes requested):

src/transformState.js
vanyachyzh: - `state` is an object. You are supposed to add, change, or delete its
  properties instead of creating a new object

vanyachyzh: I think I have to mutate state, don't I?

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const resultArray = [];
+
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
```
IvanFesenko, line 11: for of loop will be better 
```suggestion
  for (const action of actions) {
    switch (actions.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+      case 'addProperties':
+        Object.assign(state, actions[i].extraData);
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          delete state[actions[i].keysToRemove[j]];
```
IvanFesenko, line 18: Same here
```suggestion
       for (const key of action.keysToRemove) {
         delete state[key];
```

IvanFesenko (changes requested):
Good job. Let's make some improvements

etojeDenys (approved):
LGTM!

PR #1975, opened 2023-01-27
Stateful object

mdm-21:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'removeProperties':
+        action.keysToRemove.forEach(item => {
+          delete state[item];
+        });
```
ViktorHavryliuk, line 13:
```suggestion
        action.keysToRemove.forEach(key => {
          delete state[key];
        });
```

ViktorHavryliuk (approved):
Good work, only one small suggestion regarding variable names

PR #1977, opened 2023-01-27
Solution

Bohdan33:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
loralevitska, line 10: Destructure `action`

loralevitska (changes requested):
Almost perfect)
Let's improve

DarkMistyRoom (approved):

PR #1978, opened 2023-01-27
solved

danulyk05:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[ck];
+        };
+      };
+    };
+
+    if (ch.type === clearing) {
+      for (const del in state) {
+        delete state[del];
+      };
+    };
```
loralevitska, line 29: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+  const clearing = 'clear';
+
+  for (const ch of actions) {
```
loralevitska, line 12: Destructure `action`

```suggestion
  for (const action of actions) {
```

loralevitska (changes requested):
Good job! Let's improve

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
 
+  for (const action of actions) {
+    switch (action.type) {
+      case add:
+        Object.assign(state, action.extraData);
+        break;
+
+      case remove:
+        for (const ck of action.keysToRemove) {
+          if (ck in state) {
```
tetiana-ivanchenko, line 20:
```suggestion
          if (key in state) {
```

tetiana-ivanchenko (approved):

PR #1979, opened 2023-01-29
● Solution

d-hubych:


src/transformState.js
```diff
@@ -5,7 +5,48 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can make destructuring to get all the needed props of action (type, extraData, keysToRemove)

src/transformState.js
```diff
@@ -5,7 +5,48 @@
...
+
+      case 'removeProperties': {
+        removeProps(state, action.keysToRemove);
+        break;
+      }
+
+      case 'clear': {
+        clearAll(state);
+        break;
+      }
```
polosanya, line 23: image

polosanya (changes requested):

TarasHoliuk (approved):

PR #1983, opened 2023-01-31
Solution

katiailchenko:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+        // type === 'clear'
```
ivadimko, line 25: mb better to add `case clear` explicitly here just in order for everyone to know that object can clear values

I strongly believe "default" behavior shouldn't clear the object cause it's too dangerous

you can define

```
default: {
  // do nothing
}
```

ivadimko (approved):
gj 🔥

src/transformState.js
katiailchenko: 👍

PR #1988, opened 2023-02-01
'develop'

Hanna-Balabukha:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
tetiana-ivanchenko, line 9: use switch/case

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
check checklist before review, please
image

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        action.type = 'undefined';
+        break;
+    }
```
loralevitska, line 31: Better return state

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
loralevitska, line 9: Better use destructure

loralevitska (approved):

PR #1989, opened 2023-02-01
Object

Yevhen-Ivanchuk:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const removeElement of action.keysToRemove) {
+          delete state[removeElement];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
agniya-i, line 26: add default case to throw an error in case of unknown type

agniya-i (changes requested):

PR #1990, opened 2023-02-02
Solution

vlad-ptashnichenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+      }
+    }
+
+    if (i.type === 'clear') {
+      for (const itemClearState in state) {
+        delete state[itemClearState];
+      }
+    }
+  }
```
etojeDenys, line 28: it's better to use here switch/case instead of `if`

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const index of i.keysToRemove) {
+          for (const itemState in state) {
+            if (index === itemState) {
+              delete state[itemState];
+            }
+          }
+        }
```
OlEzhA0, line 21: ?
```suggestion
        for (const index of i.keysToRemove) {
          delete state[index]
        }
```

OlEzhA0 (changes requested):
hey! please check my comment

etojeDenys (approved):
LGTM

nazarmatsevych (approved):

PR #1991, opened 2023-02-02
stateful object created

yevhen1408:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const toDo in action.extraData) {
```
loralevitska, line 10: The variable name must be a noun (about `toDo`). Rename it please

src/transformState.js
loralevitska, line 9: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

loralevitska (changes requested):
Almost done)
Let's improve

tetiana-ivanchenko (changes requested):
fix checklist point 4

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        }
+        break;
+
+      case remove:
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
ViktorHavryliuk, line 25: This is not default case, it's "clear" action type. Default should be for situation when action type is neither add, remove nor clear, and throw user error informing that provided action type is not supported

ViktorHavryliuk (changes requested):

SanyaBratashchuk (approved):

PR #1993, opened 2023-02-02
Solution

diamat1997:


tetiana-ivanchenko (changes requested):
fix checklist point 4

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        for (const removeKey of action.keysToRemove) {
+          delete state[removeKey];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      default :
```
ViktorHavryliuk, line 27: Please throw an error to let user know that provided action type is not supported

ViktorHavryliuk (changes requested):

src/transformState.js
SanyaBratashchuk, line 27:
```suggestion
      default :
      throw new Error('unknown action type')
```

SanyaBratashchuk (changes requested):

DarkMistyRoom (approved):

PR #1994, opened 2023-02-03
Stateful Object v.0.1 - solution

NedobezhkinOleksandr:
Please, find first version of task

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const clear in state) {
+          delete state[clear];
+        }
+        break;
+
+      case 'addProperties':
+        for (const add in action) {
+          if (add === 'extraData') {
+            Object.assign(state, action[add]);
+          }
+        }
+        break;
+
+      case 'removeProperties':
+        for (const input of action.keysToRemove) {
+          delete state[input];
+        }
+    }
+  }
+
+  return state;
 }
 
 module.exports = transformState;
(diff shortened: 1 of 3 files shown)


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[clear];
+        }
+        break;
+
+      case 'addProperties':
+        for (const add in action) {
+          if (add === 'extraData') {
+            Object.assign(state, action[add]);
+          }
+        }
```
loralevitska, line 21: Name for variables should be noun, now you have `add` it's action. Rename it please
```suggestion
        for (const add in action.extraData) {
          Object.assign(state, action[add]);
        }
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const input of action.keysToRemove) {
+          delete state[input];
+        }
```
loralevitska, line 27: Check checklist before review, please
image

loralevitska (changes requested):
Good job! There are things that should be fixed:

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+        break;
+
+      default:
+        return state;
```
ViktorHavryliuk, line 33:
```suggestion
        throw new Error("Unsupported action type");
```

ViktorHavryliuk (approved):
Throwing error would be better as it will let the user know that something is wrong, and even better, what exactly is wrong

src/transformState.js
NedobezhkinOleksandr: OMG, how can i know this? Ahaha, thank you, will change now

PR #1997, opened 2023-02-03
Solution

RuslanLiapin:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+          if (key === keyRemove) {
+            delete state[key];
+          }
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const clearKey in state) {
+        delete state[clearKey];
+      }
+    }
```
nazarmatsevych, line 27: Looks to complicated.
Try to rewrite it using switch/case.

nazarmatsevych (changes requested):
If you have any questions - feel free to ask them in fe_chat

src/transformState.js
```diff
@@ -6,6 +6,35 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties': {
```
TarasHoliuk, line 11: That seems the same as using `if`
write `action.type` in `switch` instead of true

src/transformState.js
```diff
@@ -6,6 +6,35 @@
  */
 function transformState(state, actions) {
   // write code here
```
TarasHoliuk, line 8: delete this comment

src/transformState.js
```diff
@@ -6,6 +6,35 @@
...
+    switch (true) {
+      case action.type === 'addProperties': {
+        const addedState = action.extraData;
+
+        Object.assign(state, addedState);
+        break;
+      };
+
+      case action.type === 'removeProperties': {
+        for (const key in state) {
```
TarasHoliuk, line 19: outer loop is redundant

TarasHoliuk (changes requested):

IvanFesenko (approved):
Good job

PR #1998, opened 2023-02-05
stateful object first commit

RufatRM:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const refToState = state;
+
+  for (let i = 0; i < actions.length; i++) {
```
polosanya, line 10:
```suggestion
  for (action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    if (actions[i].type === 'addProperties') {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const refToState = state;
```
polosanya, line 8: Why do you need it? Can't you just use state?

polosanya (changes requested):

src/transformState.js
RufatRM: I can, it was late at night:)

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const {
      type,
      extraData,
      keysToRemove,
    } = action;
    
    switch (type) {
```
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

polosanya (approved):
Good job! I'm approving it, but please pay attention to my comment)

PR #2000, opened 2023-02-05
Solution

Zavalishyn-Oleksandr:
IDK if I should create variable for actions[i] or actions[i][key] to make syntax shorter so waiting for fixes

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    for (const key in actions[i]) {
+      if (key === 'type' && actions[i][key] === 'addProperties') {
+        Object.assign(state, actions[i]['extraData']);
+      }
+
+      if (key === 'type' && actions[i][key] === 'removeProperties') {
+        const arrayToDelete = actions[i]['keysToRemove'];
+
+        for (let j = 0; j < arrayToDelete.length; j++) {
+          delete state[arrayToDelete[j]];
+        }
+      }
+
+      if (key === 'type' && actions[i][key] === 'clear') {
+        for (const name in state) {
+          delete state[name];
+        }
+      }
+    }
+  }
 }
 
 module.exports = transformState;
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
     "eslint": "^5.16.0",
     "eslint-plugin-jest": "^22.4.1",
     "eslint-plugin-node": "^8.0.1",


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+        for (let j = 0; j < arrayToDelete.length; j++) {
+          delete state[arrayToDelete[j]];
+        }
+      }
+
+      if (key === 'type' && actions[i][key] === 'clear') {
+        for (const name in state) {
+          delete state[name];
+        }
```
DarkMistyRoom, line 25: Use switch statement if you have limited amount of conditions

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
DarkMistyRoom, line 8: better use for of as there is no need in the index

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    for (const key in actions[i]) {
```
DarkMistyRoom, line 9: don't use this second cycle. the `action `is an object, so you can just call `action.type` property

DarkMistyRoom: if yo have questions, ask in the chat

DarkMistyRoom (changes requested):

Zavalishyn-Oleksandr:
I've decided to use forEach insted of 'for of', it seems more readable for me, is it ok?

qoa11a (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case (elem['type'] === 'addProperties'):
+        Object.assign(state, elem['extraData']);
+        break;
+
+      case (elem['type'] === 'removeProperties'):
+        const arrayToDelete = elem['keysToRemove'];
+
+        for (let j = 0; j < arrayToDelete.length; j++) {
+          delete state[arrayToDelete[j]];
+        }
```
loralevitska, line 19:
```suggestion
        for (const key of action.keysToRemove) {
          delete state[key];
      }
```

loralevitska (changes requested):
I recommend you to refresh your knowledge about [switch/case](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/switch)

Zavalishyn-Oleksandr:
Fixed

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
+      case ('addProperties'):
```
loralevitska, line 10: You don't need parentheses in this case, fix all the others too
```suggestion
      case 'addProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        break;
+
+      case ('clear'):
+        for (const name in state) {
+          delete state[name];
+        }
+        break;
+
+      default: return 'Please, enter the valid data';
```
loralevitska, line 26:
```suggestion
      default: 
        return 'Please, enter the valid data';
```

loralevitska (approved):
Good job!

PR #2002, opened 2023-02-06
Solution

ArtemVelychko:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const transform of actions) {
```
```suggestion
  for (const action of actions) {
```
I think it's more logical to call it "action" as you are iterating through actions.
Also, you can make a destructuring to get all the needed props of it (type, extraData, keysToRemove)

src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+        for (const remove of transform.keysToRemove) {
+          delete state[remove];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 25: image

polosanya (changes requested):

polosanya (approved):
Good job!

PR #2003, opened 2023-02-07
Solution

BogdanParubchenk01:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+    for (const key in obj) {
```
DarkMistyRoom, line 9: you don't need a second for here. an action is an object and you know all its properties. you just need to check `action.type`

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+            }
+          }
+        }
+      }
+
+      if (obj[key] === 'clear') {
+        for (const result in state) {
+          delete state[result];
+        }
+      }
```
DarkMistyRoom, line 32: Use switch statement if you have limited amount of conditions.

DarkMistyRoom (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const push in action.extraData) {
```
IvanFesenko, line 11: push better call value.

IvanFesenko (approved):
Well done

PR #2006, opened 2023-02-08
Solution

yanZibierov:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
IvanFesenko, line 28: Use `switch` instead

IvanFesenko (changes requested):
Good job. Please refactor with using switch

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Viktor-Kost, line 27: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just `throw an Error` with text like `"unexpected action"`

Viktor-Kost (approved):

mykhalenych (approved):

PR #2008, opened 2023-02-08
Solution

luka4ok:
Please, check my solution.

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const deleteItem of action.keysToRemove) {
+        delete state[deleteItem];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const property in state) {
+        delete state[property];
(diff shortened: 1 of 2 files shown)


ViktorHavryliuk (changes requested):
Good work, but please use switch case instead of multiple ifs. And don’t forget to add default case for error handling

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case (action.type === 'removeProperties'):
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        };
+        break;
+
+      case (action.type === 'clear'):
```
DarkMistyRoom, line 20:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        };
        break;

      case 'clear':
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        for (const property in state) {
+          delete state[property];
+        };
+        break;
+
+      default:
+        return state;
```
DarkMistyRoom, line 27: would be great to inform user about wrong action type

DarkMistyRoom (changes requested):

slavanosatskyi (approved):
Nice work!

PR #2009, opened 2023-02-08
Task done

sergik94:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const remProp of action.keysToRemove) {
```
Viktor-Kost, line 15: I would give a name "key" in case you take `key` in `keysToRemove`
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          }
+        };
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
+    }
```
Viktor-Kost, line 27: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just `throw an Error` with text like `"unexpected action"`

Viktor-Kost (approved):
Nice!
Left some recommendations for you

PR #2010, opened 2023-02-08
solution

MSA1991:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
+        }
```
Anastasiia-Svintsova, line 25: You should add a default case for the switch/case construction to prevent bugs

Anastasiia-Svintsova (changes requested):
Good job! Fi one comment, please

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key of action.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
+        }
```
ViktorHavryliuk, line 25: This is not a default case, it's removeProperties case. Default case should handle situation when action.type is neither clear, addProperties nor removeProperties, by throwing an error "This action type is not supported", for example

ViktorHavryliuk (changes requested):

ViktorHavryliuk (approved):

PR #2013, opened 2023-02-09
Solution

nazarkharkevych:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => delete state[key]);
+    }
+
+    if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
+    }
```
DarkMistyRoom, line 19: Use switch statement if you have limited amount of conditions.

DarkMistyRoom (changes requested):
Good job, just use switch and don't forget about default case to handle errors

Volodymyr-Melnyk-Fron10der (approved):
Perfect✅

PR #2016, opened 2023-02-10
solution

the-tique:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: Try to use decomposing whenever it is possible 
```suggestion
  for (const { type, extraData, ... } of actions) {
```

Volodymyr-Melnyk-Fron10der (approved):
Great work, only one recommendation for future

PR #2018, opened 2023-02-11
Solution

dmytro-klepikov:


nazarmatsevych (approved):
It's a good practice to add default case for error handling.
Anyway good work!

PR #2019, opened 2023-02-11
Solution

vshepel:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (state[key]) {
+          delete state[key];
+        }
+      }
+    } else if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
```
Volodymyr-Melnyk-Fron10der, line 21: Transform it into `switch/case` syntax

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: use decomposition
```suggestion
  for (const { type .... } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (action.type === 'clear') {
+      for (const key of Object.keys(state)) {
+        delete state[key];
+      }
```
Volodymyr-Melnyk-Fron10der, line 12: You can write it with one line
```suggestion
      Object.keys(state).forEach(key => delete state[key])
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great, but can be better)

src/transformState.js
vshepel: That's what I thought )

src/transformState.js
vshepel: I would have used map, but since it wasn't in the lessons yet, I used the old for

vshepel: In this case, the best option is forEach ?

etojeDenys (approved):
almost done, just add default case

PR #2022, opened 2023-02-12
transformState solution

kateryna-babina:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete state[key];
+      }
+    }
+  }
+
+  return state;
```
NikitaStambul, line 26: Checklist №2. Use switch statement

NikitaStambul (changes requested):

BozhenaBodnar (approved):
well-done!!:))

src/transformState.js
```diff
@@ -3,9 +3,34 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        continue;
+    }
+  }
+
+  return state;
```
illiarerun, line 33: there's no need in returning state here, we're mutating original object

illiarerun (approved):

PR #2023, opened 2023-02-12
Solution

NataliiaHen:


ViktorHavryliuk (approved):
Good work, but you don't need newState, you can just modify state object directly

PR #2024, opened 2023-02-12
solution

toljan1:


tetiana-ivanchenko (changes requested):

IvanFesenko (changes requested):
Plase check my comments. Also add `default` case

VitaliyBondarenko1982 (approved):

PR #2031, opened 2023-02-13
solution

NikitaStambul:


arkushenko-anna (approved):

src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action['type']) {
```
Andr1yK, line 10:
```suggestion
    switch (action.type) {
```

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
Andr1yK, line 12: [CODE KNOWLEDGE]: Remember, if property key is a variable - use brackets object[key]. If it's called key - use dot access object.key.

Andr1yK (changes requested):

Andr1yK (approved):

PR #2032, opened 2023-02-13
Solution

IvanVaverchak:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+      }
+    }
+
+    if (obj.type === `clear`) {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
etojeDenys, line 28: try to rewrite it using the `switch\case` and don't forget to add default case

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const part in actions) {
+    const obj = actions[part];
```
etojeDenys, line 9: it's better to use `for of`

etojeDenys (changes requested):

OlEzhA0 (approved):

PR #2034, opened 2023-02-14
Solution

meltourme:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const char in actions) {
```
DarkMistyRoom, line 8: don't use `for in` loop for iterating over array

DarkMistyRoom (changes requested):
Good job, just fix one moment

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
+          }
+        }
+        break;
+
+      case `clear`:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
tetiana-ivanchenko, line 26: add default case

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+        }
+        break;
+
+      case `clear`:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return 'Error';
```
ViktorHavryliuk, line 27:
```suggestion
        throw new Error("This action is not supported");
```

ViktorHavryliuk (approved):
Good work. Have few suggestion how to improve it further, but it's optional

PR #2036, opened 2023-02-14
Stateful object task done

MarkBabich:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case '': //             does it fit to checklist p.4 ?
+        return '';
```
Yevhenii0536, line 12: what is this? maby you don't need it?

illiarerun: nope, it doesn't fit to checklist p4, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch#:~:text=encountered%20break.-,default,-Optional

illiarerun (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
+
+  return state;
```
illiarerun, line 29: also, there is no need in returning anything, task is about changing existing object

illiarerun (approved):

PR #2040, opened 2023-02-14
Transform State - Solution

Yevhenii0536:


MarkBabich (changes requested):
It works fine, but [Checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) p.4.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const itemOfState in state) {
+          delete state[itemOfState];
+        }
+        break;
+
+      default:
+        throw new Error('Error');
```
Andr1yK, line 27:
```suggestion
        throw new Error('Not supported type of action');
```

Andr1yK (approved):

PR #2041, opened 2023-02-14
task implemented ✅ 

andriibezkorovainyi:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        break;
+
+      case type === 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+    }
+  }
```
DenysDubas, line 28: it would be nice if you added a default value

DenysDubas (approved):
GJ!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (true) {
+      case type === 'addProperties':
```
illiarerun, line 12: nope, that's not how `switch/case` works. You should put type in first braces and then compare it's value by passing possible values to cases
```suggestion
    switch (type) {
      case  'addProperties':
```

src/transformState.js
illiarerun: agree, [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) #4

illiarerun (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
```
DenysDubas, line 11: type of what?

DenysDubas (changes requested):

src/transformState.js
andriibezkorovainyi: type of action...
image

illiarerun (approved):

PR #2043, opened 2023-02-14
created program use switch

vadosik15:


src/transformState.js
```diff
@@ -6,6 +6,26 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const dia of actions) {
```
illiarerun, line 9: what's `dia`? it's an array of `actions`, please use corresponding name - `action`

illiarerun (changes requested):

illiarerun (approved):

PR #2056, opened 2023-02-15
solution transformState

yuliia-vereshko:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const stateModified = state;
```
Andr1yK, line 8: what is the difference between stateModified and state?

Andr1yK (changes requested):

Andr1yK (approved):

PR #2057, opened 2023-02-15
Soliton done v1

yaroslav-vasilyev:


src/transformState.js
```diff
@@ -4,8 +4,39 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+const ADD_ACTION = 'addProperties';
+const REMOVE_ACTION = 'removeProperties';
+const CLEAR_ACTION = 'clear';
+
 function transformState(state, actions) {
   // write code here
```
Andr1yK, line 12: do u need this comment?

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+      case REMOVE_ACTION:
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case CLEAR_ACTION:
+        for (const key in newState) {
+          delete [key];
+          delete newState[key];
```
Andr1yK, line 30: what is the difference between newState and state?

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+
+      case REMOVE_ACTION:
+        for (const key of action.keysToRemove) {
+          delete newState[key];
+        }
+        break;
+
+      case CLEAR_ACTION:
+        for (const key in newState) {
+          delete [key];
```
Andr1yK, line 29: What are you deleting here?

Andr1yK (changes requested):

src/transformState.js
yaroslav-vasilyev: key of newState

Andr1yK: e.g. key === 'hello'

delete ['hello']

How does this relate to the state?

PR #2059, opened 2023-02-15
stateful-object

MaksymHuriev:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default :
+        break;
```
ivbutsykin, line 24: Will be nice to add `throw new Error` ...

ivbutsykin (approved):
Good job!

PR #2063, opened 2023-02-15
added solution

Vlados3611:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Andr1yK, line 10: check by checklist

Andr1yK (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
 
+      case 'clear':
```
Andr1yK, line 21: p4 [CODE STYLE]: switch/case should always have default case for error handling.

Andr1yK (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
 
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
Andr1yK, line 25: extremely risky

If I don't pass anything, I'm more likely to expect an error, or just that nothing will happen, but not that the whole object will be cleared

Andr1yK (changes requested):

Andr1yK (approved):

PR #2068, opened 2023-02-16
stateful_first

myroslava-kukhta:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+    }
```
witflash, line 25: `switch` should have `default` statement. You can check there errors or return nothing.

witflash (changes requested):
Add `default` statement for `switch/case` block

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
```
OleziO, line 10: Make const object for all case values

OleziO (approved):

PR #2069, opened 2023-02-16
solution

Nordexz:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete state[property];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
illiarerun, line 23: nope, you should do it for `clear` case and leave `default` case for working with errors

illiarerun (changes requested):

sergii-nosachenko (approved):
Good work 👍

PR #2071, opened 2023-02-18
Solution

dudar-y:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Volodymyr-Melnyk-Fron10der, line 8: use `for of` loop

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+      }
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Volodymyr-Melnyk-Fron10der, line 27: use switch/case + default

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const key in actions[i].extraData) {
+        state[key] = actions[i].extraData[key];
+      }
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key of actions[i].keysToRemove) {
+        if (state[key]) {
```
Volodymyr-Melnyk-Fron10der, line 17: you should check is key exist
```suggestion
        if (key in state) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Great job👍 Fix some comments pls

src/transformState.js
dudar-y: Ok. I can do that too

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const value of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: use decompose structure
```suggestion
  for (const { type, extraData, ... } of actions) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
One last change I would ask you about

Volodymyr-Melnyk-Fron10der (approved):
Ok👍

PR #2073, opened 2023-02-19
Stateful

OlegRak86:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const user of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+    if (user.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+
+    if (user.type === 'addProperties') {
+      Object.assign(state, user.extraData);
+    }
```
nazarmatsevych, line 23: Try to rewrite it using `switch/case` statement

nazarmatsevych (changes requested):
Good work,
pls check my comments

vlad-demchuk (approved):
GJ!

PR #2074, opened 2023-02-19
Task was done

anton-chornobai:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const elem of actions) {
```
VitaliyBondarenko1982, line 8:
```suggestion
  for (const action of actions) {
  const { type, extraData, keysToRemove } = action;
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const keyToRemove of elem.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    }
+
+    if (elem.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VitaliyBondarenko1982, line 23: better to use switch...case

VitaliyBondarenko1982 (changes requested):

OlEzhA0 (approved):

PR #2078, opened 2023-02-20
solution

serhiilysenko:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (action.type === 'removeProperties') {
+      for (const type of action.keysToRemove) {
+        delete state[type];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const del in state) {
+        delete state[del];
+      }
```
nazarmatsevych, line 22: It's better to use switch/case statement here as you rely on a single value of action.type.

nazarmatsevych (changes requested):

ViktorHavryliuk (changes requested):
Good, but it's would be better to use switch statement here instead of multiple ifs. And don't forget to add default case for error handling
```
switch (action.type){
  case 'addProperties':
  ...
}
```

Volodymyr-Melnyk-Fron10der (approved):
Cool✅

PR #2079, opened 2023-02-20
Solution for Transform State

ctk12:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      default:
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
```
qoa11a, line 23: you need to handle clear case as well (I understand it's handled in default, but default shouldn't be used for such reasons. In default you should throw an error that there's no such method or something like this)

qoa11a (changes requested):

src/transformState.js
ctk12: ok sir

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
```
polosanya, line 24: But a default case should be present too. Just throw some error in it
image

polosanya (changes requested):
Almost done!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, extraData, keysToRemove } of actions) {
```
```suggestion
  for (action of actions) {
  const { 
    type, 
    extraData, 
    keysToRemove,
 } = action
```
Your code should be not just short, but also understandable and descriptive

qoa11a (approved):

PR #2083, opened 2023-02-21
transform State js

valeriy-mikheyev:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    }
+  }
```
etojeDenys, line 30: try to rewrite it using the switch\case (don`t forget about default case)

PR #2085, opened 2023-02-22
change object

JasterHim:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (obj.type) {
+      case 'addProperties':
+        for (const key1 in obj.extraData) {
+          state[key1] = obj.extraData[key1];
+        }
```
sTorba24, line 13:
```suggestion
        Object.assign(state, action.extraData);
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
sTorba24, line 8:
```suggestion
  for (const action of actions) {
```

sTorba24: item from actions is action

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (obj.type) {
+      case 'addProperties':
+        for (const key1 in obj.extraData) {
+          state[key1] = obj.extraData[key1];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key2 of obj.keysToRemove) {
```
sTorba24, line 17:
```suggestion
        for (const key of obj.keysToRemove) {
```

sTorba24: it's other key :) Not you used before :)
For loop when you declare `const` it's creating variable for it own scope

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key2];
+        }
+        break;
+
+      case 'clear':
+        for (const key3 in state) {
```
sTorba24, line 23:
```suggestion
        for (const key of Object.keys(state)) {
```

sTorba24 (approved):
Good work! let's improve it

PR #2087, opened 2023-02-22
solution

DmytroHorshunov:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
```
etojeDenys, line 22: it's better to add "clear" case and in default just throw an error

etojeDenys (changes requested):

src/transformState.js
DmytroHorshunov: its new level for me

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Update your data');
```
DarkMistyRoom, line 27: It is not the best message for understanding where the problem is. "Action type is invalid" for example, is more descriptive

DarkMistyRoom (approved):

PR #2091, opened 2023-02-24
solution

YuriyLarin:


ViktorHavryliuk (changes requested):
Please use switch case instead of multiple ifs

YuriyLarin:
> Please use switch case instead of multiple ifs

For what?

ViktorHavryliuk (changes requested):
If you have questions it's better to ask them in chat (you may tag exact mentor if needed), so that you don't have to create new review request. Also here mentors won't see your question until they check your review again

loralevitska (approved):
👍

PR #2093, opened 2023-02-25
Solution

vanya-kalyenichenko:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
+    for (const key in i) {
```
IvanFesenko, line 9: Use correct naming. Second loop is redudant. Use switch/case for `action.type`
```suggestion
  for (const action of actions) {
```

IvanFesenko (changes requested):
Some refactor required

nazarmatsevych (approved):

PR #2095, opened 2023-02-26
transformState was created

Pavlo-Petrashevskyi:


src/transformState.js
```diff
@@ -4,8 +4,35 @@
...
+      for (const key of secondObj) {
+        delete state[key];
+      }
+    }
+
+    if (firstObj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
SanyaBratashchuk, line 32: use switch case instead of ifs

src/transformState.js
```diff
@@ -4,8 +4,35 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const firstObj = actions[i];
```
SanyaBratashchuk, line 10:
```suggestion
    const current = actions[i];
```

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+        }
+        break;
+
+      case current.type === 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  };
```
Volodymyr-Melnyk-Fron10der, line 35:
```suggestion
    switch (current.type) {
      case 'addProperties':
        const first = current.extraData;

        for (const key in first) {
          state[key] = first[key];
        }
        break;

      case 'removeProperties':
        const second = current.keysToRemove;

        for (const key of second) {
          delete state[key];
        }
        break;

      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
    }
  };
```

src/transformState.js
```diff
@@ -4,8 +4,37 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const current = actions[i];
+
+    switch (true) {
```
Volodymyr-Melnyk-Fron10der, line 12: add `default` to switch to throw an error if there is no action type like that

src/transformState.js
```diff
@@ -4,8 +4,37 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const current = actions[i];
```
Volodymyr-Melnyk-Fron10der, line 10: I suggest to use destructurization
```suggestion
    const { type, .... }= actions[i];
```

Volodymyr-Melnyk-Fron10der (changes requested):
Good, but you didn't get how to use switch case, I left some hint)

src/transformState.js
Pavlo-Petrashevskyi: done

src/transformState.js
Pavlo-Petrashevskyi: done

src/transformState.js
```diff
@@ -4,8 +4,36 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
YegorKorenkov, line 9: Please use for..of

src/transformState.js
```diff
@@ -4,8 +4,36 @@
...
+    const { type, ...Keys } = actions[i];
+
+    switch (type) {
+      case 'addProperties':
+        for (const key in Keys.extraData) {
+          state[key] = Keys.extraData[key];
```
YegorKorenkov, line 15: Please use Object.assign instead for

YegorKorenkov (changes requested):

src/transformState.js
Pavlo-Petrashevskyi: Done

src/transformState.js
Pavlo-Petrashevskyi: done

SerhiiKirik (approved):
Well done 🚀

PR #2097, opened 2023-02-27
js_stateful-object

Kirillets:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const command of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (command.type === 'addProperties') {
```
SanyaBratashchuk, line 9: use desctructuring

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
SerhiiKirik, line 23: add default and throw new Error in this case

SerhiiKirik (approved):
Good job 🔥 
Don't forget about comments

PR #2098, opened 2023-02-27
Solution

AnnaLiman:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+      for (const key in obj.extraData) {
+        state[key] = obj.extraData[key];
+      }
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete state[key];
+      }
+    }
```
loralevitska, line 26: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

loralevitska (changes requested):
Almost perfect) Let's improve

qoa11a (approved):

PR #2102, opened 2023-03-01
complete task

VladGraund:


src/transformState.js
```diff
@@ -4,8 +4,21 @@
...
+      Object.assign(state, key.extraData);
+    } else if (key.type === 'clear') {
+      for (const prop of Object.getOwnPropertyNames(state)) {
+        delete state[prop];
+      }
+    } else if (key.type === 'removeProperties') {
+      for (const val of key.keysToRemove) {
+        delete state[val];
+      }
+    }
```
VitaliyBondarenko1982, line 20: Try to avoid else if statements.
Code not good readable with them.
Better to use switch...case here

VitaliyBondarenko1982 (changes requested):

VitaliyBondarenko1982 (approved):

PR #2104, opened 2023-03-01
Solution

MaksymKos:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
IvanFesenko, line 17: This check is redudant. 
```suggestion
          delete state[key];
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        });
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
IvanFesenko, line 24: Same here
```suggestion
          delete state[key];
```

IvanFesenko (approved):
Also addd new line between cases for better readability

PR #2105, opened 2023-03-02
Solution

NVlasenko:


tetiana-ivanchenko (changes requested):
rewrite your solution according to checklist

src/transformState.js
```diff
@@ -1,11 +1,51 @@
...
 /**
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  const addProperties = 'addProperties';
+  const clear = 'clear';
+  const removeProperties = 'removeProperties';
```
sTorba24, line 10: better to use object here
const `actions = { addProperties: 'addProperties', ....}`

sTorba24 (approved):
Good work! Little comment about actions , better to store it as an object keys :)

PR #2107, opened 2023-03-03
first version

Vladimir-Zadorozhnyi:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can destructure the action to get all the needed props of it (type, extraData, keysToRemove etc)

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 22: image

polosanya (changes requested):
Almost done!

Volodymyr-Melnyk-Fron10der (approved):
NoddingSaturdayNightLiveGIF

PR #2108, opened 2023-03-03
Solution for Stateful object

julmak:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8:
```suggestion
  for (const { type, ... } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      for (const key of stateKeys) {
+        delete state[key];
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
```
Volodymyr-Melnyk-Fron10der, line 21: use switch/case

Volodymyr-Melnyk-Fron10der (changes requested):
almost there

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+    }
```
Volodymyr-Melnyk-Fron10der, line 27: always add default

src/transformState.js
Volodymyr-Melnyk-Fron10der: please destructurize action to make access to the properties easier

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        const stateKeys = Object.keys(state);
+
+        for (const key of stateKeys) {
```
Volodymyr-Melnyk-Fron10der, line 13: extra variable 

```suggestion
        for (const key of Object.keys(state)) {
```

Volodymyr-Melnyk-Fron10der (changes requested):

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Great 🔥

PR #2109, opened 2023-03-03
solution

vlad-fil:


etojeDenys (changes requested):
looks good, but let's rewrite it using  switch\case

etojeDenys (approved):
LGTM

just don`t forget to add the default case next time

PR #2110, opened 2023-03-04
Solution

92BS:
Used switch/case construction

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const kees in state) {
+          delete state[kees];
+        }
+        break;
+    }
```
witflash, line 25: The switch/case need to have `default` statement.
You can return nothing here or handle some errors.

witflash (approved):
Good job 👍🏻

PR #2112, opened 2023-03-04
Solution

HannaVasylieva:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
etojeDenys, line 24: i think we can remove here this `if`

etojeDenys (changes requested):
almost perfect, just add the default case, and just throw the error there

mykhalenych (approved):

PR #2113, opened 2023-03-05
'Solution'

denlubn:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
nazarmatsevych, line 20: In this case it's better to use `switch/case` statement as you rely only on action.type value

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
```
nazarmatsevych, line 12: Use destructuring of object properties

nazarmatsevych (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
```
loralevitska, line 8: Fix previous comment about destructuring
```suggestion
  actions.forEach({ type,  extraData, keysToRemove }=> {
```

loralevitska (changes requested):
If you have any questions, please ask them in fe_chat and tag mentors. It will be faster, because your answers in GitHub will be visible to mentor only after you re-request review
 
 Add default case

nazarmatsevych (approved):

PR #2115, opened 2023-03-06
Solution

antonina-klishch:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const char of action.keysToRemove) {
```
Volodymyr-Melnyk-Fron10der, line 12: it's a key actually
```suggestion
      for (const key of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: destructurize action
```suggestion
  for (const {type, ....} of actions) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there💪🏼

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+      default:
+        for (const key1 in state) {
+          delete state[key1];
+        }
+        break;
```
loralevitska, line 22: The default value is used in case no value fits. I don't think clearing the data is a good idea. Just return state

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
```
loralevitska, line 17:
```suggestion
      case 'addProperties':
        Object.assign(state, extraData);
        break;
        
      case 'removeProperties':
        for (const key of keysToRemove) {
          delete state[key];
        }
        break;
        
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, extraData, keysToRemove } of actions) {
```
loralevitska, line 8:
```suggestion
  for (const { type, extraData, keysToRemove, clear } of actions) {
```

Write code for `clear` case

loralevitska (changes requested):
Almost perfect)

DarkMistyRoom (approved):

PR #2116, opened 2023-03-06
Solution

hyrych-max:


src/transformState.js
```diff
@@ -6,6 +6,20 @@
...
+  actions.forEach(action => {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => {
+        if (key in state) {
+          delete state[key];
+        }
+      });
+    } else if (action.type === 'clear') {
```
polosanya, line 18: For now, your code looks a bit complicated. Use switch case instead of else...ifs
image

polosanya (changes requested):

OlEzhA0 (approved):

PR #2117, opened 2023-03-07
Solution

NadiiaR:


Virass (changes requested):
Cool but i think in this case little bit better use `for of` loop and use `switch case `instead of an `if` condition construction

Viktor-Kost (approved):

PR #2118, opened 2023-03-07
Solution

yurii-shkrobut-m:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+
+      if (object[key] === 'addProperties') {
+        Object.assign(state, object['extraData']);
+      }
+
+      if (object[key] === 'removeProperties') {
+        for (const item of object['keysToRemove']) {
+          delete state[item];
+        }
+      }
```
loralevitska, line 25: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
loralevitska, line 9:
```suggestion
  for (const action of actions) {
```

And also destructure `action`

loralevitska (changes requested):
Almost done)

src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+    for (const key in action) {
+      switch (action[key]) {
+        case 'clear':
+          for (const key2 in state) {
+            delete state[key2];
+          }
+          break;
+
+        case 'addProperties':
+          Object.assign(state, action['extraData']);
```
SerhiiKirik, line 19:
```suggestion
          Object.assign(state, action.extraData);
```

src/transformState.js
```diff
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    for (const key in action) {
```
SerhiiKirik, line 10: Try to do it with 1 loop

src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+      switch (action[key]) {
```
SerhiiKirik, line 11:
```suggestion
      switch (action.type) {
```

SerhiiKirik (changes requested):
Good job 🚀 
Let's try to fix comments & improve your code 🧑‍💻

SerhiiKirik (approved):
LGTM 🔥

PR #2119, opened 2023-03-07
add solution

allagutsul27:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[prop];
+        }
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VitaliyBondarenko1982, line 29: use switch..case instead if

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      const data = Object.entries(action.extraData);
+
+      for (const item of data) {
+        state[item[0]] = item[1];
+      }
```
VitaliyBondarenko1982, line 14:
```suggestion

       Object.assign(state, action.extraData);
```

VitaliyBondarenko1982 (changes requested):

mykhalenych (approved):

PR #2121, opened 2023-03-08
Solution

Vaskovics:


src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const obj = actions[i];
+    const { type } = obj;
```
IvanFesenko, line 10: use for of loop
```suggestion
  for (action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+
+    if (type === 'addProperties') {
+      Object.assign(state, obj.extraData);
+    }
```
IvanFesenko, line 14: Use switch for type instead if

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+
+      for (const keyState in state) {
+        for (let k = 0; k < itemsRemove.length; k++) {
+          const itemRemove = itemsRemove[k];
+
+          if (state[keyState] === state[itemRemove]) {
+            delete state[keyState];
+          }
+        }
+      }
```
IvanFesenko, line 27: use for of loop too. No need in 2 loops

IvanFesenko (changes requested):
Check my comments

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const obj = action;
+    const { type } = obj;
+
```
SanyaBratashchuk, line 11:
```suggestion
    const {type, extraData, keysToRemove } = action;
```

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        const itemsRemove = keysToRemove;
```
DarkMistyRoom, line 17: why do you need to copy this variable?

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          if (state[keyToRemove]) {
+            delete state[keyToRemove];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
DarkMistyRoom, line 29: switch/case should always have default case for error handling.

DarkMistyRoom (changes requested):
Almost there

loralevitska (changes requested):
Almost perfect) Let's improve

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of keysToRemove) {
+          if (state[keyToRemove]) {
```
Volodymyr-Melnyk-Fron10der, line 18: yea, it works but it's better to check is the key present in object like `keyToRemove in state`

Volodymyr-Melnyk-Fron10der (approved):
Good job👍

PR #2123, opened 2023-03-08
Solution

Oleksandr-Hundertailo:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    }
+
+    if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
+  return (state);
```
dashaprotsko: Since the function should modify the state directly, there is no need to return the state. Remove the `return (state);` line.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      Object.assign(state, actions[i].extraData);
```
dashaprotsko, line 14: Use the spread operator instead of `Object.assign()` to copy the object without mutating the original object. Replace with `state = { ...state, ...actions[i].extraData };`

dashaprotsko (approved):
Well done :)

PR #2124, opened 2023-03-09
Solution

pindel-kateryna:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Viktor-Kost, line 9: it's better to use `for of` loop

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+    if (actions[i].type === 'removeProperties') {
+      for (const key in state) {
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          if (key === actions[i].keysToRemove[j]) {
+            delete state[key];
+          }
+        }
+      }
+    }
+  }
```
Viktor-Kost, line 29: rewrite to `switch`. It will make your code cleaner

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+    }
+
+    if (actions[i].type === 'removeProperties') {
+      for (const key in state) {
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
+          if (key === actions[i].keysToRemove[j]) {
+            delete state[key];
+          }
+        }
+      }
```
Viktor-Kost, line 27: you can simplify 
```suggestion
        for (const key of action.keysToRemove) {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        }
```

Viktor-Kost (changes requested):

etojeDenys (changes requested):
almost done, just don`t forget to always add `default case` (you can just throw the error there)

src/transformState.js
```diff
@@ -6,6 +6,34 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const elem of actions) {
```
Volodymyr-Melnyk-Fron10der, line 9: use destructurization in future
```suggestion
  for (const { type, extraData, ... } of actions) {
```

Volodymyr-Melnyk-Fron10der (approved):
Well done👍

PR #2125, opened 2023-03-09
solution

Yurko-Komanitskyi:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
Viktor-Kost, line 28: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
```
Viktor-Kost, line 19:
```suggestion
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
```

src/transformState.js
```diff
@@ -4,8 +4,29 @@
...
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
Viktor-Kost, line 14:
```suggestion
Object.assign(state, action.extraData);
```

Viktor-Kost (changes requested):
Nice!

Need some improvements

src/transformState.js
yuron-maker: I added an empty default because output to the console is not allowed

AlisaVlasova (approved):

PR #2126, opened 2023-03-10
solution

IamLoren:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
etojeDenys, line 20: let`s rewrite it using switch\case (and do not forget about default case)

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        });
+        break;
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
etojeDenys, line 25: let`s just throw the error here

etojeDenys (changes requested):
😎

etojeDenys (approved):
awesome

IamLoren:
😇

PR #2130, opened 2023-03-13
Object transforming func added

DenysKrysanov:


VenomMB (approved):
GJ

jstmpelowycz (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw Error;
```
jstmpelowycz, line 27: this'd better
```suggestion
        throw new Error(...);
```

PR #2132, opened 2023-03-13
solved

PashaMalyshkin:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const prop of actions) {
+    switch (prop.type) {
+      case 'addProperties':
+        for (const extra in prop.extraData) {
+          state[extra] = prop.extraData[extra];
+        }
```
PavloReutskiy, line 13:
```suggestion
        Object.assign(state, prop.extraData);
```
I'm not sure but maybe it will be better than loop

PavloReutskiy (approved):

src/transformState.js
PashaMalyshkin: Yeah, i agree with u, thanks

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'clear':
+        for (const removedKey in state) {
```
```suggestion
        for (const key in state) {
```
or
```suggestion
        for (const keyToClear in state) {
```

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+          delete state[removedKey];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const removedKey of action.keysToRemove) {
```
jstmpelowycz, line 21:
```suggestion
        for (const key of action.keysToRemove) {
```
or
```suggestion
        for (const keyToRemove of action.keysToRemove) {
```

jstmpelowycz (approved):

PR #2135, opened 2023-03-13
create transformState

VenomMB:


Esteban4769 (changes requested):
Good job, need just some small adjusment to make.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const currentObject of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    switch (currentObject.type) {
+      case 'addProperties':
+        Object.assign(state, currentObject.extraData);
+        break;
+      case 'removeProperties':
```
jstmpelowycz, line 13: add spaces between cases
```suggestion
      case 'addProperties':
        Object.assign(state, currentObject.extraData);
        break;

      case 'removeProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        currentObject.keysToRemove.forEach(elem => delete state[elem]);
```
jstmpelowycz, line 14: why `elem`? it's a key we'd like to remove

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+      default:
+        return null;
+    }
+  }
+
+  return state;
 }
 
+function toClearState(state) {
```
jstmpelowycz, line 28:
```suggestion
function clearState(state) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+  for (const elem in state) {
```
jstmpelowycz, line 29: `elem` :harold:
image

jstmpelowycz (changes requested):

src/transformState.js
VenomMB: it was my favorite cigarette)))

VinnieJ-2k20 (approved):

PR #2136, opened 2023-03-13
added transform function

PavloReutskiy:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const removable of version.keysToRemove) {
+          delete state[removable];
+        }
+        break;
+
+      default:
+        break;
```
Stambuliak, line 27: maybe it is better to use `throw` instead of break?

Stambuliak (approved):
GJ

src/transformState.js
jstmpelowycz: you can implement it

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, version.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const removable of version.keysToRemove) {
```
jstmpelowycz, line 21: `removable` is too abstract
```suggestion
        for (const key of version.keysToRemove) {
```

jstmpelowycz (approved):

PavloReutskiy:
all done! I noticed it for future

PR #2137, opened 2023-03-13
done

victor-buldenko:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      for (const key in obj) {
+        delete obj[key];
+      }
+    } else if (element['type'] === 'removeProperties') {
+      element['keysToRemove'].forEach(prop => {
+        delete obj[prop];
+      });
+    } else {
+      Object.assign(obj, element['extraData']);
+    }
```
Volodymyr-Melnyk-Fron10der, line 21: use switch/case/default

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const obj = state;
+
+  actions.forEach(element => {
```
Volodymyr-Melnyk-Fron10der, line 10: destructurize element

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const obj = state;
```
Volodymyr-Melnyk-Fron10der, line 8: do copy, now obj === state

Volodymyr-Melnyk-Fron10der (changes requested):
almost👍

src/transformState.js
victor-buldenko: `const obj = { ...state };`
if i return a copy, tests fail 11/13
image
What i have to do?

src/transformState.js
victor-buldenko: Task:
`state` is an object. You are supposed to add, change, or delete its
  properties instead of creating a new object.
  
 maybe I don't understand something? 
 `destructurize element` what for?

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Well done 🔥

PR #2138, opened 2023-03-13
Done

Ilia991:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const { type, keysToRemove, extraData } = actions[i];
+
+    if (type === 'removeProperties') {
```
AlisaVlasova, line 11: use switch case instead of if

AlisaVlasova (changes requested):
Good job

AlisaVlasova (approved):

PR #2139, opened 2023-03-13
write code

Batick3737:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
agniya-i, line 9: it would be better to use `switch/case` instead of `if`

agniya-i (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can destructure the action to get all the needed props of it (type, extraData, keysToRemove).
It will make your code cleaner

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 22: image

polosanya (changes requested):
Almost done!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        default:
+          break;
```
agniya-i, line 24:
```suggestion
          throw new Error("Unknown action type");
```
better to add error handling for default case

agniya-i (changes requested):

polosanya (approved):
Well done!

PR #2142, opened 2023-03-14
Transforming function was created

Vasyl-Zhyliakov:


src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+    }
```
SerhiiKirik, line 31: add default case

SerhiiKirik (approved):
Well done 🚀

PR #2145, opened 2023-03-14
Solution

KosolapovAnatolii:


Vlad-Lischenovskiy (approved):
Well done!!!

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const st in state) {
+          delete state[st];
+        }
+        break;
+      default:
+        break;
```
jstmpelowycz, line 26:
```suggestion
        break;

      default:
        break;
```

jstmpelowycz: also, you could've thrown an error

jstmpelowycz (approved):

PR #2146, opened 2023-03-14
solution

glots-a:


vinnykalisa (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(element => {
```
jstmpelowycz, line 8: it's rather an action, not just an element
```suggestion
  actions.forEach(action => {
```

jstmpelowycz (approved):
great job 🔥

PR #2148, opened 2023-03-14
add solution

Vlad-Lischenovskiy:


Anna-Kuzmenko (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      default:
+        break;
```
jstmpelowycz, line 28:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
        action.keysToRemove.forEach(key => {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        });
        break;

      case 'clear':
        for (const key in state) {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        }
        break;

      default:
        break;
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+            delete state[key];
+          }
+        });
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
jstmpelowycz, line 25: you actually don't need these ifs

PR #2150, opened 2023-03-14
solution

Olekk17:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```
kukla1989, line 19: you should add only code that you add manually, delete package.json from commit

kukla1989 (approved):

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const item of actions) {
+    switch (item.type) {
+      case 'clear':
+        for (const prop in state) {
```
jstmpelowycz, line 12:
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+          delete state[prop];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of item.keysToRemove) {
+          if (state.hasOwnProperty(keyToRemove)) {
+            delete state[keyToRemove];
+          }
```
jstmpelowycz, line 21: do you need this validation?

src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const item of actions) {
```
```suggestion
  for (const action of actions) {
```

jstmpelowycz (changes requested):

jstmpelowycz (changes requested):
oh, didn't see it — where is your default case?)

jstmpelowycz (approved):

PR #2152, opened 2023-03-14
solution

Anastasia2403:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  const changedState = state;
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(changedState, action.extraData);
+        continue;
```
Rage1502, line 14: Interesting solution with continue, but I think it would be better to use break in such cases, and default to display an unknown command message.

Rage1502 (changes requested):

jstmpelowycz (approved):
good job 🔥

PR #2153, opened 2023-03-14
Solution

Anna-Kuzmenko:


Anastasia2403 (approved):
Looks good 👍

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case ('removeProperties'):
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        }
+        break;
+
+      default:
+        return state;
```
jstmpelowycz, line 27: perhaps, it'd be better to throw an error
however, we've not been told what to do, so OK 🙂

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+    }
+  }
+
+  return state;
```
jstmpelowycz, line 31: you don't have to return state, we just mutate it

jstmpelowycz (approved):

PR #2154, opened 2023-03-14
State transformation function implemented

roshirin:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const transformedState = state;
```
makandriival, line 8: Nice and neat job! Why do you need a variable here? The link to the state obj is unnecessary in this case in my opinion
Screenshot 2023-03-15 at 15 37 34

makandriival (changes requested):

jstmpelowycz (changes requested):
good job! just one thing to improve

jstmpelowycz (approved):

PR #2155, opened 2023-03-15
create function for stateful object

kukla1989:


danylo-dobrovolskyi (commented):
Good job!

danylo-dobrovolskyi (approved):

src/transformState.js
```diff
@@ -5,7 +5,17 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
jstmpelowycz, line 9: this is the case to use switch-case
why? we have the limited number of conditions — 3, as we have 3 action types

jstmpelowycz (changes requested):
the checklist, p.2

VinnieJ-2k20 (approved):
Good job 👍

PR #2156, opened 2023-03-15
solution

artem-kyrylenko:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+  const errorMessage = 'Error';
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case `removeProperties`:
+        for (const value of action.keysToRemove) {
```
jstmpelowycz, line 17: would be better to call it so as we iterate over keys
```suggestion
        for (const key of action.keysToRemove) {
```

jstmpelowycz (changes requested):

src/transformState.js
artem-kyrylenko: fixed

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw errorMessage;
+    }
+  }
+
+  return state;
```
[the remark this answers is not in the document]
artem-kyrylenko: fixed

jstmpelowycz (approved):
LGTM 🔥

PR #2157, opened 2023-03-15
Solution trasnform states

Evgenb000:


vksazonov (approved):
Well done!

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
jstmpelowycz, line 27: break-statement is meaningless in default cases, you could've thrown an error

jstmpelowycz (approved):
good job!

PR #2159, opened 2023-03-15
Solve the task using switch case

Hamuud:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'removeProperties': {
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+      }
+
+      case 'clear': {
+        for (const data in state) {
```
jstmpelowycz, line 23: what's `data`? let's name it another way

jstmpelowycz (changes requested):
the checklist, p.4

jstmpelowycz (changes requested):
so, what's about the checklist, p.4?))

Hamuud:
I add this error for next task, but forgot about this, sorry Dmytro🥲

jstmpelowycz (approved):
🔥

PR #2164, opened 2023-03-15
Solution

vksazonov:


ymys1 (approved):
Well done!)
 Correct your code and it will be even better)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const operation of actions) {
+    switch (true) {
```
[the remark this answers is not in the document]
jstmpelowycz: agree, let's fix it

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+      case operation.type === 'addProperties': {
+        Object.assign(state, operation.extraData);
+        break;
+      }
+
+      case operation.type === 'removeProperties': {
+        for (const property of operation.keysToRemove) {
```
jstmpelowycz, line 16: I'd rather name it `key`
```suggestion
        for (const key of operation.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+      case operation.type === 'clear': {
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      }
+
+      default: {
+        break;
+      }
```
jstmpelowycz, line 31: you've not been told what to do in the default case, however, it's still better to throw an error

jstmpelowycz (changes requested):

jstmpelowycz (approved):

PR #2165, opened 2023-03-15
function works

BondarenkoDmitriy:


VladislavaT (approved):
Good job!
I can`t see something to goes wrong

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties' :
+        for (const keyForState of action.keysToRemove) {
+          if (state[keyForState]) {
+            delete state[keyForState];
+          }
```
jstmpelowycz, line 24: you don't need this validation

jstmpelowycz (changes requested):
see the checklist, p.4

jstmpelowycz (approved):

PR #2167, opened 2023-03-15
Solution

Zad0rnik:


AP-Solution (approved):
looks good, working solution

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const prop of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+    switch (prop.type) {
+      case 'addProperties':
+        for (const extra in prop.extraData) {
+          state[extra] = prop.extraData[extra];
+        }
```
jstmpelowycz, line 13: let's use `Object.assign`

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const del in state) {
```
jstmpelowycz, line 17: we iterate over object keys
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          delete state[del];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const del of prop.keysToRemove) {
```
jstmpelowycz, line 23: we iterate over keys array
```suggestion
        for (const key of prop.keysToRemove) {
```

jstmpelowycz (changes requested):
where's your default case?)

jstmpelowycz (approved):

PR #2169, opened 2023-03-15
transform state solution

VladislavaT:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default: return null;
+    }
+  }
+
+  return state;
```
yapelenua, line 33: maybe you dont need to return state

yapelenua (approved):

src/transformState.js
VladislavaT: idk
but I thought that we should to return transformed state

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+
```
```suggestion
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        for (const key of action.keysToRemove) {
+          if (state[key]) {
+            delete state[key];
+          }
```
jstmpelowycz, line 19: you don't need this validation

src/transformState.js
jstmpelowycz: nah :D
no return statement required, let's remove it
Screenshot 2023-03-16 at 09 57 48

jstmpelowycz (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
```suggestion
        break;

      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
```

jstmpelowycz (approved):

PR #2171, opened 2023-03-15
Develop

AP-Solution:


iHateAntichrist (approved):
Good job

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
```
jstmpelowycz, line 13: don't forget to format code properly
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        for (let key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
jstmpelowycz, line 24: up to you, but you could've thrown an error (putting break-statement in default cases is meaningless)

jstmpelowycz (changes requested):

liliia-livandovskaya (approved):

PR #2172, opened 2023-03-15
Solution

Goto-Jesus:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
jstmpelowycz, line 13: use `Object.assign`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
```
jstmpelowycz, line 20: add spaces between cases
```suggestion
        }
        break;

      case 'clear':
```

jstmpelowycz (changes requested):
also, the checklist, p.4

src/transformState.js
Goto-Jesus: thx

jstmpelowycz (approved):

PR #2174, opened 2023-03-15
Develop

Kavon-Skela:


DimaDvm (approved):

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  try {
```
jstmpelowycz, line 8: try-catch is not for this purpose
if you have a condition to throw an error, you throw it, and then you catch it, decide what to do, how your program will proceed, etc.

here, you do not throw anything in try-block

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    for (const action of actions) {
+      switch (action['type']) {
```
jstmpelowycz, line 10: the checklist, p.6

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        case 'addProperties':
+          for (const fieldToAdd in action['extraData']) {
+            state[fieldToAdd] = action['extraData'][fieldToAdd];
+          };
```
jstmpelowycz, line 14: 2 points:
- let's use `Object.assign`
- the checklist, p.6

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          break;
+        case 'removeProperties':
+          for (const keyToRemove of action['keysToRemove']) {
```
jstmpelowycz, line 17: > the checklist, p.6

same here

jstmpelowycz (changes requested):
the checklist, p.6, p.4

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        };
+        break;
+      case 'clear':
```
jstmpelowycz, line 18: add spaces between cases
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
        for (const keyToRemove of action.keysToRemove) {
          delete state[keyToRemove];
        };
        break;

      case 'clear':
```

jstmpelowycz (approved):

PR #2177, opened 2023-03-15
solution

iHateAntichrist:


asmd564 (approved):

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
VinnieJ-2k20, line 8: Let's use `for .. of` loop here, you don't really need index value.

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+      case 'addProperties':
+        const extraData = action.extraData;
+
+        for (const key in extraData) {
+          if (extraData.hasOwnProperty(key)) {
+            state[key] = extraData[key];
+          }
+        }
+        break;
+      case 'removeProperties':
```
VinnieJ-2k20, line 21:
```suggestion
        break;
        
      case 'removeProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+        const keysToRemove = action.keysToRemove;
+
+        for (let j = 0; j < keysToRemove.length; j++) {
```
VinnieJ-2k20, line 24: Same, use `for ... of`

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      default:
+        break;
```
VinnieJ-2k20, line 40: Add some error message or Error throwing in case of invalid input.

VinnieJ-2k20 (changes requested):
You are on the right way, but some changes needed.

src/transformState.js
```diff
@@ -5,7 +5,44 @@
...
+
+        for (const removeKey of keysToRemove) {
+          const key = removeKey;
+
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      case 'clear':
```
liliia-livandovskaya, line 33:
```suggestion
        break;
        
      case 'clear':
```

liliia-livandovskaya (approved):

PR #2182, opened 2023-03-16
switch case solution

TolikSerhiychuk:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const relatedStateKey of action.keysToRemove) {
+          if (state[relatedStateKey]) {
+            delete state[relatedStateKey];
+          }
```
jstmpelowycz, line 18: you don't need this validation

jstmpelowycz (approved):

PR #2183, opened 2023-03-16
Solution

Zibi95:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete initialState[key];
+      }
+    } else {
+      for (const key in initialState) {
+        delete initialState[key];
+      }
+    }
+  });
+
+  return initialState;
```
Radoslaw-Czerniawski, line 26: Also this doesn't do much here as you are mutating the original object, it will either way be changed with the difference you return it from this function. If you did something like 
```
 const state = { x: 1 };

 transformState(state, [
   { type: 'addProperties', extraData: { yet: 'another property' } }
   { type: 'clear' },
   { type: 'addProperties', extraData: { foo: 'bar', name: 'Jim' } }
 ]);
``` 
The state object will be changed after the function call, in your case you are also piping it through the function.

Radoslaw-Czerniawski (approved):
It's a good implementation, I see you have no problems with understanding and working on collections. Some comments from my side for future reference, nothing wrong with your pr but rather some insight from me :)

PR #2187, opened 2023-03-16
Solution

DanielSorokowski:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          state[key] = object.extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        object.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      default:
```
solaryasha, line 21: No, let's define specific action `clear`.
Right now, if I provide some random `action` it will clear the object.

solaryasha (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return 'Error, Wrong action provided';
```
solaryasha, line 26: It's better to throw an error
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw

solaryasha (approved):

PR #2188, opened 2023-03-16
solution

swiftbuddy03:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
```
tetiana-ivanchenko, line 23: image

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
tetiana-ivanchenko, line 9: destructure the action object

tetiana-ivanchenko (changes requested):

src/transformState.js
swiftbuddy03: I have no idea what must be there, can tell you me?

src/transformState.js
swiftbuddy03: and what to destruct? actions ?

PR #2190, opened 2023-03-16
solution

mzaloha:


etojeDenys (changes requested):
GJ, but let's rewrite it using switch\case

VitaliyBondarenko1982 (approved):

PR #2191, opened 2023-03-17
solution

maksym-kostetskyi:


package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
SerhiiKirik, line 8: Let's use for of here
Also, use a switch case for this

SerhiiKirik (changes requested):
Well done 🚀 
Let's improve your code 🧑‍💻

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    for (const value of Object.values(object)) {
```
DarkMistyRoom, line 9: why do you need this nested loop? you just need to check value of `action.type`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      switch (value) {
+        case 'addProperties':
+          for (const [key1, value1] of Object.entries(object.extraData)) {
+            state[key1] = value1;
+          };
```
DarkMistyRoom, line 14: I see that you really like loops, but would be more effectively to use `Object.assign` here

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          break;
+        case 'removeProperties':
+          for (const key2 of object.keysToRemove) {
+            delete state[key2];
+          };
```
DarkMistyRoom, line 19: here it's ok)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          break;
+        case 'clear':
+          for (const key3 of Object.keys(state)) {
+            delete state[key3];
+          };
```
DarkMistyRoom, line 24: It is ok, but also you can just use `for in` to iterate through object keys

DarkMistyRoom (changes requested):

mykhalenych (approved):

PR #2192, opened 2023-03-18
develop

nata-prog123:


src/transformState.js
```diff
@@ -4,8 +4,27 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
+    if (actions[key].type === 'addProperties') {
```
TarasHoliuk, line 10: that's the perfect case to use `switch`

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+      Object.assign(state, actions[key].extraData);
+    }
+
+    if (actions[key].type === 'removeProperties') {
+      for (const t in actions[key].keysToRemove) {
```
TarasHoliuk, line 15: `t` is a bad variable name

src/transformState.js
```diff
@@ -4,8 +4,27 @@
...
+        if (actions[key].keysToRemove[t] in state) {
+          delete state[actions[key].keysToRemove[t]];
+        }
+      }
+    }
+
+    if (actions[key].type === 'clear') {
+      for (const n in state) {
```
TarasHoliuk, line 23: `n` is a bad variable name

src/transformState.js
```diff
@@ -4,8 +4,27 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
```
TarasHoliuk, line 9: use `for of` loop to iterate through an array

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,28 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
```
TarasHoliuk, line 9:
```suggestion
  for (const action of actions) {
```

TarasHoliuk (approved):

PR #2193, opened 2023-03-19
Solution

Tarasova09:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
```
etojeDenys, line 30: let's rewrite it using switch\case

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #2194, opened 2023-03-20
Solution

skinnyjazz:


generatedman (approved):
good job!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
naziamoff: You should add some empty lines between cases for improving readability
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const key of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+            delete state[key];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
naziamoff: you're iterating the keys in state here, right? Do you need to check whether the key is in the state?

naziamoff (changes requested):
looks good, but there are some small suggestions

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error(`Invalid action type: ${action.type}`);
```
snuber-w, line 29: looks interesting 👍 
nice

snuber-w (commented):
GJ 🔥

snuber-w (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
```
silchyn, line 20:
```suggestion
        }
        
        break;
```

silchyn (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+      case 'clear':
+        for (const key in state) {
```
33Orange, line 23: Consider using `Object.keys(state).forEach(key => delete state[key]);` instead of the for-in loop, to improve code readability.

33Orange (approved):
Well done!

For improve:
You can use array methods instead `for in` loop.

naziamoff (approved):

PR #2196, opened 2023-03-20
Solution

yarik-bachynskyi:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 9: you can use destructurization here, do not forget about it

Volodymyr-Melnyk-Fron10der (approved):
Well done👍

PR #2198, opened 2023-03-21
commit

bohdanmayor:


src/transformState.js
```diff
@@ -6,6 +6,22 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (let i = 0; i < actions.length; i++) {
```
mykhalenych, line 10: better will be 
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -6,6 +6,22 @@
...
+    if (actions[i]['type'] === 'addProperties') {
```
mykhalenych, line 11: also better will be change if/else if ... to switch case/...

mykhalenych (approved):

PR #2201, opened 2023-03-22
solution

Fedos17:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      Object.assign(state, ofType.extraData);
+    }
+
+    if (ofType.type === 'removeProperties') {
+      for (let i = 0; i < ofType.keysToRemove.length; i++) {
+        delete state[ofType.keysToRemove[i]];
+      }
+    }
+
+    if (ofType.type === 'clear') {
```
vlad-demchuk, line 21: let's use `switch` when we compare one value

vlad-demchuk (changes requested):

YegorKorenkov (approved):
Looks good

PR #2202, opened 2023-03-22
solution

tkalindima:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
OlEzhA0, line 9: use switch case to avoid a lot of `if` statements

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[value];
+        };
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        };
+        break;
+    }
```
Volodymyr-Melnyk-Fron10der, line 25: always add default handler

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: you can destructurize action right away

Volodymyr-Melnyk-Fron10der (changes requested):

PR #2203, opened 2023-03-22
Solution

vlad-liakh:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const newState = state;
```
etojeDenys, line 8: redundant

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    for (const key in action.keysToRemove) {
+      delete addKey[action.keysToRemove[key]];
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
etojeDenys, line 22: lets rewrite it using switch\case

etojeDenys (changes requested):

PR #2204, opened 2023-03-24
Solution

NikaNika12:


src/transformState.js
```diff
@@ -6,6 +6,28 @@
  */
 function transformState(state, actions) {
   // write code here
+
+  for (const action of actions) {
```
tetiana-ivanchenko, line 10:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+    const { type } = action;
+
+    switch (type) {
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
```
tetiana-ivanchenko, line 15: use for in, it help avoid you one loop

src/transformState.js
```diff
@@ -6,6 +6,28 @@
...
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const item of action.keysToRemove) {
+          delete state[item];
+        }
+        break;
+      default:
+        break;
+    }
```
tetiana-ivanchenko, line 27: image

tetiana-ivanchenko (changes requested):

src/transformState.js
NikaNika12: заменила

src/transformState.js
NikaNika12: добавила значение default

src/transformState.js
sTorba24: it was about removing blank lines and comments

sTorba24 (approved):
Well done! Don't forget to clean comments before review :) For production comments doesn't make sense, so we don't need to keep it for deploy

PR #2205, opened 2023-03-24
Mission completed!

RatuhWD:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        delete result[deletedProperties[d]];
+      }
+    } else {
+      const clear = Object.keys(result);
+
+      for (let c = 0; c < clear.length; c++) {
+        delete result[clear[c]];
+      }
+    }
+  }
```
sTorba24, line 30: use switch: case for action.type
also you can use `.forEach` method instead of standart for
to add properties use `Object.assign(obj, second obj)`
you should check if type === 'clear' and only then delete all field
by the way, you can do it just using result = {}

sTorba24 (changes requested):
Good work! Lets improve it

PR #2209, opened 2023-03-27
Transform-State

Soi4An:


VitaliyBondarenko1982 (changes requested):
good work.
let's replace if on switch case

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case (action.type === 'addProperties'):
```
DarkMistyRoom, line 10:
```suggestion
    switch (action.type) {
      case 'addProperties':
```

DarkMistyRoom: looks simpler, isn't it?

DarkMistyRoom (approved):
Good job

PR #2210, opened 2023-03-27
solution

MikiOOOO:


DanielSorokowski (approved):
You can throw an error for default case

michalcywka (changes requested):
Checklist point 4

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
+        break;
+      }
+
+      case `removeProperties`: {
```
d-wasilewski: you don't need template string here

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2213, opened 2023-03-28
solution

ghostargg:


DanielSorokowski (changes requested):
u can use switch case here :D

DanielSorokowski (approved):

michalcywka (changes requested):
Checklist point 2

src/transformState.js
```diff
@@ -6,6 +6,37 @@
...
+      case 'clear':
+        if (Object.keys(state).length !== 0) {
+          for (const key in state) {
+            delete state[key];
+          }
+        }
+        break;
+
+      default:
+        return 0;
```
d-wasilewski: throw an error here instead of returning

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2217, opened 2023-03-29
solution v1

bjeziorek:


src/transformState.js
```diff
@@ -6,6 +6,43 @@
...
+        for (let j = 0; j < keysToRemove2.length; j++) {
+          delete robotState[keysToRemove2[j]];
+        }
+        break;
+      default:
+        return {};
+    }
+  }// for
+
+  return robotState;
```
natalia-klonowska, line 45: "state is an object. You are supposed to add, change, or delete its properties instead of creating a new object"

natalia-klonowska (changes requested):
"state is an object. You are supposed to add, change, or delete its properties instead of creating a new object"

src/index.html
d-wasilewski, line 1: this file is not necessary

src/transformState.js
```diff
@@ -6,6 +6,42 @@
...
+        const keysToRemove2 = Object.keys(state);
+
+        for (let j = 0; j < keysToRemove2.length; j++) {
+          delete state[keysToRemove2[j]];
+        }
+        break;
+      default:
+        return {};
+    }
+  }// for
```
d-wasilewski: leftover

src/transformState.js
```diff
@@ -6,6 +6,42 @@
...
+
+      case 'removeProperties':
+
+        for (let j = 0; j < robotAction.keysToRemove.length; j++) {
+          delete state[robotAction.keysToRemove[j]];
+        }
+        break;
+
+      case 'clear':
+        const keysToRemove2 = Object.keys(state);
```
d-wasilewski: give more descriptive names to variables

d-wasilewski: and dont use numbers like variable1, variable2

src/transformState.js
```diff
@@ -6,6 +6,42 @@
  */
 function transformState(state, actions) {
   // write code here
+  const robotActions = Object.assign({}, actions);
+  const a = Object.keys(robotActions);
```
d-wasilewski: The Object.assign() method used to create a copy of the actions object is unnecessary since the actions object is not modified inside the function. You can simply use the actions object directly.

src/transformState.js
```diff
@@ -6,6 +6,42 @@
...
+
+  for (let actionNumber = 0; actionNumber < a.length; actionNumber++) {
+    const robotAction = robotActions[actionNumber];
+
+    switch (robotAction.type) {
+      case 'addProperties':
+
+        const keys = Object.keys(robotAction.extraData);
+
+        for (let j = 0; j < keys.length; j++) {
```
d-wasilewski: he for...in loop in the addProperties case can be replaced with a simpler for...of loop that iterates over the keys of the extraData object directly.
sth like `for (const key of Object.keys(action.extraData)) {`

d-wasilewski (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,39 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let actionNumber = 0;
+    actionNumber < Object.keys(actions).length;
+    actionNumber++) {
```
d-wasilewski: you dont need to use Object.keys here, just use the length of the actions, also, its a good practice to use `i` as an iterator variable name in the for loops

d-wasilewski (changes requested):

d-wasilewski (approved):

PR #2218, opened 2023-03-29
solution

danielpopek94:


kacper-lyczba (approved):

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete state[key]);
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+
+      default:
+        return console.error('Error, probably the wrong parameter was specified.');
```
d-wasilewski: Either return something or log it
```suggestion
        console.error('Error, probably the wrong parameter was specified.');
```

d-wasilewski (changes requested):

src/transformState.js
danielpopek94: lint doesn't allow console.error so I returned a regular string

d-wasilewski (approved):

PR #2223, opened 2023-03-29
Solution

kacper-lyczba:


src/transformState.js
```diff
@@ -6,6 +6,35 @@
...
 function transformState(state, actions) {
   // write code here
+  for (let n = 0; n < actions.length; n++) {
+    switch (actions[n].type) {
+      case 'addProperties':
+        const actKeys = Object.keys(actions[n].extraData);
+        const actVals = Object.values(actions[n].extraData);
+
+        for (let i = 0; i < actKeys.length; i++) {
+          state[actKeys[i]] = actVals[i];
```
d-wasilewski: use `Object.entries` inside of for...of loop isntead

d-wasilewski (changes requested):

src/transformState.js
kacper-lyczba: I didn't know it could be used that way, I will look into that

src/transformState.js
```diff
@@ -6,6 +6,34 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let n = 0; n < actions.length; n++) {
+    switch (actions[n].type) {
+      case 'addProperties':
+        const actEntries = Object.entries(actions[n].extraData);
+
+        for (const [key, value] of actEntries) {
```
d-wasilewski: Again, use full names in variables or just do this
```suggestion
        for (const [key, value] of Object.entries(actions[n].extraData)) {
```

d-wasilewski (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,34 @@
...
+
+      case 'removeProperties':
+        for (let i = 0; i < actions[n].keysToRemove.length; i++) {
+          delete state[actions[n].keysToRemove[i]];
+        }
+
+        break;
+
+      case 'clear':
+        const stateKeys = [ ...Object.keys(state) ];
```
d-wasilewski: Redundant variable creation: The stateKeys variable is not necessary, as you can simply iterate over the keys of the state object directly

d-wasilewski (changes requested):

src/transformState.js
kacper-lyczba: The variable is necessary because it is not a reference, it copies the keys table
When iterating over the keys of the state object with a for loop, it outputs the following error:

>   ● Should clear the state                                                                                                                                                                                                       
                                                                                                                                                                                                                                 
    expect(received).toEqual(expected) // deep equality

    - Expected
    + Received

    - Object {}
    + Object {
    +   "name": "Jim",
    + }

      161 |
      162 |   expect(state)
    > 163 |     .toEqual({});
          |      ^
      164 | });
      165 |
      166 | test('Should not fails when clearing an empty state', () => {

      at Object.<anonymous> (src/transformState.test.js:163:6)

kacper-lyczba: I will look into other ways of iterating through the object

kacper-lyczba: Worked perfectly fine while iterating from the end of the array instead; going to apply the approach

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+        for (let i = 0; i < actions[n].keysToRemove.length; i++) {
+          delete state[actions[n].keysToRemove[i]];
+        }
+
+        break;
+
+      case 'clear':
+        for (let i = (Object.keys(state).length - 1); i >= 0; i--) {
+          delete state[Object.keys(state)[i]];
+        }
```
d-wasilewski: regarding your comments - you don't need a variable because you have to modify this `state` anyway
You don't need to use Object.keys just to get the length of the state, you can just use state.length or for...in loop. Alternatively you can probably just do state = {} to clear the state. Try some of it but don't use Object.keys here and don't add redundant variables

d-wasilewski (changes requested):

src/transformState.js
kacper-lyczba: Cannot use state = {} as state in this assignment is a reference passed as a parameter rather than a clone
And with state.length... back when I tried doing that on the Mate website, it outputted errors, that's why I've avoided doing that, I will try a for...in loop

d-wasilewski (changes requested):
Looks good! Just add the default case (checklist 4)

d-wasilewski (approved):

PR #2228, opened 2023-03-30
Solution

sylcym:


d-wasilewski (changes requested):
checklist point 4

d-wasilewski (changes requested):
[4] [CODE STYLE]: switch/case should always have default case for error handling.

d-wasilewski (approved):

PR #2232, opened 2023-03-31
Solution

RomanOstrous:


src/transformState.js
```diff
@@ -1,11 +1,29 @@
...
-module.exports = transformState;
+'use strict';
+
+/**
+ * @param {Object} state
+ * @param {Object[]} actions
+ */
+function transformState(state, actions) {
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
agniya-i: it would be better to use `switch/case` instead of `if/else`, in case of the limited amount of conditions

agniya-i (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,34 @@
...
+      case 'addProperties':
+        const extraData = action.extraData;
+
+        for (const [key, value] of Object.entries(extraData)) {
+          state[key] = value;
+        }
+        break;
+
+      case 'removeProperties':
+        const keysToRemove = action.keysToRemove;
```
nazarmatsevych, line 19: You can also you destructuring of action object

nazarmatsevych (approved):
Good work

PR #2240, opened 2023-04-03
Develop

Ir-ra:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
slavanosatskyi:
```suggestion
        state = {}
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      default:
+        return state;
```
slavanosatskyi:
```suggestion
        return;
```

slavanosatskyi (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove } = actions;
    
    switch (type) {
```
You can also destructure all the needed props to make your code a bit cleaner)

polosanya (approved):
Good job!

PR #2241, opened 2023-04-03
js_stateful-object

stnayn:


src/transformState.js
```diff
@@ -4,8 +4,25 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
+    if (i.type === 'addProperties') {
```
AlisaVlasova: use switch case

AlisaVlasova (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,26 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
+  }
```
anstsot, line 26: it would be great to return state
```suggestion
  }
  return state;
```

src/transformState.js
```diff
@@ -4,8 +4,26 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const i of actions) {
+    switch (i.type) {
```
anstsot, line 10: don't forget to add default case for switch/case

anstsot (approved):
Good job, approved

PR #2242, opened 2023-04-03
Solution

Viktoriia6666:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
SerhiiKirik, line 25: Alsways try to add default case

package.json
```diff
@@ -16,7 +16,7 @@
   "license": "GPL-3.0",
   "devDependencies": {
     "@mate-academy/eslint-config": "*",
-    "@mate-academy/scripts": "^0.3.5",
+    "@mate-academy/scripts": "^1.2.8",
```

SerhiiKirik (approved):
Great 🔥

PR #2244, opened 2023-04-05
Solution

DefaultGitUser:


src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+        break;
+      default:
+        clear(state);
+    };
+  };
+};
+
+function addProperties(state, obj) {
+  Object.assign(state, obj.extraData);
+}
```
anstsot, line 24: if you're using small atomic functions it would be better to return the value at the end
```suggestion
function addProperties(state, obj) {
  Object.assign(state, obj.extraData);
  return state;
}
```

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+  for (const obj of actions) {
+    switch (true) {
+      case obj.type === 'addProperties':
+        addProperties(state, obj);
+        break;
+      case obj.type === 'removeProperties':
+        removeProperties(state, obj);
+        break;
+      default:
+        clear(state);
```
anstsot, line 17: it's better not to clear by default
```suggestion
      case obj.type === 'clear':
        clear(state);
      default:
        break;
```

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+    };
+  };
```
anstsot, line 19: and don't forget to return value at the end of function
```suggestion
  };
  return state;
```

anstsot (approved):
Good job, approving. But I've left several comments to improve the code

src/transformState.js
DefaultGitUser: Дякую, в цьому є сенс

PR #2246, opened 2023-04-07
stateful-object

IvanOrinchak:


src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+          }
+        });
+        break;
 
+      case 'clear':
+        Object.keys(state).map(function(key) {
+          delete state[key];
+        });
+        break;
+    }
```
agniya-i, line 33: add default case 
Screenshot 2023-04-07 at 20 20 00

agniya-i (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,35 @@
...
+        break;
 
+      case 'clear':
+        Object.keys(state).map(function(key) {
+          delete state[key];
+        });
+        break;
+
+      default:
+        break;
```
Volodymyr-Melnyk-Fron10der: you should throw the Error in default case, like 'Unknown action type'

Volodymyr-Melnyk-Fron10der (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,35 @@
...
 
+      case 'clear':
+        Object.keys(state).map(function(key) {
+          delete state[key];
+        });
+        break;
+
+      default:
+        return 'Unknown action type';
```
Anastasiia-Svintsova, line 32:
```suggestion
        throw new Error('Unknown action type');
```

Anastasiia-Svintsova (approved):

PR #2247, opened 2023-04-07
Solution

AnastassiaH:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+          delete state[item];
+        }
+      });
+    }
+
+    if (action['type'] === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      };
+    }
```
nazarmatsevych: Good work but to improve readibility let's rewrite it using switch/case statement.

You can also destructure action object

nazarmatsevych (changes requested):

nazarmatsevych (approved):

PR #2251, opened 2023-04-09
solution

mariuszgit:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  let tempState = state;
```
DorotaLeniecDev, line 8: This is unnecessary, as you are assigning here the reference to the state object so you are opereting on a state object.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+  for (const singleAction of actions) {
```
DorotaLeniecDev: I think that you could simply name it action
```suggestion
  for (const action of actions) {
```

DorotaLeniecDev (changes requested):
Good solution, 2 comments left

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+      case 'removeProperties':
+        singleAction.keysToRemove.forEach(el => {
+          delete tempState[el];
+        });
+        break;
+
+      case 'clear':
+        for (const el in tempState) {
+          delete tempState[el];
+        };
```
DorotaLeniecDev, line 28: Add default case for error handling
_[CODE STYLE]: switch/case should always have default case for error handling._

src/transformState.js
DorotaLeniecDev: Ok understandable

DorotaLeniecDev (approved):
GJ!

PR #2254, opened 2023-04-10
solution ♡

dseq0:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
SerhiiKirik, line 28: always add default case

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
```
SerhiiKirik: Try to use Object.assign

SerhiiKirik (changes requested):
Well done 🚀 
Let's try to improve your code 🧑‍💻

SerhiiKirik (approved):
Good job 🔥

PR #2255, opened 2023-04-10
done

maxonshev:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const value of action.keysToRemove) {
+        if (state[value]) {
+          delete state[value];
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
etojeDenys: better to rewrite it using switch\case instead of if`s, it will improve readability a lot

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,48 @@
...
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
 }
 
+// function transformState(state, actions) {
```
loralevitska: Remove comments

src/transformState.js
```diff
@@ -5,7 +5,48 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
loralevitska: Destructure `action`
```suggestion
  for (const { type, extraData, keysToRemove } of actions) {
```

loralevitska (changes requested):
Almost perfect) Let's improve

etojeDenys (approved):
lgtm

PR #2260, opened 2023-04-11
Solution

OlyaProkofieva:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (true) {
+      case action.type === 'addProperties':
```
```suggestion
    switch (action.type) {
      case 'addProperties':
```

tetiana-ivanchenko: please, read the documentation and fix all cases

src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+          delete state[key];
+        }
+        break;
+      case action.type === 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
tetiana-ivanchenko: image

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Error');
```
tetiana-ivanchenko: please, use more meanfull message, f.e. "Invalid action type"

tetiana-ivanchenko (approved):

PR #2263, opened 2023-04-13
Solution

yuliia-nudyk:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
```

polosanya (changes requested):
Good job! Let's improve your solution according to the checklist. Use switch...case instead of `if`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
witflash, line 9: If you use `switch/case` block it should have a `default` statement. You can handle any error there or just return nothing.

witflash (approved):
Well done 👍🏻 
Check my comment for your future work.

PR #2266, opened 2023-04-14
tranformState

oleh-dehtiarenko:


src/transformState.js
```diff
@@ -4,8 +4,36 @@
...
+        break;
+
+      case 'clear':
+        for (const prop in state) {
+          delete state[prop];
+        }
+        break;
+
+      default:
+        break;
```
loralevitska, line 32: throw error or return initial state

loralevitska (approved):
Well done)

PR #2268, opened 2023-04-14
done

MayerLansky:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
Volodymyr-Melnyk-Fron10der, line 8: use destructure assignment 
```suggestion
  for (const { type, keysToRemove, ... } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (act.type === 'addProperties') {
+      for (act.type of actions) {
```
Volodymyr-Melnyk-Fron10der: what is the purpose of this for loop? I don't see

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        Object.assign(state, act.extraData);
+      }
+    }
+
+    if (act.type === 'removeProperties') {
+      for (const type of act.keysToRemove) {
```
Volodymyr-Melnyk-Fron10der: same

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete state[type];
+      }
+    }
+
+    if (act.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Volodymyr-Melnyk-Fron10der, line 24: use switch/case/default structure, it's more useful in this case

Volodymyr-Melnyk-Fron10der (changes requested):

vlad-demchuk (approved):

PR #2271, opened 2023-04-16
Transform State task

VKonoplianyi:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
Volodymyr-Melnyk-Fron10der, line 28: I would throw some error here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
```suggestion
  for (const { type, extraData } of actions) {
```

Volodymyr-Melnyk-Fron10der (changes requested):
almost there

SerhiiKirik (approved):
Great 🔥

PR #2274, opened 2023-04-18
Solution

artyomwhite:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
Viktor-Kost, line 25: it's a good practice to add a default behavior at switch statement. You may either throw an error or break

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2275, opened 2023-04-18
fist commit

vladimirshub:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
loralevitska, line 22: Destructure `action`
```suggestion
  for (const { type, .... } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
loralevitska:
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;
        
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
```

loralevitska: Add `default` case (throw error or return state)

loralevitska (changes requested):
Almost perfect) Let's improve

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, ...action } of actions) {
+    switch (type) {
+      case 'addProperties':
```

OleziO (changes requested):

loralevitska (approved):
good job!

PR #2276, opened 2023-04-18
solution

LiliiaDanylenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
vlad-demchuk, line 28: image

vlad-demchuk (changes requested):

LiliiaDanylenko:
Can it be?

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
VitaliyBondarenko1982, line 27:
```suggestion
      return throw new Error('Error message'):
```
it can be something like thate

VitaliyBondarenko1982 (approved):

PR #2277, opened 2023-04-18
Solution

ychaiun:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya: You can desctructure the action to make your code a bit cleaner
`const { type, extraData, keysToRemove } = actions`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 26: Add default case for error handling
image

polosanya (changes requested):
Almost perfect!

agniya-i (approved):

PR #2278, opened 2023-04-19
Stateful object resolved

IzzetYusufov:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (action.type === 'removeProperties') {
+      for (const remove of action.keysToRemove) {
+        delete state[remove];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const clear in state) {
+        delete state[clear];
+      }
```
IvanFesenko, line 22: For case like this switch/case looks better

IvanFesenko (approved):

PR #2279, opened 2023-04-19
v01

Yuliia-Kyrychenko:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const step of actions) {
```
```suggestion
  for (const action of actions) {
```

DarkMistyRoom: or we can decompose it 
```suggestion
  for (const { type, extraData, ... } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    switch (step.type) {
+      case 'addProperties':
+        for (const addData in step.extraData) {
+          state[addData] = step.extraData[addData];
+        }
```
DarkMistyRoom, line 13: would be shorter to use `Object.assign`

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        break;
+      case 'removeProperties':
+        for (const removeData of step.keysToRemove) {
+          delete state[removeData];
+        }
+        break;
+      case 'clear':
```
DarkMistyRoom, line 20:
```suggestion
        break;
        
      case 'removeProperties':
        for (const removeData of step.keysToRemove) {
          delete state[removeData];
        }
        break;
        
      case 'clear':
```

DarkMistyRoom (approved):
Good job! Pay attention to the comments

PR #2280, opened 2023-04-20
Solution

Anastasiia-Shahoval:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
tetiana-ivanchenko, line 27: [CODE STYLE]: switch/case should always have default case for error handling.

tetiana-ivanchenko (changes requested):

VitaliyBondarenko1982 (approved):

PR #2282, opened 2023-04-20
Solution

vitalii-fedusov:


src/transformState.js
```diff
@@ -6,6 +6,21 @@
...
+      for (const j of i.keysToRemove) {
+        delete state[j];
+      }
+    } else if (i.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
```
etojeDenys: better to use `switch` instead of if`s

etojeDenys: Using `switch` instead of `if` statements can be preferable in certain situations for a few reasons:

1. Readability: When you have multiple conditions to check, `switch` statements can make your code more readable by grouping similar cases together.

2. Efficiency: `switch` statements can often be more efficient than `if` statements when you have a large number of cases to check. This is because `switch` statements use jump tables or hash tables for lookups, which can be faster than checking each condition in turn.

3. Default case handling: `switch` statements have a built-in mechanism for handling cases that don't match any of the other conditions. This can be useful in situations where you need to handle unexpected inputs.

However, `if` statements still have their place and may be preferable in some cases, such as when you need to evaluate more complex conditions or when you want to perform additional operations before or after the condition is evaluated. Ultimately, the choice between `switch` and `if` statements depends on the specific requirements of your code.

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,26 @@
...
+        for (const j of action.keysToRemove) {
+          delete state[j];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
Anastasiia-Svintsova, line 25: Follow the checklist, please. Add a default case
image

Anastasiia-Svintsova (changes requested):
Not all requirements from the checklist have been met. Fix it, please

OleziO (approved):

PR #2283, opened 2023-04-21
Solution

ollavka:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+  const REMOVE_PROPERTIES_TYPE = 'removeProperties';
+  const CLEAR_PROPERTIES_TYPE = 'clear';
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ADD_PROPERTIES_TYPE:
+        Object.assign(state, action.extraData);
+        break;
+      case REMOVE_PROPERTIES_TYPE:
+        action.keysToRemove.forEach(key => delete state[key]);
```
anstsot, line 18: it will be better to open the brackets and write the code here because now you're returning `delete state[key]`

anstsot (approved):

PR #2284, opened 2023-04-23
Solution

anpiet:


src/transformState.js
```diff
@@ -1,11 +1,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    for (const key in actions[i]) {
+      if (actions[i][key] === 'addProperties') {
+        Object.assign(state, actions[i]['extraData']);
+      };
+
+      if (actions[i][key] === 'removeProperties') {
+        for (const key1 in actions[i]['keysToRemove']) {
```
AndrzejPec: according to the checklist: 1. [CODE STYLE]: don't use for in loop for iterating over array

src/transformState.js
```diff
@@ -1,11 +1,27 @@
 'use strict';
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    for (const key in actions[i]) {
```
AndrzejPec: according to checklist 5. [CODE STYLE]: Nested loops === EVIL.

Avoid evil ;)

src/transformState.js
```diff
@@ -1,11 +1,27 @@
...
+      if (actions[i][key] === 'addProperties') {
```
AndrzejPec: according to the checklist: 2. [CODE STYLE]: Use switch statement if you have limited amount of conditions.

AndrzejPec (changes requested):
I've found 3 cases of checklist violation, see if there be more, I'm not sure :)

src/transformState.js
```diff
@@ -1,11 +1,22 @@
...
+    switch (key.type) {
+      case 'addProperties':
+        Object.assign(state, key.extraData);
+        break;
+
+      case 'removeProperties':
+        key.keysToRemove.forEach(el => delete state[el]);
+        break;
+
+      default:
```
AndrzejPec: Consider changing this condition to "case" and use "default" to throw an error (checklist -> "should always have default case for error handling.")

AndrzejPec (commented):
lookin' all good, only one small improvement suggested :)

src/transformState.js
anpiet: good point! It's already changed.

choeqq (approved):
Good job ✅

PR #2285, opened 2023-04-24
1

Vitalii-Maistrenko:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Anastasiia-Svintsova: Use for...of loop here

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
+
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+
+        for (let j = 0; j < action.keysToRemove.length; j++) {
```
Anastasiia-Svintsova: Same

Anastasiia-Svintsova (changes requested):
- Check the checklist one more time, please 
image

src/transformState.js
anstsot: not fixed

src/transformState.js
anstsot: not fixed

anstsot (changes requested):
previous comments are not fixed. maybe you forgot to push the changes?
and please remove all unnecessary empty lines

src/transformState.js
```diff
@@ -4,8 +4,30 @@
...
+      case 'addProperties':
+        return Object.assign(newState, action.extraData);
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => {
+          if (newState.hasOwnProperty(key)) {
+            delete newState[key];
+          }
+        });
+        return newState;
```
agniya-i: You can also move `return newState;` after `switch` to not repeat it each time

agniya-i (changes requested):

anstsot (approved):
Good job, approved!

PR #2288, opened 2023-04-26
Solution

artpetrenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const current of actions) {
+    if (current.type === 'addProperties') {
```
anstsot: please use `switch/case`
image

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const current of actions) {
+    switch (current.type) {
+      case 'addProperties': Object.assign(state, ...[current.extraData]);
+        break;
```
anstsot, line 11: check the code formatting
```suggestion
      case 'addProperties':
        Object.assign(state, ...[current.extraData]);
        break;
```

anstsot (approved):
Good job, approved!

PR #2289, opened 2023-04-26
Solution

generatedman:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+        delete state[removed];
+      }
+    } else if (action.type === 'clear') {
+      for (const prop in state) {
+        delete state[prop];
+      }
+    }
+  }
+
+  return state;
```
33Orange: You don't need to return the `state` object at the end of the function since the changes are made directly to the original object.

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
33Orange: Use Switch case instead `if else` statement if you have limited amount of conditions.

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const removed of action.keysToRemove) {
```
33Orange: Don't use nested `for of` loops.

33Orange (changes requested):
- Change `if else` to `switch`
- Avoid nested for loops. You can use array methods instead.
- This function should change the state but not return anything else.

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        clear(state);
+        break;
+    }
+  }
+
+  return state;
+}
+
+function remove(object, val) {
+  for (const key of val) {
```
Anna-Tkachenko: Check the variable name, it's not clear what is `val` here

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2290, opened 2023-04-26
Solution

SerhiiShapovalenko:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

OleziO (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
 function transformState(state, actions) {
-  // write code here
+  const actionsType = {
+    add: 'addProperties',
+    remove: 'removeProperties',
+    clear: 'clear',
+  };
+
+  for (const action of actions) {
+    switch (action.type) {
```
anstsot, line 15: don't forget about default case for `switch/case`

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,40 @@
...
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case actionsType.clear:
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
```
loralevitska, line 30: Of course, this key is in the object, but you get it from the object through a loop. Remove redundant checker

src/transformState.js
```diff
@@ -5,7 +5,40 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case actionsType.add :
+        Object.assign(state, action.extraData);
+        break;
+
+      case actionsType.remove:
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
loralevitska, line 22: Same here

loralevitska (approved):

PR #2292, opened 2023-04-26
finished js_stateful-object

EvhenIvchenko:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
OleziO: Replace with switch case.

OleziO (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        }
+        break;
+
+      default:
+        // do nothing for unknown action types
+        break;
+    }
+  }
+
+  return state;
```
sTorba24, line 32: Avoid returning the mutated state directly. Instead, return a copy of the state.

sTorba24 (approved):
Well done!

PR #2293, opened 2023-04-27
Solution

vintwp:


src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+      clear(state);
+    }
+
+    if (type === 'removeProperties') {
+      removeProperties(state, rest);
+    }
+
+    if (type === 'addProperties') {
+      addProperties(state, rest);
+    }
```
OleziO: Replace with switch case construction.

OleziO (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
nazarmatsevych:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+    const { type, ...rest } = actions[i];
+
+    switch (true) {
```
nazarmatsevych:
```suggestion
    switch (type) {
```

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+      case type === 'clear':
```
nazarmatsevych:
```suggestion
      case 'clear':
```

Fix all such cases

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const { type, ...rest } = actions[i];
```
nazarmatsevych:
```suggestion
    const { type, ...rest } = action;
```

nazarmatsevych (changes requested):
Almost done, 
pls check my comments

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, ...rest } = action;
+
+    switch (type) {
+      case 'clear':
```
OleziO, line 12: Move all cases to the object (additional)

OleziO (approved):

PR #2295, opened 2023-04-28
Solution

BiletskyiVolodymyr:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+  for (const key of action) {
+    switch (key.type) {
+      case 'addProperties':
+        for (const property in key.extraData) {
+          state[property] = key.extraData[property];
+        }
+        break;
+
+      case 'removeProperties':
```
OleziO, line 14: Move all cases to the object.

OleziO (changes requested):

src/transformState.js
BiletskyiVolodymyr: I don't understand what you mean. Can you clarify?

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const action = Object.values(actions);
+
+  for (const key of action) {
```
Anastasiia-Svintsova:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    switch (key.type) {
+      case 'addProperties':
+        for (const property in key.extraData) {
+          state[property] = key.extraData[property];
+        }
```
Anastasiia-Svintsova: Use Object.assign instead

Anastasiia-Svintsova (changes requested):
Good job! Let's add some improvements

src/transformState.js
Volodymyr-Melnyk-Fron10der: not fixed 
image

Volodymyr-Melnyk-Fron10der (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
```
agniya-i, line 8: One more thing from previous comments - please rename `key` to `action` 🙂

Checklist: 

Screenshot 2023-04-29 at 18 28 21

agniya-i (approved):
I approved, but pay attention to small comment

PR #2297, opened 2023-04-29
Solution

BeeGooDee:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let action = 0; action < actions.length; action++) {
```
witflash: It's not an `action` really, it's just an iterator's temp variable. So you can name it just `i` for example.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    const act = actions[action];
```
witflash: And then here you can use full name for `action`

witflash (changes requested):
Good job!

But could you correct a little your naming, please?

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    if (action.type === 'addProperties') {
```
anstsot: please use `switch/case` construction. do not forget about `default` case
image

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 12: Try to move all cases to the object.

OleziO (approved):

PR #2298, opened 2023-05-01
Solution

vladislav-tsymbalist:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
mykhalenych, line 25: Add a default case in the switch statement to handle unknown action types

mykhalenych (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
```
agniya-i, line 23:
```suggestion
      default:
        throw new Error('Unknown action type'):
```

it also would be nice to throw an error here

agniya-i (approved):

PR #2300, opened 2023-05-01
add transformState solution

DidenkoDenis306:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
+      case (action.type === 'addProperties'):
+        Object.assign(state, action.extraData);
+        break;
+
+      case (action.type === 'removeProperties'):
+        for (const key of action.keysToRemove) {
```
33Orange, line 15: Use array methods here instead nested loop. `Nested loops === EVIL`

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case (action.type === 'clear'):
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return null;
```
33Orange: [CODE STYLE]: switch/case should always have default case for error handling.

33Orange (changes requested):
The code works, but doesn't match several items in the checklist.

Anna-Tkachenko (approved):
⭐️

PR #2301, opened 2023-05-02
add task solution

maksym-nemera:


generatedman (approved):
good job!

naziamoff (changes requested):
let's avoid using things like `values[0]` here. It makes your code difficult to read as I don't understand what exactly is stored there

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const data of action.keysToRemove) {
```
naziamoff: is this `data` or a `key`?

naziamoff (approved):
looks good to me 👍

PR #2302, opened 2023-05-02
obj

diana-tuz:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key in actions) {
+    if (actions[key].type === 'addProperties') {
```
vlad-demchuk: use switch case here

vlad-demchuk: it will be clearer when you compare one value

vlad-demchuk (changes requested):

YegorKorenkov (approved):
Looks good

PR #2303, opened 2023-05-03
Solution

Oleksii-Hrytsenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i].type;
+
+    if (action === 'clear') {
+      for (const stateKey in state) {
+        delete state[stateKey];
+      }
+    } else if (action === 'addProperties') {
```
PolinaRomanenko: Use the correct action property name. Replace `else if (action === 'addProperties')` with `else if (action.type === 'addProperties')`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      const addObject = actions[i].extraData;
```
PolinaRomanenko: Use the correct action property name. Replace `const addObject = actions[i].extraData;` with `const addObject = action.extraData;`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i].type;
+
+    if (action === 'clear') {
```
PolinaRomanenko: Use the correct action property name. Replace `if (action === 'clear')` with `if (action.type === 'clear')`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      for (const stateKey in state) {
+        delete state[stateKey];
+      }
+    } else if (action === 'addProperties') {
+      const addObject = actions[i].extraData;
+
+      for (const addKey in addObject) {
+        state[addKey] = addObject[addKey];
+      }
+    } else if (action === 'removeProperties') {
```
PolinaRomanenko: Use the correct action property name. Replace `else if (action === 'removeProperties')` with `else if (action.type === 'removeProperties')`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i].type;
```
PolinaRomanenko: Use action object instead of action type directly. Replace `const action = actions[i].type;` with `const action = actions[i];`

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        delete state[stateKey];
+      }
+    } else if (action === 'addProperties') {
+      const addObject = actions[i].extraData;
+
+      for (const addKey in addObject) {
+        state[addKey] = addObject[addKey];
+      }
+    } else if (action === 'removeProperties') {
+      const removeArray = actions[i].keysToRemove;
```
PolinaRomanenko: Use the correct action property name. Replace `const removeArray = actions[i].keysToRemove;` with `const removeArray = action.keysToRemove;`

PolinaRomanenko (changes requested):
For case like this switch/case looks better

src/transformState.js
Oleksii-Hrytsenko: fixed

src/transformState.js
Oleksii-Hrytsenko: fixed

src/transformState.js
Oleksii-Hrytsenko: fixed

src/transformState.js
Oleksii-Hrytsenko: fixed

src/transformState.js
Oleksii-Hrytsenko: fixed

src/transformState.js
Oleksii-Hrytsenko: fixed

Ivan-Shvorak (approved):
LGTM

PR #2306, opened 2023-05-05
task done

zisuan-od:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const obj of actions) {
```
DarkMistyRoom:
```suggestion
  for (const action of actions) {
```

DarkMistyRoom: try to use more descriptive names

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+
+    if (obj.type === 'removeProperties') {
+      for (const key in state) {
+        for (const arrElem of obj.keysToRemove) {
+          if (arrElem === key) {
+            delete state[key];
+          }
+        }
+      }
+    }
```
DarkMistyRoom: image

src/transformState.js
DarkMistyRoom: would be good to decompose `action` here

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+
+    if (obj.type === 'removeProperties') {
+      for (const key in state) {
+        for (const arrElem of obj.keysToRemove) {
+          if (arrElem === key) {
+            delete state[key];
+          }
+        }
```
DarkMistyRoom: no need to use a nested loop. You can just iterate by `keysToRemove` and delete `state[keyToRemove]`

DarkMistyRoom (changes requested):
please look at the checklist before making the PR

src/transformState.js
```diff
@@ -6,6 +6,31 @@
...
+
+    switch (whatToDo) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(function(keyToRemove) {
+          delete state[keyToRemove];
+        });
```
IvanFesenko, line 20: its better to use arrow function
```suggestion
        action.keysToRemove.forEach((keyToRemove) => {
          delete state[keyToRemove];
        });
```

IvanFesenko (approved):

PR #2311, opened 2023-05-06
transform state

pos1bl:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        for (const key of actions[i].keysToRemove) {
+          delete state[key];
+        }
+        break;
+      default:
+        return 'Something went wrong';
+    }
+  }
+
+  return state;
```
Volodymyr-Melnyk-Fron10der: Returning the state object is not necessary, as the function is modifying the original object directly. Just remove this line.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Volodymyr-Melnyk-Fron10der: use `for of`

Volodymyr-Melnyk-Fron10der: and use de-structure assignment

Volodymyr-Melnyk-Fron10der (changes requested):

vlad-demchuk (approved):

PR #2315, opened 2023-05-07
Solution

bodyabandura:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
+    }
+  }
+
+  return state;
```
sergii-nosachenko: You don't need to return anything from this function, just modify the original object by action.

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      for (const data in action.extraData) {
+        state[data] = action.extraData[data];
+      }
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
+    }
```
sergii-nosachenko: Use `switch...case` instead of `if...else if...` - it's easier to read and maintain.

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const data in action.extraData) {
+        state[data] = action.extraData[data];
+      }
```
sergii-nosachenko: Use `Object.assign` method here - no need to use the loop

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
```
sergii-nosachenko: You can use `forEach` method here too

sergii-nosachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,20 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
```
sergii-nosachenko, line 16:
```suggestion
        action.keysToRemove.forEach(key => delete state[key]);
```

src/transformState.js
```diff
@@ -5,7 +5,20 @@
...
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+    }
```
sergii-nosachenko, line 20:
```suggestion
        break;
        
      default:
        break;
    }
```

It is considered good practice to always include a "default" case.

sergii-nosachenko (approved):
Good work 👍🏻 
Commented

PR #2317, opened 2023-05-08
add task solution

AleksandrTsivilin:


src/transformState.js
```diff
@@ -6,6 +6,43 @@
...
+  }
+
+  return state;
+}
+
+function add(object, data) {
+  for (const key in data) {
+    object[key] = data[key];
+  }
+}
```
amashutikov, line 32: We could use for this case Object.assign

amashutikov (approved):
Good job!

src/transformState.js
```diff
@@ -6,6 +6,43 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    const type = action.type;
+
+    switch (type) {
```
silchyn:
```suggestion
    switch (action.type) {
```

silchyn (approved):

PR #2318, opened 2023-05-08
add solution

serentinos:


Shevtsooov (approved):
Great job 🔥

src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+    for (const key of dataArray) {
+      delete inputState[key];
+    }
+  };
+
+  const clear = (inputState) => {
+    for (const key in inputState) {
+      delete inputState[key];
+    }
+  };
```
33Orange, line 24: You can move this functions outside function transformState.
Now these functions are created afresh every time the transformState function is called.

33Orange (approved):
Well done!

Just one small nuance with the declaration of functions.

PR #2320, opened 2023-05-08
Solution

amashutikov:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        Object.assign(state, action.extraData);
+        break;
+      case 'clear':
+        clearObject(state);
+        break;
+      case 'removeProperties':
+        removeProperties(state, action.keysToRemove);
+        break;
+      default:
+        return state;
```
serentinos, line 32: The error message about invalid state will be better here I guess:)

serentinos (approved):
Good job!!!

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+      delete object[key];
+    }
+  }
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'clear':
```
silchyn, line 25:
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        
        break;
        
      case 'clear':
```

silchyn (approved):

PR #2322, opened 2023-05-08
solution

artem-ivashchenko:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: It would be great to destructure the action to get all the needed properties of it (type, extraData, ketsToRemove)

polosanya (approved):
Good job!

PR #2323, opened 2023-05-08
Stateful

DenisGurskiy:
All tests works. But I don't know how do this task without a nested loops.


vlad-demchuk (approved):
WDYM `nested loops`? 

nested is when you have a loop inside a loop

PR #2325, opened 2023-05-09
add solution

niaYaro:


cephy77 (approved):
Great!

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
33Orange: Use proper variable names in `for of` loop.

`const action of actions ` would be better

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+    switch (act.type) {
+      case 'addProperties':
+        for (const data in act.extraData) {
+          state[data] = act.extraData[data];
+        };
```
33Orange: Use `Object.assign` here to avoid unnecessary iterations.

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        for (const key of act.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
33Orange: Extra `if` statement.
You can iterate over the entire object and remove all fields without `if` statement here

33Orange (changes requested):
A few points to fix

okosohor (approved):

dinaminahi (approved):

PR #2328, opened 2023-05-09
add functionality to transformState function

cephy77:


IrynaBalandiukh (approved):
Good job!)

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
```
dinaminahi, line 21: The default case should not delete all the state properties. Instead, it should not modify the state if the action type does not match any of the expected cases.

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+          delete state[key];
+        }
```
dinaminahi, line 23: The default case should not delete all the state properties. Instead, try to show some error if the action type does not match any of the expected cases.

dinaminahi (changes requested):

silchyn (approved):

PR #2329, opened 2023-05-09
Add task solution

IrynaBalandiukh:


julshvets (approved):
Great job!

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+  }
+}
+
+function removeProperty(object, properties) {
+  for (const property of properties) {
+    delete object[property];
+  }
+}
+
+function clearObject(object) {
```
33Orange: Consider renaming the function 'clearObject' to 'clearState' to better reflect its purpose.

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+      case 'removeProperties':
+        removeProperty(state, action.keysToRemove);
+        break;
+
+      case 'clear':
+        clearObject(state);
+        break;
+
+      default:
+        return state;
```
33Orange: In the default case statement, you should not return state, instead use 'break' as this can cause the function to terminate early.

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+    }
+  }
+}
+
+function removeProperty(object, properties) {
```
33Orange: Consider renaming the function 'removeProperty' to 'removeProperties' as it is removing multiple properties.

33Orange (changes requested):
Good job!

Some comments:
- Switch/case should always have default case for error handling.
- Use names for functions that correctly describe actions

vladyslav-lapshyn (approved):
GJ 🔥

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+        removeProperties(state, action.keysToRemove);
+        break;
+
+      case 'clear':
+        clearState(state);
+        break;
+
+      default:
+        break;
+    }
```
33Orange: You didn't add error handling in default state.

33Orange (changes requested):
Check my previous comment please.
You still haven't added error handling to the default state. (Checklist p4)

Anna-Tkachenko (approved):
⭐️

PR #2330, opened 2023-05-09
solution

okosohor:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
Yuliya1907: don't use for in loop for iterating over array

Yuliya1907 (approved):
Almost done!Check p.1 of the checklist

src/transformState.js
33Orange: Instead of using 'for...in' loop, use 'for...of' loop to iterate over the 'actions' array.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (actions[action].type === 'addProperties') {
+      Object.assign(state, actions[action].extraData);
+    } else if (actions[action].type === 'removeProperties') {
+      for (const key of actions[action].keysToRemove) {
+        delete state[key];
+      }
+    } else if (actions[action].type === 'clear') {
+      for (const key in state) {
+        if (state.hasOwnProperty(key)) {
```
33Orange: The 'hasOwnProperty' check is unnecessary as you're iterating only over the object's own properties with 'for...in' loop. You can safely remove this check.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+          delete state[key];
+        }
```
33Orange: Extra `if` statement.
You can iterate over the entire object and remove all fields without `if` statement here

33Orange (changes requested):
Good job.

A few moments:
- Don't use `for in` loop for iterating over array.
- Use `switch` statement if you have limited amount of conditions. And remember that switch/case should always have default case for error handling.

panch1811 (approved):
Good job! Nice readability ;)

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties' :
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        } break;
```
dinaminahi: Move the 'break' statement to a new line for better readability

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'clear' :
+        for (const key in state) {
+          delete state[key];
+        } break;
+
+      default:
+        throw new Error('Error');
```
dinaminahi: Replace the generic error message with a more descriptive one, like 'Invalid action type'

dinaminahi (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2331, opened 2023-05-09
Solution

julshvets:


andrusb21 (changes requested):
Check point 4 in checklist: switch/case should always have default case for error handling

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const object of actions) {
```
silchyn:
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+    switch (true) {
+      case object.type === 'addProperties':
```
silchyn: think about how to simplify this `switch-case` structure

silchyn (changes requested):

booobryyyk (approved):
GJ🚀

src/transformState.js
```diff
@@ -5,7 +5,41 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        const { extraData } = action;
```
dinaminahi, line 11: No need  to extract property from object here, it just adds few extra lines to your `switch-case`

dinaminahi (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+
+        removeProperties(state, keysToRemove);
+        break;
+
+      case 'clear':
+        deleteAllProperties(state);
+        break;
+
+      default:
+        return state;
```
silchyn: error could be thrown there

silchyn (approved):

src/transformState.js
julshvets: Thank you! Done.

PR #2332, opened 2023-05-09
Solution

BogdanParubchenk01:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
etojeDenys: Consider using 'switch' statement instead of multiple 'if' statements for better readability and performance.

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+      case 'removeProperties':
+        const deleteData = obj.keysToRemove;
+
+        for (const deleteKey of deleteData) {
+          delete state[deleteKey];
+        }
+        break;
+
+      default:
```
anstsot: it's better to do nothing or throw an error on `default` case
```suggestion
      case 'clear':
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        const arayKeys = Object.keys(state);
+
+        for (const key of arayKeys) {
```
anstsot: it can be siplified by using `for .. in`
```suggestion
        for (const key in state) {
```

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return false;
```
etojeDenys: Instead of returning false, you should continue to the next iteration when an unknown action is encountered.

etojeDenys: or throw the error

etojeDenys (changes requested):

mykhalenych (approved):

PR #2333, opened 2023-05-09
solution

Yuliya1907:


thienlaoo (approved):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
33Orange: Switch/case should always have default case for error handling.

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
```
33Orange: Would be better to use `Object.assign` here without loop.

33Orange (changes requested):
Good job.

A few moments:
- Use `Object.assign` instead loop over Object.entries in 'addProperties' type.
- Switch/case should always have default case for error handling.

33Orange (approved):
Well done! 🔥

PR #2334, opened 2023-05-09
Solution

andrusb21:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const item of action.keysToRemove) {
```
ghost:
```suggestion
        for (const key of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[item];
+        }
+
+        break;
+
+      case 'clear':
+        for (const item in state) {
```
ghost:
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          delete state[item];
+        }
+
+        break;
+
+      default:
+        throw new Error('Error');
```
ghost: it would be better to specify which error we got

ghost (approved):

KatyaZaremska (approved):

src/transformState.js
33Orange: use proper variable names in for of loop

src/transformState.js
33Orange: use proper variable names in for of loop

src/transformState.js
33Orange: It will be better to submit a bug with a minimal description of the bug.

For example:
`'Undefined action type'`

33Orange (changes requested):
Good job!

A few moments:
- Use proper variable names in for of loop.
- Describe error in `default` state

src/transformState.js
andrusb21: ok, it will be better

ivankhrsh (changes requested):
gj :fire:

ivankhrsh (approved):

33Orange (approved):
Well done! 🔥

PR #2335, opened 2023-05-09
solution

thienlaoo:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (key in state) {
+          delete state[key];
+        }
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const q in state) {
```
ghost: q?

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (key in state) {
+          delete state[key];
+        }
+      }
+    }
```
ghost: checklist p2

> [CODE STYLE]: Use switch statement if you have limited amount of conditions.

ghost (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
```
YeMyr: 🔥

YeMyr (approved):

33Orange (changes requested):
- Your `switch` statement don't have default value.
`Switch/case should always have default case for error handling.`

Everything else is well done.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
```
Anna-Tkachenko, line 12: Add empty line before different cases to increase code readability

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (key in state) {
```
Anna-Tkachenko: This check is redundant
If `key` is not in `state` - nothing will happen

Anna-Tkachenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return ('Invalid action type:', action.type);
```
naziamoff: let's use interpolation here instead

naziamoff (approved):
good job 🔥 
Just one small suggestion

PR #2337, opened 2023-05-09
(ym) transformState

YeMyr:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        kindOfAction.keysToRemove.forEach(el => {
+          delete state[el];
+        });
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
ghost: let's think about the readability of our code and add spaces between cases

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const kindOfAction of actions) {
```
ghost: you don't need such complicated name, we can leave it as `action`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    switch (kindOfAction.type) {
+      case 'addProperties':
+        Object.assign(state, kindOfAction.extraData);
+        break;
+      case 'removeProperties':
+        kindOfAction.keysToRemove.forEach(el => {
+          delete state[el];
+        });
```
ghost: why have we iterated here through `forEach`, but our `actions` through `for of`?

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const kindOfAction of actions) {
+    switch (kindOfAction.type) {
+      case 'addProperties':
+        Object.assign(state, kindOfAction.extraData);
+        break;
+      case 'removeProperties':
+        kindOfAction.keysToRemove.forEach(el => {
```
ghost: el?

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[el];
+        });
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
ghost, line 24: let's add default error case (checklist p4)

> [CODE STYLE]: switch/case should always have default case for error handling.

ghost (changes requested):

src/transformState.js
YeMyr: 👌 fixed

src/transformState.js
YeMyr: well, I like it)))
but, if you think, it would be better to change them, there is no problem
Done 👌

src/transformState.js
YeMyr: Why not? 
If there is some restriction about it? 
Why couldn't we use both of them? 🤷‍♂️
👌 fixed

src/transformState.js
YeMyr: el is element of array
Why not? As for me, this is just common and understandable practice for naming 
fixed 👌

src/transformState.js
YeMyr: fixed 👌

Yuliia-kucheriavenko (approved):
Good work*))

33Orange (approved):
Well done ! 🔥

src/transformState.js
ghost: @YeMyr there is no restriction, as for me, it degrades the readability of our code if we can iterate with forEach, it would be better to use this whenever possible, unless it requires additional conversions

src/transformState.js
ghost: @YeMyr it's good idea when we don't have enough time to think about variable names, but in this case `key` would be a better option

PR #2338, opened 2023-05-10
Solution

MykolaLopushenko:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const key of actions) {
+    const { type, extraData, keysToRemove } = key;
+
+    switch (type) {
+      case 'addProperties':
```
OleziO, line 12: Try to move all cases to js object, it is good practice.

OleziO (approved):

PR #2339, opened 2023-05-10
add solution

hekaq:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
ghost: we should throw an error in case our `type` does not match the above (checklist p4)

> [CODE STYLE]: switch/case should always have default case for error handling.

ghost (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const type = action.type;
+
+    switch (type) {
+      case 'addProperties':
```
SofiaLiubchenko, line 12: from my point of view, it is better to write `case something: {do something}` using {}

SofiaLiubchenko (approved):

dimoshen (approved):
clean solution gj

ghost (approved):
🔥

src/transformState.js
Anna-Tkachenko: It can depends on code style in the project, so it's ok to write in both styles ✅

Anna-Tkachenko (approved):
Good job!

PR #2340, opened 2023-05-10
solution

kirilshorin:


maryanmats (approved):

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const extraData = action.extraData;
+    const keysToRemove = action.keysToRemove;
+
+    if (action.type === 'addProperties') {
```
silchyn: `switch` would be better fit there

silchyn (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const extraData = action.extraData;
+    const keysToRemove = action.keysToRemove;
```
dinaminahi: You can use destructuring here: `const { type, extraData, keysToRemove } = action`

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
```
dinaminahi: It is more convenient to use Object.assing or spread operator here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const value of keysToRemove) {
+          delete state[value];
+        }
```
dinaminahi: Let's create separate function for this, to make `switch-case` smaller and more readable

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
dinaminahi: Let's create separate function for this, to make `switch-case` smaller and more readable

dinaminahi (changes requested):
Great job 👍 
Few things to make the readability better

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  function clear() {
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  function removeSomeKeys(keysArray) {
```
naziamoff, line 14: this way it is clear enough
```suggestion
  function removeKeys(keysArray) {
```

naziamoff, line 14: also it is considered a bad practice to mention dataType in the name of the variable - as it gives us very little context.
Here are some better options:

```suggestion
  function removeSomeKeys(keys) {
```

```suggestion
  function removeSomeKeys(keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  function clear() {
```
naziamoff, line 8: also, it is a good practice to move your helper functions to a different file

naziamoff (approved):
🔥

silchyn (approved):

PR #2341, opened 2023-05-10
Solution

dimoshen:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'clear') {
+      for (const stateKey in state) {
```
PanyaPrimal: nested loops

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
```
PanyaPrimal, line 7: use switch for limited amount of condititons

PanyaPrimal (changes requested):
Working solution, but checklist have another ways.

src/transformState.js
dimoshen: fixed. used switch instead of if

src/transformState.js
dimoshen: added func removeKey(action)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
-  // write code here
+  const stateCopy = state;
+  const removeKey = function(actionObject) {
+    for (const keyToRemove of actionObject.keysToRemove) {
+      delete stateCopy[keyToRemove];
+    }
+  };
```
dinaminahi: Let's move this function outside `transformState`, so it will be more readable and it won't be declared each time `transformState` is called.

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'clear':
+        for (const key in stateCopy) {
+          if (key) {
+            delete stateCopy[key];
+          }
+        }
```
dinaminahi: Let's create separate function for this, to make `switch-case` smaller and more readable

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
-  // write code here
+  const stateCopy = state;
+  const removeKey = function(actionObject) {
+    for (const keyToRemove of actionObject.keysToRemove) {
+      delete stateCopy[keyToRemove];
+    }
+  };
+
+  for (const action of actions) {
+    const { type } = action;
```
dinaminahi: You can use destructuring here: const { type, extraData, keysToRemove } = action

dinaminahi (changes requested):
Great job 👍
Few things to make the readability better

src/transformState.js
```diff
@@ -4,8 +4,41 @@
...
+    delete object[keyToRemove];
+  }
+};
+const clearProperties = function(actionObject, object) {
+  for (const key in object) {
+    if (object.hasOwnProperty(key)) {
+      delete object[key];
+    }
+  }
+};
```
naziamoff: I would say arrow function syntax looks better here

```suggestion
const removeKey = (actionObject, object) => {
  const { keysToRemove } = actionObject;

  for (const keyToRemove of keysToRemove) {
    delete object[keyToRemove];
  }
};
const clearProperties = (actionObject, object) => {
  for (const key in object) {
    if (object.hasOwnProperty(key)) {
      delete object[key];
    }
  }
};
```

src/transformState.js
```diff
@@ -4,8 +4,41 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+const removeKey = function(actionObject, object) {
```
naziamoff: also, it is a bad practice to add the dataType to the variable's name. It basically does not give us any information, just takes space. 

I would stick to just `actions`

naziamoff: also, `object` is not the best way either. Think of a better way - your names should be maximally clear and descriptive

src/transformState.js
```diff
@@ -4,8 +4,41 @@
...
+  for (const keyToRemove of keysToRemove) {
+    delete object[keyToRemove];
+  }
+};
+const clearProperties = function(actionObject, object) {
+  for (const key in object) {
+    if (object.hasOwnProperty(key)) {
+      delete object[key];
+    }
+  }
```
naziamoff: you are already iterating through the keys of the object, do you really need to check whether they have the property?

Also, you can just reset the object itself instead of removing properties one by one

naziamoff (changes requested):
not bad, but a couple of fixes to go 🚀

dinaminahi (changes requested):
Please, check @naziamoff comments

src/transformState.js
dimoshen: changed to arrow funcs

src/transformState.js
dimoshen: they are just 'data' now

src/transformState.js
dimoshen: > Also, you can just reset the object itself instead of removing properties one by one

do you mean stateCopy = {} ? but I'll lose the link to the state object, cause i should make changes in the state. also, i can't make assignments to a function parameter. am I wrong?

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+};
+const clearProperties = (action, data) => {
+  for (const key in data) {
+    delete data[key];
+  }
+};
+
 function transformState(state, actions) {
-  // write code here
+  const stateCopy = state;
```
33Orange: Creating a stateCopy is unnecessary, as the original state should be modified directly. Remove this line and replace all occurrences of stateCopy with state.
You are not making a copy. You just created another link to the same object

src/transformState.js
```diff
@@ -4,8 +4,39 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+const removeKey = (action, data) => {
+  const { keysToRemove } = action;
```
33Orange: Why do you need the whole action object inside the function when you only need the `keysToRemove` array ?

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+      case 'addProperties':
+        Object.assign(stateCopy, extraData);
+        break;
+      case 'removeProperties':
+        removeKey(action, stateCopy);
+        break;
+    }
+  }
+
+  return stateCopy;
```
33Orange: You don't need to return `stateCopy` here.
The function mutates the object passed to it via parameters and returns nothing.

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+      case 'clear':
+        clearProperties(action, stateCopy);
+        break;
+      case 'addProperties':
+        Object.assign(stateCopy, extraData);
+        break;
+      case 'removeProperties':
+        removeKey(action, stateCopy);
+        break;
+    }
```
33Orange, line 23: You forgot to make a `default` state with an error when `action.type` is unknown.
Try to use `throw new Error()` and describe error with `action.type`.
`[CODE STYLE]: switch/case should always have default case for error handling.`

33Orange (changes requested):
Almost done!

A few moments to fix.

src/transformState.js
dimoshen: removed

src/transformState.js
dimoshen: fixed. working with state directly now

src/transformState.js
dimoshen: using ' const { type, extraData, keysToRemove } = action;' destructuring and using keysToRemove in the loop

src/transformState.js
dimoshen: yes. added now

src/transformState.js
```diff
@@ -4,8 +4,35 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+const removeKey = (keysToRemove, data) => {
+  for (const keyToRemove of keysToRemove) {
+    delete data[keyToRemove];
+  }
+};
+const clearProperties = (action, data) => {
```
dinaminahi: Redundant argument  `action`, you`re not using it

dinaminahi (changes requested):
Almost there, one more thing to do

src/transformState.js
dimoshen: I'll put it on QnA

dimoshen: fixed now

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+      case 'clear':
+        clearProperties(state);
+        break;
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+      case 'removeProperties':
+        removeKey(keysToRemove, state);
+        break;
+      default:
```
naziamoff, line 21: would be nice to add some empty lines between different cases for improving readability
```suggestion
    switch (type) {
      case 'clear':
        clearProperties(state);
        break;
        
      case 'addProperties':
        Object.assign(state, extraData);
        break;
        
      case 'removeProperties':
        removeKey(keysToRemove, state);
        break;
        
      default:
```

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        throw new Error(`Invalid action type: ${action.type}`);
+    }
+  }
 }
 
+const removeKey = (keysToRemove, data) => {
```
naziamoff, line 27:
```suggestion
const removeKeys = (keysToRemove, data) => {
```

naziamoff (approved):
Well done 👍

PanyaPrimal (approved):

PR #2342, opened 2023-05-10
transformState solution

panch1811:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
bohdan-mykhailenko: Using 'for of' instead of 'for ...' is better in that case because we needn't indexes only values. And it would make code shorter and more readable.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          if (action.keysToRemove.includes(keyToRemove)) {
+            delete state[keyToRemove];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
bohdan-mykhailenko: Use empty line between case blocks.

bohdan-mykhailenko (approved):
good job 👍

src/transformState.js
dinaminahi: As @bohdan-mykhailenko said consider using 'for...of' loop instead of 'for' loop for better readability

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        for (const keyToRemove in state) {
```
dinaminahi: Instead of looping through all keys in the state, it's more efficient to loop through 'keysToRemove' and check if the key exists in the state.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
```
dinaminahi: Instead of loop, you can use Object.assign here

dinaminahi (changes requested):

silchyn (approved):

PR #2343, opened 2023-05-10
Solution

k1samehoshigaki:


growingananas (commented):
Well done!

vchervinko (approved):

src/transformState.js
```diff
@@ -4,8 +4,53 @@
...
+//         delete state[key1];
+//       }
+//     }
+//   }
+
+//   return state;
+// }
 function transformState(state, actions) {
-  // write code here
+  for (const index of actions) {
```
silchyn: which index is it?

src/transformState.js
```diff
@@ -4,8 +4,53 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key1 in state) {
+          delete state[key1];
+        }
+        break;
+      default:
+        break;
```
silchyn: try throwing an error with meaningful text there

src/transformState.js
```diff
@@ -4,8 +4,53 @@
...
+      case 'addProperties':
+        for (const key in index.extraData) {
+          state[key] = index.extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        for (const property of index.keysToRemove) {
+          if (state[property]) {
+            delete state[property];
+          }
```
silchyn: could be separate function

silchyn (changes requested):

src/transformState.js
k1samehoshigaki: Ok, I will.

src/transformState.js
k1samehoshigaki: I hope you meant this...

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
dinaminahi: It`s more convenient to use Object.assign or spread operator here

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        break;
+      case 'removeProperties':
+        removeProperties(action, state);
```
dinaminahi, line 15: Good for you 👍  Creating a separate function made it more readable

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        break;
+      case 'clear':
+        for (const key1 in state) {
+          delete state[key1];
+        }
```
dinaminahi: Do it also here

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        break;
+      default:
+        throw new Error(`Unknown action type ${action.type}.`);
+    }
+  }
+
+  return state;
+}
+
+function removeProperties(a, b) {
```
dinaminahi: a and b tells nothing

Great naming makes code more understandable, we should look at the names and know what this is about and what it is going to do 

So  something like 'object' and 'keys to remove' would be better here

dinaminahi (changes requested):

src/transformState.js
k1samehoshigaki: You are absolutely right! Thx!

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        removeProperties(action, state);
```
naziamoff: Would be nice to add some empty lines between different cases to make your switch more readable
```suggestion
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
      
      case 'removeProperties':
        removeProperties(action, state);
```

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+      default:
+        throw new Error(`Unknown action type ${action.type}.`);
+    }
+  }
+
+  return state;
+}
+
+function removeProperties(object, keysToRemove) {
+  for (const property of object.keysToRemove) {
```
naziamoff: usually we call this variable this way

```suggestion
  for (const key of object.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+    delete keysToRemove[property];
+  }
+}
+
+function clear(object) {
+  for (const key1 in object) {
```
naziamoff: `key1` - 🙅‍♀️ 🙅‍♀️ 🙅‍♀️ 
Never use numbers in your variable names, even if you can't resist the temptation. 
You want your names to be maximally clear and descriptive

naziamoff: also you can just reset the object here instead of removing properties one by one

naziamoff (approved):
good job 👍 
A couple of adjustments to go

naziamoff (changes requested):
oops, I meant to reject

33Orange (approved):
Well done ! 🔥

naziamoff (approved):

silchyn (approved):

PR #2344, opened 2023-05-10
add solution

ulositska:


src/transformState.js
```diff
@@ -3,9 +3,35 @@
...
+          delete stateModified[key];
+        }
+        break;
+
+      case 'default':
+        throw new Error(`No ${action.type} action type found`);
+    }
+  }
+
+  return stateModified;
```
33Orange: The function should not return anything. Remove the return statement.

src/transformState.js
```diff
@@ -3,9 +3,35 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in stateModified) {
+          delete stateModified[key];
+        }
+        break;
+
+      case 'default':
```
33Orange: Replace 'case' with 'default' for the error handling block in the switch statement.

[A `default` clause -  if provided, this clause is executed if the value of expression doesn't match any of the case clauses. A switch statement can only have one default clause.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)

src/transformState.js
```diff
@@ -3,9 +3,35 @@
 /**
  * @param {Object} state
  * @param {Object[]} actions
- */
+*/
+
 function transformState(state, actions) {
-  // write code here
+  const stateModified = state;
```
33Orange: Why do we need a new object reference?
Do not use a separate variable `stateModified`. Modify the `state` directly.

src/transformState.js
```diff
@@ -3,9 +3,35 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(stateModified, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete stateModified[key];
+        }
```
33Orange: Let's move this code into a separate function outside of our

src/transformState.js
```diff
@@ -3,9 +3,35 @@
...
+        break;
+
+      case 'clear':
+        for (const key in stateModified) {
+          delete stateModified[key];
+        }
```
33Orange: Let's move this code into a separate function outside of our

33Orange (changes requested):
Good job.

A few moments:
- Do not use a separate variable `stateModified`. Modify the `state` directly.
- Let's move code from `removeProperties`  and `clear` to separated functions outside our `transformState` function.
- Change `case 'default'` to default.
- The function should not return anything. Remove the return statement.

src/transformState.js
ulositska: Sorry, forgot the syntax of the switch statement🤦

src/transformState.js
ulositska: For now kept it inside the main function, but outside the for loop. I'm not reusing it somewhere else, so it seems fine. But tell me if it's a bad decision:pray:

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
+      case 'clear':
+        removeAllKeys();
+        break;
+
+      default:
+        throw new Error(`No ${action.type} action type found`);
+    }
+  }
+
+  function removeAllKeys() {
```
33Orange: Add the 'state' parameter to the 'removeAllKeys' function, so it can access and modify the state object.

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  function removeSpecificKeys(arrayToClean) {
```
33Orange: Change the 'removeSpecificKeys' function to accept the state object and keys to remove as parameters: 'function removeSpecificKeys(state, keysToRemove)'.

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
+    for (const key of arrayToClean.keysToRemove) {
```
33Orange: Replace 'arrayToClean.keysToRemove' with 'keysToRemove' to use the parameter directly.

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        removeSpecificKeys(action);
+        break;
+
+      case 'clear':
+        removeAllKeys();
```
33Orange: Pass the state object as an argument to the 'removeAllKeys' function, like this: 'removeAllKeys(state)'.

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        removeSpecificKeys(action);
```
33Orange: Pass the state object as an argument to the 'removeSpecificKeys' function, like this: 'removeSpecificKeys(state, action.keysToRemove)'.

src/transformState.js
```diff
@@ -3,9 +3,39 @@
...
+    for (const key in state) {
+      delete state[key];
+    }
+  }
+
+  function removeSpecificKeys(arrayToClean) {
+    for (const key of arrayToClean.keysToRemove) {
+      delete state[key];
+    }
+  }
```
33Orange: Would be better to move this functions outside the `transformState` function. Since these functions are now created every time the `transformState` function is called

33Orange (changes requested):
Good job.

A few moments with new functions. Everything else is well done.
- Would be better to move functions outside the `transformState` function. Since these functions are now created every time the `transformState` function is called. And change this functions to accept the parameters with state object and keys(if needed)

dinaminahi (approved):
Well done 💪

PR #2345, opened 2023-05-10
add transformState solution

booobryyyk:
add transformState solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -8,15 +8,13 @@ function transformState(state, actions) {
   for (const action of actions) {
     switch (action.type) {
       case 'addProperties':
-        for (const data in action.extraData) {
-          state[data] = action.extraData[data];
-        }
+        Object.assign(state, action.extraData);
+
         break;
 
       case 'clear':
-        for (const key in state) {
-          delete state[key];
-        }
+        clearObject(state);
+
         break;
 
       case 'removeProperties':
@@ -28,6 +26,12 @@ function transformState(state, actions) {
       default:
         throw new Error(`There is no such action type: ${action.type}`);
     }
+  }
+}
+
+function clearObject(obj) {
+  for (const key in obj) {
+    delete obj[key];
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+      case 'addProperties':
+        for (const data in action.extraData) {
+          state[data] = action.extraData[data];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
dinaminahi: Let's create a separate function for this to make `switch-case` smaller and more readable

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const data in action.extraData) {
+          state[data] = action.extraData[data];
+        }
```
dinaminahi: It`s better to use Object.assign or spread operator here

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error(`There is no such action type: ${action.type}`);
```
dinaminahi, line 26: I like that you throw an error with a clear description 👍

dinaminahi (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+      case 'clear':
+        clearObject(state);
+
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
```
33Orange: Let's put this in a separate function, just like `clearObject`

33Orange (changes requested):
Well done! 🔥 

Just one moment with readability

Anna-Tkachenko (approved):
⭐️

PR #2346, opened 2023-05-10
Solution

vchervinko:


src/transformState.js
```diff
@@ -5,7 +5,22 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
dinaminahi: Let's use `switch-case` here according to checklist p.2

dinaminahi (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        }
+        break;
+
+      case 'removeProperties':
+        for (const propertiesToRemove of action.keysToRemove) {
+          delete state[propertiesToRemove];
+        }
+        break;
 
+      case 'addProperties':
```
Anna-Tkachenko, line 22: Check point 4 in checklist

Anna-Tkachenko (changes requested):

dinaminahi (approved):

PR #2347, opened 2023-05-10
Solution

maryanmats:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    if (type === 'addProperties') {
```
dinaminahi: Let's use switch-case here according to checklist p.2

src/transformState.js
```diff
@@ -6,6 +6,29 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    const { type } = action;
```
dinaminahi, line 9: I like that you decided to extract data from action object 👍 
But we can extract here all the data we will need in one line.

dinaminahi (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,38 @@
...
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = action;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
dinaminahi, line 25: We can make a separate function for this, so our `switch-case` will be smaller and more readable

src/transformState.js
dinaminahi: I meant that instead of 3 lines of code, you could do it in one: 
`
const { type } = action;
const { extraData } = action;
const { keysToRemove } = action;
`

`
`

dinaminahi (approved):
Well done 👍 
But check comments how you could improve it

PR #2350, opened 2023-05-10
Solution

SofiaLiubchenko:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let changedState = state;
```
dinaminahi: There is no need of creating one more variable. This will have same reference as `state`, and when we change `changedState`, `state` will change the same. So we can work directly with `state`,

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+    switch (action.type) {
+      case 'addProperties': {
+        changedState = Object.assign(changedState, action.extraData);
+        break;
+      }
+
+      case 'removeProperties': {
+        for (const key of action.keysToRemove) {
+          delete changedState[key];
+        }
```
dinaminahi: Think of creating separate function for this, so that `switch-case` can be smaller and more readable

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+
+      case 'clear': {
+        for (const property in changedState) {
+          delete changedState[property];
+        }
+        break;
+      }
+
+      default:
+        throw new Error('Unknown action');
```
dinaminahi: Not critical, but can we add also a type of action in error, so it will be more descriptive?

dinaminahi (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #2357, opened 2023-05-10
Latest solution

PanyaPrimal:


src/transformState.js
```diff
@@ -4,8 +4,44 @@
...
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties': {
+        const { extraData } = action;
+
+        for (const [key, value] of Object.entries(extraData)) {
+          state[key] = value;
+        }
```
dinaminahi: Better here would be to use Object.assign or spread operator, which is easier and more readable

src/transformState.js
```diff
@@ -4,8 +4,44 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
dinaminahi: Let`s extract all data we need at once

src/transformState.js
```diff
@@ -4,8 +4,44 @@
...
+      }
+
+      case 'removeProperties': {
+        const { keysToRemove } = action;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
dinaminahi: it's better to create separate function for this, so our `switch-case` would e smaller and more readable

src/transformState.js
```diff
@@ -4,8 +4,44 @@
...
+        break;
+      }
+
+      case 'clear': {
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
```
dinaminahi: same here

dinaminahi (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,46 @@
...
+        delete obj[key];
+      }
+    }
+  }
+
+  function clearProperties(obj) {
+    for (const key of Object.keys(obj)) {
+      delete obj[key];
+    }
+  }
```
naziamoff, line 46: I guess you can just reset the object instead of removing properties one by one

naziamoff (approved):
great job 🔥

PR #2361, opened 2023-05-10
code

liza-strykharchuk:
ignore my comments in the code, i need  them in the future exercise

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,44 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
 
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,44 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
```
polosanya, line 21:
```suggestion
        for (const key in state) {
```
You can simplify it

src/transformState.js
```diff
@@ -5,7 +5,44 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can also destructure the action and get all the needed props of it (type, extraData, keysToRemove)

polosanya (approved):
Good job! I'm approving it, but please pay attention to the comment

PR #2362, opened 2023-05-10
add solution of stateful object

vladyslav-lapshyn:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key in state) {
```
Anna-Tkachenko: Let's go through the keys that we need to remove here, instead of go through all state keys

Anna-Tkachenko (changes requested):

src/transformState.js
vladyslav-lapshyn: @Anna-Tkachenko thanks, cnanged that code

dinaminahi (approved):

PR #2365, opened 2023-05-10
add solution

vitvadym:


silchyn (approved):

silchyn (changes requested):
checklist rule `#4`

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const currentState = state;
```
33Orange: It is unnecessary to create a new variable 'currentState'. You can directly modify 'state' without creating a new reference.

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+        }
+
+        break;
+
+      default:
+        throw Error(`Wrong action: ${action.type}`);
+    }
+  }
+
+  return state;
```
33Orange: Returning 'state' is not necessary, as the function is supposed to modify the original object. Remove the 'return' statement.

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+  const currentState = state;
+
+  for (const action of actions) {
+    const { type, keysToRemove, extraData } = action;
+
+    switch (type) {
+      case 'clear':
+        for (const key in currentState) {
+          delete currentState[key];
+        };
```
33Orange: Let's move this code into a separate function after `transformState`

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+
+      case 'addProperties':
+        Object.assign(currentState, extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const keyToRemove of keysToRemove) {
+          delete currentState[keyToRemove];
+        }
```
33Orange: Let's move this code into a separate function after `transformState`

33Orange (changes requested):
Good job.

A few moments to fix

naziamoff (approved):
great job 👍 
If we create a helper function - we usually move it to a separate file. You can practice this if you want

silchyn (approved):

PR #2367, opened 2023-05-11
Solution

EvgenBilenkiy:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
```
OleziO, line 8: Remove all comments from the code.

OleziO (approved):

PR #2368, opened 2023-05-11
Solution

RoksolanaTylipska:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          delete state[keyToRemoveInState];
+        }
+        break;
+
+      default:
+        return `Unknown type ${type} `;
+    }
+  }
+
+  return state;
```
33Orange: No need to return the state object at the end of the function. The function should modify the state object directly without returning a value.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const keyToRemoveInState in state) {
+          delete state[keyToRemoveInState];
+        }
+        break;
+
+      default:
+        return `Unknown type ${type} `;
```
33Orange: Use `throw new Error()` to error handling

33Orange (changes requested):
Well done ! ⭐ 

A few moments:
- Use `throw new Error()` to error handling
-No need to return the state object at the end of the function. The function should modify the state object directly without returning a value.

silchyn (approved):

PR #2369, opened 2023-05-11
Solution

Sergej161995:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    if (type === 'addProperties') {
```
TarasHoliuk: GJ, but use `switch` instead of `if else` blocks

TarasHoliuk (changes requested):

TarasHoliuk (approved):
GJ :+1: 
Just FYI: it's a good practice to always add a default case in the end of switch block, even empty, like this:
```
default:
  break;
```

PR #2370, opened 2023-05-11
Solution

semchuk-rodion:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Viktor-Kost: Consider using a 'for of' loop instead of a 'for' loop for better readability (const action of actions)

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
+
+  return state;
```
Viktor-Kost: The function should not return the state as it is modified by reference. Remove the return statement.

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+
+    if (action.type === 'addProperties') {
```
Viktor-Kost: it's better to use switch statement instead of if/else

Viktor-Kost (changes requested):

etojeDenys (approved):
lgtm

PR #2371, opened 2023-05-11
Solution

olehmahera:


silchyn (changes requested):
What about the default case?

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-}
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const keys of action.keysToRemove) {
```
33Orange: Consider renaming 'keys' to 'key' in the for loop to better represent that it's a single key.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[keys];
+        }
```
33Orange: Let's move this code into a separate function after `transformState`.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+        break;
 
+      case 'clear':
+        for (const keys in state) {
+          delete state[keys];
+        }
```
33Orange: Let's move this code into a separate function after `transformState`.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+        break;
+
+      default:
+        return 'Unknown command';
```
33Orange: `switch/case should always have default case for error handling.` (Checklist p4)

Let's try `throw new Error()`. In which we describe the error and the action.type that fell into default

33Orange (changes requested):
Good job!

A few moments: 
- Let's move actions in `removeProperties` and 'clear' in a separate function after `transformState`.
- Refactor default case with error handling.

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          deleteParam(key);
+        }
+
+        break;
+
+      default:
+        throw new Error('Unknown command');
```
33Orange: Add more descriptive message to Error.
`${action.type} is not defined !` for example

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  function deleteParam(params) {
```
33Orange: Consider moving the 'deleteParam' function outside of 'transformState' to improve readability and prevent unnecessary re-declarations.

33Orange (changes requested):
1. Create 2 functions outside `transformState`. 1 function for `removeProperties` and 1 function for `clear`.
2. Change Error msg to something more descriptive. Try to add action type with template string.

src/transformState.js
```diff
@@ -4,8 +4,36 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          deleteParam(state, key);
+        }
```
33Orange: The idea was not to move the delete `object[item]` line, but to move all the logic of these actions into specific functions. To make it easier to read the code.

```
function deleteParam(state, params) {
  for (const key of params) {
    delete state[params];
  }
}
```

src/transformState.js
```diff
@@ -4,8 +4,36 @@
...
+
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          deleteParam(state, key);
+        }
```
33Orange: The same situation as with the remove function

src/transformState.js
```diff
@@ -4,8 +4,36 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+
+function deleteParam(state, params) {
+  delete state[params];
 }
```
33Orange: Declare these functions under the `transformState` function

33Orange (changes requested):
Almost done.

A few moment to fix.

src/transformState.js
```diff
@@ -4,8 +4,41 @@
...
+  }
+}
+
+function removeProperties(object, params) {
+  for (const property of params) {
+    delete object[property];
+  }
+}
+
+function clear(object) {
```
dinaminahi, line 38: Please use arrow function instead of regular function for consistency with other functions in the code.

src/transformState.js
```diff
@@ -4,8 +4,41 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
```
dinaminahi, line 8: Please use arrow function instead of regular function for consistency with other functions in the code.

src/transformState.js
```diff
@@ -4,8 +4,41 @@
...
+
+        break;
+
+      default:
+        throw new Error(`${action.type} is not defined !`);
+    }
+  }
+}
+
+function removeProperties(object, params) {
```
dinaminahi, line 32: Please use arrow function instead of regular function for consistency with other functions in the code.

dinaminahi, line 32: The function name is `removeProperties` so better name for second argument would be `propertiesToRemove` or `keysToRemove`

dinaminahi (approved):

silchyn (approved):

PR #2375, opened 2023-05-12
Solution

reseneweb:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const meaning in actions) {
+    const point = actions[meaning];
+
+    for (const keys in point) {
```
witflash: Instead of using an additional cycle, use type checking:
```
if (action.type === 'addProperties') {
 ...
}
```

As I remember, I checked your prev similar work. Try to improve this approach a bit.

witflash (changes requested):
Check my comment and improve your code a little bit.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    const point = actions[action];
+
+    if (point.type === 'addProperties') {
```
agniya-i, line 11:
```suggestion
  for (const action of actions) {
    if (action.type === 'addProperties') {
```

you can also use `for...of` to make your code simpler and to not create extra variable

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    }
+
+    if (point.type === 'removeProperties') {
+      for (let i = 0; i < point.keysToRemove.length; i++) {
+        delete state[point.keysToRemove[i]];
+      }
+    }
+
+    if (point.type === 'clear') {
+      for (const n in state) {
```
agniya-i, line 22:
```suggestion
      for (const key in state) {
```

don't forget to use more descriptive variables names

agniya-i (approved):

PR #2376, opened 2023-05-12
add task solution

sasdaiv:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+    }
```
witflash, line 29: Switch/case should have a default block.
You can handle some errors in other tasks for example. Here you can just do "break" inside "default"

witflash (approved):
Good job!
Take attention to my comment, plz

PR #2377, opened 2023-05-13
State can be transform

1v43k:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const prop of act.keysToRemove) {
+          delete state[prop];
+        }
+        break;
+
+      case 'clear':
+        for (const keys in state) {
```
Anastasiia-Svintsova: Use a more descriptive variable name instead of 'keys'. For example, 'key'.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
+    switch (act.type) {
```
Anastasiia-Svintsova: Consider using 'action' instead of 'act' to make the code more readable.

Anastasiia-Svintsova (changes requested):
Good job! Let's add some improvements

etojeDenys (approved):
lgtm

PR #2378, opened 2023-05-14
Solution

Svyrydov-Alex:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
anstsot: rewrite the code to use `switch/case`. do not forget about `default` case
image

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 10: Try to move all cases to the object

OleziO (approved):

PR #2379, opened 2023-05-15
Solution

Ihor-Zakharov:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
```suggestion
            delete state[key];
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        for (const [key, value] of Object.entries(action.extraData)) {
+          state[key] = value;
+        }
```
```suggestion
       Object.assign(state, action.extraData)
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
```suggestion
    const { type, extraData, keysToRemove } = action;
```

SanyaBratashchuk (changes requested):

tetiana-ivanchenko (changes requested):
fix all prev comments, also check all points from checklist (default case in the switch)

SanyaBratashchuk (approved):

PR #2381, opened 2023-05-17
Solution

kropiva:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
```
OleziO, line 14: Try to create object with cases texts.

OleziO (approved):

PR #2383, opened 2023-05-18
add task solution

w315:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      }
+    } else if (actions[i].type === 'removeProperties') {
+      for (const removingProp in actions[i].keysToRemove) {
+        delete state[actions[i].keysToRemove[removingProp]];
+      }
+    } else if (actions[i].type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
nazarmatsevych: Rewrite it using switch/case since we rely on a single value

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
```suggestion
  for (const action of actions) {
```

nazarmatsevych (changes requested):

mykhalenych (approved):

PR #2384, opened 2023-05-18
SolutionTransformState

anastasiia-burkovska:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
Viktor-Kost: it's better to name like this
```suggestion
  for (const action of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    for (const key in object) {
+      if (key === 'extraData') {
```
Viktor-Kost: You can apply `switch` statement by `action.type` instead of nested `loop` and `if` statements

Viktor-Kost (changes requested):

mykhalenych (approved):

PR #2385, opened 2023-05-18
Solution

mkorulczyk:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  const add = 'addProperties';
+  const remove = 'removeProperties';
+  const clearProp = 'clear';
+
+  for (let i = 0; i < actions.length; i++) {
```
Radoslaw-Czerniawski, line 12: You can just do
```js
actions.forEach(action => {
```
instead of 
```js
for (let i = 0; i < actions.length; i++) {
  const action = actions[i]
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+    const action = actions[i];
+
+    if (action.type === add) {
```
Radoslaw-Czerniawski, line 15: This here is a perfect case for a [switch statement](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) rather than a bunch of ifs

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  const add = 'addProperties';
```
Radoslaw-Czerniawski, line 8: There is no point in declaring variables with these types, just use them explicitly, an example with swtich statement
```js
switch (action.type) {
  case 'addProperties':
```

Radoslaw-Czerniawski (changes requested):
Needs some slight refactoring :)

PR #2386, opened 2023-05-19
Solution

IraPolchanin:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
mykhalenych, line 20: The default case should not clear the entire state object. Instead, it should be removed or handled with a proper case if needed.

mykhalenych (approved):

PR #2389, opened 2023-05-19
Solution

Lidia-Pintos-Medina:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Anastasiia-Svintsova: Follow the checklist, please. Use switch/case 
image

Anastasiia-Svintsova (changes requested):

src/transformState.js
Lidia-Pintos-Medina: You're right, sorry, I forgot to check it. I've just made the changes and seems to work too.

src/transformState.js
```diff
@@ -4,8 +4,33 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
```
anstsot, line 12: do not forget to add `default` case

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+    }
+  }
```
anstsot, line 36: please return the value at the end of function

anstsot (changes requested):

polosanya (approved):
Good job!

PR #2391, opened 2023-05-20
Solution

hulchak:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      Object.assign(state, actions[action].extraData);
+    }
+
+    if (actions[action]['type'] === 'removeProperties') {
+      for (const del of actions[action].keysToRemove) {
+        delete state[del];
+      }
+    }
+
+    if (actions[action]['type'] === 'clear') {
```
TarasHoliuk: Access the 'type' property using dot notation (actions[action].type) instead of bracket notation

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    if (actions[action]['type'] === 'addProperties') {
+      Object.assign(state, actions[action].extraData);
+    }
+
+    if (actions[action]['type'] === 'removeProperties') {
```
TarasHoliuk: Access the 'type' property using dot notation (actions[action].type) instead of bracket notation

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
+    if (actions[action]['type'] === 'addProperties') {
```
TarasHoliuk: Access the 'type' property using dot notation (actions[action].type) instead of bracket notation

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action in actions) {
```
TarasHoliuk: Use 'for...of' loop instead of 'for...in' loop to iterate over the 'actions' array

src/transformState.js
TarasHoliuk: use `switch case` block here instead of multiple `if` blocks

TarasHoliuk (changes requested):

src/transformState.js
sTorba24: still not resolved

sTorba24 (changes requested):
Fix previous comment

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const del of action.keysToRemove) {
```
witflash, line 14: It's better to name it: `for (cons key of action.keysToRemove)`. Because you iterate literally the keys.

witflash (approved):
Well done 👍

PR #2393, opened 2023-05-22
Solution

andrii-rykal:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const data in action.extraData) {
+        state[data] = action.extraData[data];
+      }
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (state[key]) {
```
mykhalenych, line 17: Replace the if statement with 'if (key in state)'. The current condition will not delete the property if its value is falsy.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
mykhalenych, line 8: You can destruct your action
`const {type, extraData, keysToRemove} = action`

mykhalenych (approved):

PR #2394, opened 2023-05-22
Add state transformation

AmadeuszAndroid:


DorotaLeniecDev (changes requested):
I think using recurency here complicates the code, please refactor without recurency.

Issues:
function name is `addFirstProperty` and as parameters you are passing all properties so it is a bit confusing.

This can be simply done using loops and Object.assign

AmadeuszAndroid:
I changed recurency to iteration. I ignored in this purpose [prohibition nested loops](https://github.com/mate-academy/layout_product-cards/blob/master/checklist.md?plain=1#L97).

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+      case 'clear':
+        for (const stateKey of Object.keys(state)) {
+          delete state[stateKey];
+        }
+        break;
+      default:
+        throw new Error('Bad state.');
+    }
+  }
```
DorotaLeniecDev: For readability add spaces:
And I think this error message would be more suitable
```suggestion
  for (const action of actions) {
    switch (action.type) {
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const keyToRemove of action.keysToRemove) {
          delete state[keyToRemove];
        }
        break;
        
      case 'clear':
        for (const stateKey of Object.keys(state)) {
          delete state[stateKey];
        }
        break;
        
      default:
        throw new Error('Incorrect action type');
    }
  }
```

DorotaLeniecDev (changes requested):
One more for perfection :)

src/transformState.js
AmadeuszAndroid: I added new lines between cases and more descriptive error message.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const stateKey of Object.keys(state)) {
+          delete state[stateKey];
+        }
+        break;
+
+      default:
+        throw new Error('Incorrect action type.');
```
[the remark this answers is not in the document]
DorotaLeniecDev: ignore this comment, I clicked AI review by mistake

DorotaLeniecDev (approved):
Great :)

PR #2395, opened 2023-05-22
transformState

Dorosh90:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
sTorba24: use switch case instead

sTorba24 (changes requested):
GJ! Lets improve it

sTorba24 (approved):
Nice!

PR #2396, opened 2023-05-22
Solution

superzuc:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    if (action.type === 'addProperties' && action.extraData) {
```
anstsot, line 9: use `switch/case`

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
```
anstsot, line 8: better to use `for...of`

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      });
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  });
```
anstsot, line 22: return the value at the end of function

anstsot (changes requested):
Please review the [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) and fix accrodingly

PR #2397, opened 2023-05-23
Solution

SerhiiVoitiuk:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
```
anstsot, line 11: please add `default` case for `switch/case`

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
```
anstsot, line 37: return some value at the end of function

anstsot (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,38 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
```
OleziO, line 12: Try to move all cases into the object fields.

OleziO (approved):

PR #2398, opened 2023-05-24
Solution

alina-isakova:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: instead of using many `if` blocks use `switch`
also, you may destruct all properties you need from action object before `switch` in loop

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
```
Volodymyr-Melnyk-Fron10der, line 30: add `default ` case and throw the error in action type is unknown

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
`for (const { type, ...so on } of actions) {`

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
```
Volodymyr-Melnyk-Fron10der: check is key in the state, before deleting it

Volodymyr-Melnyk-Fron10der (changes requested):

demetrxx (changes requested):
Instead of returning a string in default case of transformState function, throw an error, as mentioned by Volodymyr. Read more and see examples [here ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw)

demetrxx (approved):
Well done 👍 Go ahead!

alina-isakova:
> Well done 👍 Go ahead!

Thank you!

PR #2399, opened 2023-05-24
added solution

vladyslav10shevchenko:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case removeAction:
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      case clearAction:
+        for (const key in state) {
```
Viktor-Kost, line 21: You can remove the 'if (state.hasOwnProperty(key))' check, as the 'for...in' loop already iterates only over the object's own properties.

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const addAction = 'addProperties';
+  const removeAction = 'removeProperties';
+  const clearAction = 'clear';
```
Viktor-Kost: these variables are redundant, you can just use values at your cases directly

Viktor-Kost (changes requested):
GJ!
Some minor fixes needed

Viktor-Kost (approved):

PR #2400, opened 2023-05-25
commit task transformState

IllyaMyhalyk:


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+          break;
+  
+        case 'clear':
+          for (const key in state) {
+            delete state[key];
+          }
+          break;
+  
+        default:
+          return 0;
```
etojeDenys: let's throw the error in default

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
anstsot, line 9: please add the `default` case for `switch/case` construction. you can `throw Error` in that case

anstsot (changes requested):

src/transformState.js
IllyaMyhalyk: ))misunderstood the previous comment))

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return {};
```
loralevitska, line 26:
```suggestion
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
      default:
        return {};
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```

loralevitska (approved):
Well done)

PR #2401, opened 2023-05-28
Solution

ArtemBehter:


DorotaLeniecDev (changes requested):
Good job, add default case to switch statement

mkyloa (approved):

PR #2402, opened 2023-05-29
decision

mukutiuk:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: use `switch` block instead of multiple `if` blocks

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    }
+
+    if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach(element => {
+        delete state[element];
+      });
+    }
+
+    if (action.type === 'clear') {
+      for (const stat in state) {
```
TarasHoliuk:
```suggestion
      for (const key in state) {
```
inside `for in` loop you have a  key of each property, so it better name it `key` explictly

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[element];
+        });
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
TarasHoliuk, line 25:
```suggestion
        break;
        
      default:
        break;
    }
```
it's a good practice to always add `default` case, even empty (FYI)

TarasHoliuk (approved):

PR #2404, opened 2023-05-30
add solution

prostonadobobb:
// eslint-disable-next-line no-console - use for debug 'console.log()'

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,22 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  for (const act of actions) {
-    if (act.type === 'addProperties') {
-      for (const addItem in act.extraData) {
-        state[addItem] = act.extraData[addItem];
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties': {
+        for (const addItem in action.extraData) {
+          state[addItem] = action.extraData[addItem];
+        }
+        break;
       }
-    }
 
-    if (act.type === 'removeProperties') {
-      for (const removeItem of act.keysToRemove) {
-        delete state[removeItem];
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+    if (act.type === 'addProperties') {
+      for (const addItem in act.extraData) {
+        state[addItem] = act.extraData[addItem];
+      }
+    }
+
+    if (act.type === 'removeProperties') {
+      for (const removeItem of act.keysToRemove) {
+        // eslint-disable-next-line no-console
+        console.log(state, state[removeItem], removeItem);
```
vlad-demchuk: do not leave debug statements, it should be only during developing

vlad-demchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      for (const removeItem of act.keysToRemove) {
+        delete state[removeItem];
+      }
+    }
+
+    if (act.type === 'clear') {
+      for (const st in state) {
+        delete state[st];
+      }
+    }
```
nazarmatsevych: Rewrite it using switch/case statement because we rely on single value

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
nazarmatsevych:
```suggestion
  for (const action of actions) {
```

nazarmatsevych (changes requested):

volodymyr-soltys97 (approved):

PR #2409, opened 2023-06-03
Develop

NataliMax:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const Action of actions) {
+    switch (Action.type) {
```
VitaliyBondarenko1982, line 9:
```suggestion
  for (const action of actions) {
    switch (action.type) {
```
general convention to name variables in camelCase in js. so first word should be from lower case

VitaliyBondarenko1982 (approved):
Well done.
Just pay attention on comment about naming variables

src/transformState.js
NataliMax: Thanks, I'll be attention

PR #2410, opened 2023-06-03
add task js stateful object

Lazarend:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: use `switch` block instead of multiple `if` blocks

TarasHoliuk (changes requested):

src/transformState.js
Lazarend: Hi. Ok, I try it.

OleziO (approved):

PR #2412, opened 2023-06-04
Solution

hrynkevych:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
etojeDenys: Instead of using 'switch (true)', it is more appropriate to use 'switch (action.type)'. This will make the code easier to read and understand.

etojeDenys (changes requested):
add default case

mykhalenych (approved):

PR #2415, opened 2023-06-06
Solution Develop

RuslanMelnykFE:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const newState = state;
```
anastasiiavorobiova: You shouldn't assign the state object to newState directly as it creates a reference to the same object, rather than a new one. Use a spread operator or Object.assign to create a new object.

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+    }
+
+    if (type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+    }
+  }
+
+  return newState;
```
anastasiiavorobiova, line 34: Since the original state object is being modified, newState wouldn't be new object

anastasiiavorobiova (changes requested):

src/transformState.js
RuslanMelnykFE: Why do the tests fail when using the spread operator?

nazarmatsevych (approved):
👍

PR #2416, opened 2023-06-06
statefulObject

OleksiiBielovolov:


src/transformState.js
```diff
@@ -5,7 +5,45 @@
...
+  for (const key of keysToRemove) {
+    delete state[key];
+    delete state.key;
+  }
+}
+
+function clear(state) {
+  for (const key in state) {
+    delete state[key];
+    delete state.key;
```
anastasiiavorobiova: The line 'delete state.key;' is unnecessary and should be removed, as 'delete state[key];' is sufficient to remove the property.

src/transformState.js
```diff
@@ -5,7 +5,45 @@
...
+  //   state.key = extraData[key];
+  // }
+
+  Object.assign(state, extraData);
+}
+
+function removeProperties(state, keysToRemove) {
+  for (const key of keysToRemove) {
+    delete state[key];
+    delete state.key;
```
anastasiiavorobiova: The line 'delete state.key;' is unnecessary and should be removed, as 'delete state[key];' is sufficient to remove the property.

anastasiiavorobiova (changes requested):

nazarmatsevych (approved):
👍

PR #2417, opened 2023-06-06
Solution

rvrazovskyi:


src/transformState.js
```diff
@@ -4,8 +4,30 @@
...
+          delete state[key];
+        }
+      }
+    } else if (type === 'clear') {
+      for (const key in state) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    }
```
4ksandrey: Use switch statement if you have limited amount of conditions. switch (action.type) Look at the checklist.md

4ksandrey (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,39 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function modifyState(state, actions) {
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
```
4ksandrey, line 11: It would be better to assign action type inside the switch statement
switch (action.type)

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+      case 'addProperties':
+        const { extraData } = action;
+
+        Object.assign(state, extraData);
```
4ksandrey, line 15: it isn't necessary to declare const { extraData } 
it would be better to assign it right away Object.assign(state, action.extraData);
fix it everywhere

src/transformState.js
```diff
@@ -4,8 +4,39 @@
...
+        break;
+
+      case 'removeProperties':
+        const { keysToRemove } = action;
```
4ksandrey, line 19: Same as above

4ksandrey (changes requested):

src/transformState.js
rvrazovskyi: I don't understand why i can't push changes
Screenshot_38

PR #2418, opened 2023-06-06
Adde task solution

ruslan-ihnatenko:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const a of actions) {
+    if (a.type === 'addProperties') {
+      Object.assign(state, a.extraData);
+    } else if (a.type === 'removeProperties') {
+      for (const b of a.keysToRemove) {
+        delete state[b];
+      }
+    } else {
```
OleziO: Replace with switch case construction

src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const a of actions) {
+    if (a.type === 'addProperties') {
+      Object.assign(state, a.extraData);
+    } else if (a.type === 'removeProperties') {
+      for (const b of a.keysToRemove) {
```
OleziO: Rename b to more meaningful name

src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+        delete state[b];
+      }
+    } else {
+      for (const c in state) {
```
OleziO: Rename c to more meaningful name

OleziO (changes requested):

sTorba24 (approved):
Well done! Keep it up :)

PR #2419, opened 2023-06-06
Solution

rvrazovskyi:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        }
+      }
+    } else if (type === 'clear') {
+      for (const key in state) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    }
+  }
```
4ksandrey: Use switch statement, also apply changes which were mentioned on the previous reviews about const { type } = action bacause it isn't necessary

4ksandrey (changes requested):

4ksandrey (approved):

PR #2422, opened 2023-06-07
task implementation

Titech0:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+      case 'addProperties':
+        const dataToExtract = action.extraData;
+
+        Object.assign(state, { ...dataToExtract });
+        break;
+
+      case 'removeProperties':
+        const dataToRemove = action.keysToRemove;
+
+        [ ...dataToRemove ].forEach(list => delete state[list]);
```
Viktor-Kost: 1. no needs to create one more variable and spread it
2. it's better to name 'property' instead of `list`
```suggestion
        action.keysToRemove.forEach(property => delete state[property]);
```

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        const dataToExtract = action.extraData;
+
+        Object.assign(state, { ...dataToExtract });
```
Viktor-Kost:
```suggestion
        Object.assign(state, action.extraData);
```

Viktor-Kost (changes requested):
Hello!)
please apply minor fixes

vlad-demchuk (approved):

PR #2425, opened 2023-06-08
add task solution

MaxSchmide:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const tempState = state;
```
ilona-borkovska, line 8: you don't need to create extra variable for state because you still work with initial obj ("state")

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+
+  actions.map(action => {
```
ilona-borkovska, line 10: you could return this map straightly

ilona-borkovska (approved):
Good job 👍

juljaVysotska (approved):
🔥

PR #2426, opened 2023-06-08
stateful-object

Krivets:


witflash (approved):
Good work!
Don't forget the `switch/case` should have a `default` statement. 
You can handle some errors there or do nothing for example.

PR #2427, opened 2023-06-09
create pr

Yetzir:
js_stateful-object

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -14,7 +14,7 @@ function transformState(state, actions) {
         break;
       case 'removeProperties':
         for (const key of Object.values(keysToRemove)) {
-          if (state[key]) {
+          if (key in state) {
             delete state[key];
           }
         }
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const comands of actions) {
+    if (comands.type === 'addProperties') {
+      for (const [key, value] of Object.entries(comands.extraData)) {
+        state[key] = value;
+      }
+    }
+
+    if (comands.type === 'removeProperties') {
+      for (const key of Object.values(comands.keysToRemove)) {
+        delete state[key];
+      }
+    }
+
+    if (comands.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
 }
 
 module.exports = transformState;
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,23 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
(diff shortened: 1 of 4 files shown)


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        delete state[key];
+      }
+    }
+
+    if (comands.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
```
Volodymyr-Melnyk-Fron10der: use switch/case/default structure

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const comands of actions) {
```
```suggestion
  for (const { type, ...so on } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (comands.type === 'addProperties') {
+      for (const [key, value] of Object.entries(comands.extraData)) {
+        state[key] = value;
+      }
+    }
+
+    if (comands.type === 'removeProperties') {
+      for (const key of Object.values(comands.keysToRemove)) {
+        delete state[key];
```
Volodymyr-Melnyk-Fron10der: check does property exist in the object before deleting it

Volodymyr-Melnyk-Fron10der (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+            delete state[key];
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
VitaliyBodnar: add default case (might be even empty - with break only) - that's a good practice

VitaliyBodnar (changes requested):
Well done!
Let's improve your solution 🧑‍💻

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, extraData, keysToRemove } of actions) {
+    switch (type) {
+      case 'addProperties':
+        for (const [key, value] of Object.entries(extraData)) {
+          state[key] = value;
+        };
```
DarkMistyRoom, line 13: consider using Object.assign. It may be simpler

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+    }
```
DarkMistyRoom, line 28: empty default case makes no sense. Handle the case when the action type is wrong. Inform the user - console.log or throw an Error

DarkMistyRoom (approved):

PR #2429, opened 2023-06-09
add task solution

irynabolbat:


KandaskalovOleg (approved):
Чудова робота

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+  for (const action of Object.values(actions)) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const [addKey, addValue] of Object.entries(action.extraData)) {
+          changedState[addKey] = addValue;
+        }
+        break;
+
+      case 'removeProperties':
+        for (const removeKey of Object.values(action.keysToRemove)) {
```
anstsot, line 19: `keysToRemove` is an array so you use directly:
```suggestion
        for (const removeKey of action.keysToRemove) {
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const changedState = state;
+
+  for (const action of Object.values(actions)) {
```
anstsot, line 10: can be simplified to
```suggestion
  for (const action of actions) {
```

anstsot (approved):
But please pay attention to my comments to simplify the code

src/transformState.js
irynabolbat: Thank you 😊

PR #2430, opened 2023-06-09
Solution

dianadiv:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
olena-ole: 1. Here a destructuring assignment can be used:
    _const { type, extraData, keysToRemove } = action;_

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+      case 'addProperties' :
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties' :
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
```
olena-ole: 2. It might be useful to create separate functions for lines 11, 15-17 and 21-23, so that all your functions have single responsibility and can be used in future

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear' :
```
olena-ole, line 20: The space before colon in lines 10, 14, 20 is unnecessary

olena-ole (approved):
Great job, added a few suggestions for you to consider!

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'Error';
```
```suggestion
        return throw new Error('Invalid type');
```

src/transformState.js
tetiana-ivanchenko: use destructuring assignment

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const { type, extraData, keysToRemove } of actions) {
```
anstsot, line 8: for better readability it can be rewritten:
```suggestion
  for (const action of actions) {
    const { type, extraData, keysToRemove }  = action;
```

anstsot (approved):

PR #2431, opened 2023-06-10
Solution

MariiaMonakhova:


cebanoleksandr (approved):
Great job:)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Error');
```
tetiana-ivanchenko, line 33: use more informative message, like "Invalid type"

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        const { extraData } = action;
```
tetiana-ivanchenko, line 11: use destructuring assignment in the upper scope for all cases

tetiana-ivanchenko (approved):

PR #2432, opened 2023-06-10
Solution

olena-ole:


adv555 (approved):
Interesting solution Will keep in mind :blush: , approve ✔️

IraShtynda:
Looks good to me, approve!

src/transformState.js
```diff
@@ -5,7 +5,44 @@
...
+        return;
+    }
+  }
+}
+
+function addProperties(obj, properties) {
+  for (const property in properties) {
+    obj[property] = properties[property];
+  }
+}
```
NataliGru: It will be better if you use `Object.assign(target, source)` instead of using a `for...in` loop ✨

src/transformState.js
```diff
@@ -5,7 +5,44 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
NataliGru: Oh, that's a really cool approach! I like it too! ✨

NataliGru (approved):

tetiana-ivanchenko (approved):

PR #2434, opened 2023-06-10
Solution

yehor-kor:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      Object.assign(modifiedState, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        delete modifiedState[key];
+      }
+    } else if (action.type === 'clear') {
+      for (const key in modifiedState) {
+        delete modifiedState[key];
+      }
+    }
```
OleziO: Replace with switch case construction.

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
```
OleziO: Remove all comments.

OleziO (changes requested):

src/transformState.js
yehor-kor: Done

src/transformState.js
yehor-kor: Thank you for your remark! I replaced that.

src/transformState.js
```diff
@@ -1,11 +1,29 @@
 'use strict';
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
-  // write code here
+  const modifiedState = state;
```
agniya-i, line 4: Not necessary to create a separate variable, since in the case of objects you assign reference, so by changing `modifiedState` you change  `state` as well

agniya-i (approved):

PR #2435, opened 2023-06-11
solution

cebanoleksandr:


src/transformState.js
```diff
@@ -5,7 +5,39 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    reducer(action);
+  }
+
+  function reducer(action) {
+    switch (action.type) {
```
tetianakostiuk1: you can use object destructuring to directly access its properties instead of dot notation
 --> const { type, extraData, keysToRemove } = action;

src/transformState.js
```diff
@@ -5,7 +5,39 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    reducer(action);
```
tetianakostiuk1: I think that using reducer function in this exact case is unnecessary and adds more complexity,
directly handling the actions within the  function is a simpler and more efficient approach

src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
+
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
tetianakostiuk1: it's better to use object.hasOwnProperty() to ensure that only the object's own properties are considered because the loop will also iterate over inherited properties.
also, it potentially improves performance)

tetianakostiuk1 (changes requested):

tetianakostiuk1 (approved):
👍

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+        return state;
```
tetiana-ivanchenko: [CODE STYLE]: switch/case should always have default case for error handling.

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+  for (const action of actions) {
+    reducer(action);
+  }
+
+  function reducer({ type, extraData, keysToRemove }) {
+    switch (type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
```
tetiana-ivanchenko: you can use Objesct.assign

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+      case 'addProperties':
+        Object.assign(state, extraData);
+
+        break;
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
And678: No need for this check here, `delete` works fine for non existent properties.
```suggestion
          delete state[key];
```

And678: E.g. no errors here:
```js
const a = {};
delete a["b"];
```

And678 (changes requested):
Great work, but your reducer function indeed complicates things a bit.
Mostly because it is declared in a function and accesses the parent scope (closure).

Unnecessary use of closures make code harder to read, better move your `reducer` outside of `transformState` and declare it like this:
```js
function reducer(state, { type, extraData, keysToRemove }) {
```

anstsot (approved):

PR #2438, opened 2023-06-11
transformState solution

NataliGru:


artkas03 (approved):
good job)

melnyksergey7:
great job!

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+}
+
+function addProperties(object, extraData) {
+  Object.assign(object, extraData);
+}
+
+function removeProperties(object, keys) {
+  for (const key of keys) {
+    delete object[key];
+  }
```
MarianVolytskyi, line 37: we can use loop and Object.assign inside switch, it no nessesary to create additional functions

MarianVolytskyi (approved):

MarianVolytskyi (approved):
Good job!)

src/transformState.js
NataliGru: Yeah, I understand. But using a separate function is more practical because you can use it in other places without repeating the same code. It follows the main rule `don't repeat yourself`. It makes the code easier to understand and has a nicer structure)

And678 (approved):
Nice! Your style is very clean, great idea to move similar logic to separate functions.

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+        addProperties(state, extraData);
+        break;
+
+      case 'removeProperties':
+        removeProperties(state, keysToRemove);
+        break;
+
+      case 'clear':
+        removeProperties(state, Object.keys(state));
+        break;
```
And678, line 22: :fire:

PR #2439, opened 2023-06-11
Solution

anya-shults:


src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
 function transformState(state, actions) {
-  // write code here
+  const ADD_OPTION = 'addProperties';
+  const REMOVE_OPTION = 'removeProperties';
+  const CLEAR_OPTION = 'clear';
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ADD_OPTION:
+        for (const key in action.extraData) {
```
zoriana-melnychuk: Checklist: [CODE STYLE]: don't use `for in` loop for iterating over array

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ADD_OPTION = 'addProperties';
+  const REMOVE_OPTION = 'removeProperties';
+  const CLEAR_OPTION = 'clear';
```
zoriana-melnychuk, line 10: i'm not sure If it's really necessary to create variables for these values

zoriana-melnychuk (changes requested):

src/transformState.js
anya-shults: It is an outdated file

zoriana-melnychuk (approved):
Good job :)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ADD_OPTION:
+        for (const key of Object.keys(action.extraData)) {
+          state[key] = action.extraData[key];
+        }
```
And678: Check out [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign). You can use it here to avoid a nested loop.

And678: [Checklist #5](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ADD_OPTION = 'addProperties';
+  const REMOVE_OPTION = 'removeProperties';
+  const CLEAR_OPTION = 'clear';
```
And678, line 10: Nice usage of CONSTANTS :fire:

And678 (changes requested):
Good work, one thing to improve and this is done :100:

And678 (approved):
:+1: :+1:

PR #2441, opened 2023-06-11
Solution

fawnique:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
-  // write code here
-}
+  for (const action of actions) {
+    switch (action['type']) {
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
+        break;
+
+      case 'removeProperties':
+        action['keysToRemove'].forEach(key => delete state[key]);
```
artkas03: You can use object._keyname_ but not object['_keyname_'] when you know exactly which key do you have. It would work the same way but it is a better practice if you know exactly name of key and it doesn't start with number or special character.
Good job)

artkas03 (approved):

And678 (changes requested):
Good work! 
Indeed, @artkas03 is correct. Use dot notation (object.keyname), when your keyname is known and not a variable.

And678 (commented):
Nice :fire: , approve :heavy_check_mark:

And678 (approved):
Approve 2

PR #2442, opened 2023-06-11
add task solution

parmenid89:


NazarBaraban (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      // eslint-disable-next-line
+      case 'addProperties': {
+        Object.assign(state, action.extraData);
+      } break;
```
And678: There needs to be a reason to disable eslint for certain lines. Here, it is easily fixable, same for the rest of cases.
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
```

And678 (changes requested):
GJ, one thing to fix below.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const stateKey in state) {
+          delete state[stateKey];
+        }
+        break;
+      default:
+        break;
```
tetiana-ivanchenko: default case should handling error

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
```
```suggestion
        break;
        
      case 'removeProperties':
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):
GJ

PR #2452, opened 2023-06-12
Solution

melnyksergey7:


IgorOksentyuk (approved):
Clear and understandable work! GJ!)

klmyana (approved):
Great code!

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+          }
+        }
+
+        break;
+
+      case CLEAR_TYPE:
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
anstsot: no need to check the the `key` is in `state` object
```suggestion
          delete state[key];
```

src/transformState.js
```diff
@@ -5,7 +5,41 @@
...
+        break;
+
+      case REMOVE_PROPERTY_TYPE:
+        for (let j = 0; j < action.keysToRemove.length; j++) {
+          const key = action.keysToRemove[j];
+
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
anstsot: it can be simplified bu using `for ... of` construction and there is no need to check if the key is in `state`

anstsot (changes requested):
Almost done :+1:
Let's improve a little bit the code

anstsot (approved):

PR #2453, opened 2023-06-12
solution

GGsmou:


OKhalanskyi (approved):
Good job👍

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+      default:
+        break;
```
anstsot, line 27: it would be great to `throw new Error` on default case with message about wrong action type

anstsot (approved):

PR #2454, opened 2023-06-12
add task solution

MarianVolytskyi:


maksymPtashnyk (approved):
Good job, maybe don't need to iterate the same variable `key`  in different loops)

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+
+  for (const curentAction of actions) {
+    switch (curentAction.type) {
```
tetiana-ivanchenko, line 10: optional impovements:
1. `action` instead of `curentAction` will be enought (especially in this word a typo)
2. use destructuring assignment

tetiana-ivanchenko (approved):

PR #2455, opened 2023-06-12
add task solution

NazarBaraban:


src/transformState.js
```diff
@@ -5,7 +5,38 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
AlyonaSarapina: Default case should handle an error.

AlyonaSarapina (changes requested):

AlyonaSarapina (approved):
good job !

src/transformState.js
```diff
@@ -5,7 +5,38 @@
...
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        const keysToRemove = action.keysToRemove;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
YegorKorenkov, line 24: Redundant check, because the **delete** operator will not return an error if the matching key is not found

YegorKorenkov (approved):
Good! Please pay attention on comment!

PR #2457, opened 2023-06-12
Finished task

IgorOksentyuk:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const allActions = Object.values(actions);
```
AnastasiiaKhemych: You don't need this `Object.values(actions)` because you already have array of actions

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+  for (let i = 0; i < allActions.length; i++) {
```
AnastasiiaKhemych: You should use `for of`  because you don't need index. 
If you use `for of` you don't need reassign `const action = allActions[i];`

AnastasiiaKhemych (changes requested):

AnastasiiaKhemych (approved):
God job!!!

YegorKorenkov (approved):
Looks good

PR #2459, opened 2023-06-12
Solution js_stateful-object

Svitlanchick:
https://github.com/mate-academy/js_stateful-object/pull/2459

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      const obj = action.extraData;
+
+      for (const key in obj) {
+        state[key] = obj[key];
+      }
+
+      // console.log(action.type);
+    } else if (action.type === 'removeProperties') {
+      const array = action.keysToRemove;
+
+      for (const item of array) {
+        if (state[item] !== undefined) {
+          delete state[item];
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
+  }
 }
 
 module.exports = transformState;
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -7,25 +7,28 @@
 
 function transformState(state, actions) {
   for (const action of actions) {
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      const obj = action.extraData;
+
+      for (const key in obj) {
+        state[key] = obj[key];
+      }
+
+      // console.log(action.type);
```
agniya-i: Don't forget to remove comments

src/transformState.js
```diff
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
agniya-i: It would be better to use `switch/case` instead of `if/else` ([checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) p2 + p4)

src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+      const obj = action.extraData;
```
agniya-i: Not necessary to create this extra variable.
Use just `action.extraData` or destructure `action`

agniya-i (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,34 @@
...
+      }
+        break;
+
+      case (action.type === 'clear'): {
+        for (const key in state) {
+          delete state[key];
+        }
+      }
+    }
+  }
```
agniya-i:
```suggestion
    switch (action.type) {
      case 'addProperties':
        for (const key in action.extraData) {
          state[key] = action.extraData[key];
        }
        break;

      case 'removeProperties' : 
        const array = action.keysToRemove;

        for (const item of array) {
          if (state[item] !== undefined) {
            delete state[item];
          }
        }
      
        break;

      case 'clear' : {
        for (const key in state) {
          delete state[key];
        }
      }
    }
  }
```

agniya-i (changes requested):
Almost done! 
Check syntax of switch/case one more time 

Also, don't forget about default case

qoa11a (approved):

PR #2460, opened 2023-06-12
Solution

outofs:


src/transformState.js
```diff
@@ -1,11 +1,36 @@
-'use strict';
+"use strict";
```
vladmakohon, line 1: Be consistent with your usage of quotes. In the previous commit, you had single quotes, now you have double quotes. Also, we have a 'single-quote' eslint rule in this project.

src/transformState.js
```diff
@@ -1,11 +1,36 @@
...
+          delete state[key];
+        });
+        break;
+
+      default:
+        throw new Error("Unknwon action!");
+    }
+  }
+
+  return state;
```
vladmakohon, line 33: Nothing wrong here, in my opinion. But I guess this line is just unrequired.

vladmakohon (approved):
Great work! I have a few suggestions for you.

And678 (approved):
Approve :heavy_check_mark: , but check out @vladmakohon s comments

PR #2462, opened 2023-06-12
Solution

MelnychukVol:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        continue;
```
volodymyr-soltys97, line 24: The 'continue' statement is not necessary in this context. You can simply remove it.

volodymyr-soltys97 (approved):
Nice work 👍

PR #2463, opened 2023-06-12
Solution

AytajBabayeva:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can destructure the action to get all the needed properties of it (type, keysToRemove, extraData)

polosanya (approved):
Good job!

PR #2464, opened 2023-06-12
feat: transform state

OKhalanskyi:


NadiaKovalchuk (approved):

src/transformState.js
```diff
@@ -6,6 +6,41 @@
  */
 function transformState(state, actions) {
   // write code here
```
And678: Remove placeholder comments
```suggestion
```

src/transformState.js
```diff
@@ -6,6 +6,41 @@
...
+  function addProperties(extraData) {
+    for (const [key, value] of Object.entries(extraData)) {
+      state[key] = value;
+    }
```
And678: Better use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) here, to avoid nested loop

And678 (changes requested):
Nice, but check my comments below

tetiana-ivanchenko (approved):

PR #2465, opened 2023-06-12
solution

ihorhladkov:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    switch (act.type) {
+      case 'addProperties':
+        Object.assign(state, act.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of act.keysToRemove) {
+          delete state[key];
+        }
+        break;
```
ArtemAndrushchenko, line 18: Checklist: [CODE STYLE]: switch/case should always have default case for error handling.

ArtemAndrushchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
```

tetiana-ivanchenko (approved):

PR #2468, opened 2023-06-13
Solution

dipinchuk:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const data in action.extraData) {
+          state[data] = action.extraData[data];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const key in action.keysToRemove) {
```
ovynarchuk: [CODE STYLE]: don't use for in loop for iterating over array
keysToRemove is an array, please don't use for in loop here as it says in the checklist. I suggest using for of loop

ovynarchuk (changes requested):
please check the comment and change your code

TamaraShukalo (approved):
Good!✨

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const data in action.extraData) {
+          state[data] = action.extraData[data];
+        }
```
And678: You can use [`Object.assign`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) here to avoid a nested loop.

And678 (approved):
Good work, approve :heavy_check_mark: , but see my comment

PR #2470, opened 2023-06-13
Add task solution

vladmakohon:


FlameInBrains (approved):
Great code.  A little remark: maybe it would be better to use 'throw' in the default case for errors

src/transformState.js
```diff
@@ -5,7 +5,39 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
And678: How about using this syntax to destructure an object here, instead of every case of switch.

```suggestion
  for (const { type, extraData, keysToRemove } of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+        keysToRemove.forEach(key => {
+          delete state[key];
+        });
+
+        break;
+
+      case 'clear':
+        const stateKeys = Object.keys(state);
+
+        stateKeys.forEach(key => {
```
And678: No need to declare variables to be used only once on the next line 
```suggestion
        Object.keys(state).forEach(key => {
```

And678 (changes requested):
Looks good, but can be a bit improved

tetiana-ivanchenko (approved):

PR #2471, opened 2023-06-13
solution

Valentyna08:


perfectAcedia (approved):
Instead of hardcoding the name of the properties('addProperties' etc.), it would be better to create corresponding variables.

tetiana-ivanchenko (approved):

PR #2472, opened 2023-06-13
implement transformState

kochetov-yevhen-may29:


src/transformState.js
```diff
@@ -5,7 +5,45 @@
...
+          state[key] = extraData[key];
+        }
+        break;
+      }
+
+      case REMOVE_PROPERTIES: {
+        const { keysToRemove } = action;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
AnatoliiLykhohodenko: better use Object.hasOwn(). It prevents inheritance

src/transformState.js
```diff
@@ -5,7 +5,45 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ADD_PROPERTIES = 'addProperties';
+  const REMOVE_PROPERTIES = 'removeProperties';
+  const CLEAR = 'clear';
```
AnatoliiLykhohodenko: I'm not sure about this is helpful because you used this variables only 1 time each)

AnatoliiLykhohodenko (approved):
Good job, but I left some advices

src/transformState.js
```diff
@@ -5,7 +5,45 @@
...
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case ADD_PROPERTIES: {
+        const { extraData } = action;
+
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
```
And678: How about using [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) here to avoid a nested loop.

And678 (changes requested):
Check my comment below

And678 (changes requested):
Sync

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
```suggestion
    const { type } = action;
```

add here all properties instead of destructuring assignment in each case

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2479, opened 2023-06-13
add task solution

ivankovbohdan:


src/transformState.js
```diff
@@ -4,8 +4,35 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
tetiana-ivanchenko: add here all properties instead of using destructuring assignment in the each case

src/transformState.js
```diff
@@ -4,8 +4,35 @@
...
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
+
+        Object.assign(state, extraData);
+        break;
+      case 'removeProperties':
```
```suggestion
        break;
        
      case 'removeProperties':
```

tetiana-ivanchenko (changes requested):

anstsot (approved):

PR #2482, opened 2023-06-13
Solution

olegpaliy:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'removeProperties':
+        for (const val of value.keysToRemove) {
+          delete state[val];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
tetiana-ivanchenko, line 24: switch/case should always have default case for error handling.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const value of actions) {
```
tetiana-ivanchenko: image

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    switch (value.type) {
+      case 'addProperties':
+        Object.assign(state, value.extraData);
+        break;
+      case 'removeProperties':
```
```suggestion
        break;
        
      case 'removeProperties':
```

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (changes requested):
you should fix all comments before re-requesting review
https://github.com/mate-academy/js_stateful-object/pull/2482/files#r1229097367

tetiana-ivanchenko (approved):

PR #2483, opened 2023-06-13
solution

FlameInBrains:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
tetiana-ivanchenko: [CODE KNOWLEDGE]: Remember, if property key is a variable - use brackets object[key]. If it's called key - use dot access object.key.

tetiana-ivanchenko (changes requested):

tetiana-ivanchenko (approved):

PR #2484, opened 2023-06-13
Solution

perfectAcedia:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+  const addOption = 'addProperties';
+  const removeOption = 'removeProperties';
+  const clearOption = 'clear';
+
+  for (const action of actions) {
+    switch (action.type) {
+      case addOption:
+        Object.assign(state, action.extraData);
+        break;
+      case removeOption:
```
```suggestion
        break;
        
      case removeOption:
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+        break;
+      case clearOption:
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+
+        break;
+      default:
+        return 'There is no options to change';
```
```suggestion
        throw new Error('Invalid type')
```

tetiana-ivanchenko (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const addOption = 'addProperties';
+  const removeOption = 'removeProperties';
+  const clearOption = 'clear';
```
anstsot, line 10: it's better to use uppercase naming for constants
```suggestion
  const ADD_OPTION = 'addProperties';
  //...
```

anstsot (approved):

PR #2485, opened 2023-06-13
add task solution

EvgeniyChumak:


src/transformState.js
```diff
@@ -5,7 +5,38 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
```
tetiana-ivanchenko: add here all properties instead of using destructuring assignment in the each case

tetiana-ivanchenko (changes requested):

And678 (approved):
Good work, approve :heavy_check_mark:

PR #2489, opened 2023-06-13
added task solution

vlopatko:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
-  // write code here
-}
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'addProperties': {
+        Object.assign(state, extraData);
+        break;
+      }
```
tetiana-ivanchenko, line 15:
```suggestion
      case 'addProperties':
        Object.assign(state, extraData);
        break;
```

tetiana-ivanchenko: according to documentation this braces is redundant

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+      case 'clear': {
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        };
+        break;
+      }
+
+      default: {
+        throw new Error('Error');
```
tetiana-ivanchenko, line 32:
```suggestion
        throw new Error('Invalid action type');
```

tetiana-ivanchenko (approved):

PR #2492, opened 2023-06-14
Solution

valet2102:


Vayts (approved):

Vayts:
good job!

Vayts (approved):
good job!

src/transformState.js
```diff
@@ -6,6 +6,40 @@
...
+  // for (let i = 0; i < keys.length; i++) {
+  //   delete baseObj[keys[i]];
+  // }
+  for (const key of keys) {
+    delete baseObj[key];
+  }
+}
+*/
+
+// addProperties
```
qoa11a: I think you should remove redundant comments.

src/transformState.js
```diff
@@ -6,6 +6,40 @@
...
+
+function addProperties(baseObj, properties) {
+  for (const property in properties) {
+    baseObj[property] = properties[property];
+  }
 }
```
qoa11a: It's complicated logic. I recommend you read about an `assign` method.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/transformState.js
```diff
@@ -6,6 +6,40 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
```
qoa11a: You may use the `for ... of` loop here.

src/transformState.js
```diff
@@ -6,6 +6,40 @@
...
+      case 'addProperties':
+        addProperties(state, actions[i].extraData);
+
+        break;
+      case 'removeProperties':
+        /* removeProperties(state, actions[i].keysToRemove) */
+        actions[i].keysToRemove.forEach(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
```
qoa11a: Don't forget to add `break` at the end of any case. Also, you forgot to add the `default` case. Please, read the [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) carefully (point 4).

qoa11a (changes requested):
Thanks for submitting your homework! I see you put in a lot of effort, but there's room for improvement.

src/transformState.js
```diff
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
   // write code here
```
Moroz-Dmytro: This comment also should be deleted

src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+  actions.forEach(item => {
```
Moroz-Dmytro: image

src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+        break;
+
+      case `clear`:
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      default:
+        break;
```
Moroz-Dmytro: image


In this case would be nice to throw an error

Moroz-Dmytro (changes requested):
Please, take a look at checklist more attentively next time, before sending your solution to review

src/transformState.js
valet2102: but its not for of loop...

Moroz-Dmytro: Naming is important in every loop 🙃

Moroz-Dmytro (approved):
👍

PR #2494, opened 2023-06-15
add task solution

oleksandrshtonda:


src/transformState.js
```diff
@@ -1,11 +1,29 @@
...
+        delete state[key];
+      }
+      continue;
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        delete state[keyToRemove];
+      }
+      continue;
```
volodymyr-soltys97: Remove the unnecessary 'continue' statement, as it does not provide any benefit in this case.

src/transformState.js
```diff
@@ -1,11 +1,29 @@
...
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+      continue;
```
volodymyr-soltys97: Remove the unnecessary 'continue' statement, as it does not provide any benefit in this case.

volodymyr-soltys97 (changes requested):
Good job 👍 
But let's make your code more perfect

src/transformState.js
oleksandrshtonda: Ok no problem

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+    if (action.type === 'removeProperties') {
+      for (const keyToRemove of action.keysToRemove) {
+        delete state[keyToRemove];
+      }
+    }
+
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+  }
```
volodymyr-soltys97, line 24: You can use a switch construction

volodymyr-soltys97 (approved):
Great job 👍

PR #2495, opened 2023-06-15
transformState

zipic:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    }
+
+    if (type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
loralevitska: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
+    const { type } = action;
```
loralevitska: Destructure all keys: extraData, keysToRemove

loralevitska (changes requested):
Almost done) Let's improve

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
```
OleziO: Remove comments.

OleziO (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,26 @@
...
+      case 'addProperties':
+        Object.assign(state, extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
```
loralevitska, line 17:
```suggestion
      case 'removeProperties':
        for (const key of keysToRemove) {
          delete state[key];
        }
        break;
        
      case 'clear':
```

loralevitska (approved):
Well done)

PR #2496, opened 2023-06-16
stateful object solved

roman-tarasov90:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    for (const key in action) {
+      if (action[key] === 'addProperties') {
```
VitaliyBondarenko1982:
```suggestion
    for (const key in action) {
      if (action[key] === 'addProperties') {
```
redundant for in loop. destructure action and check type
const { type, extraData , keysToRemove } = action;

VitaliyBondarenko1982: also instead if consider to use switch/case 
```
  switch(type) {
    //do something
  }

VitaliyBondarenko1982 (changes requested):

mykhalenych (approved):

PR #2499, opened 2023-06-18
solution

DanielLinker17:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
SerhiiKirik: Why true?
Put here valid data
```suggestion
    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case action.type === 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case action.type === 'clear':
```
SerhiiKirik:
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;

      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;

      case 'clear':
```

src/transformState.js
SerhiiKirik: I recommend you take a look at this [link](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const el in state) {
```
```suggestion
        for (const key in state) {
```

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
```suggestion
        for (const keyToRemove of action.keysToRemove) {
```

SanyaBratashchuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
OleziO, line 10: Move to the object. Example of usage: `case ActionsTypes.AddProperties:`

OleziO (changes requested):

PR #2500, opened 2023-06-18
task was solved

LinaDo1609:


src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
  * @param {Object} state
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    switch (obj.type) {
+      case "addProperties":
+        for (const keys in obj.extraData) {
+          state[keys] = obj.extraData[keys];
```
SerhiiKirik: use `Object.assing` here for it

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+        break;
+
+      case "removeProperties":
+        for (const key of obj.keysToRemove) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
```
SerhiiKirik: it's not a default case, it's a `clear`, change it

SerhiiKirik (changes requested):
Well done 🔥 
Let's make some improvements 🧑‍💻

src/transformState.js
LinaDo1609: @SerhiiKirik I am not sure that I can use this method. 
The task is to modify "state" according to the type, if i use Object.assing, it means that I create a new object.
but in this part of the code i have to add extra object to existing object 'state'

src/transformState.js
LinaDo1609: i corrected it

src/transformState.js
LinaDo1609: i corrected it

src/transformState.js
```diff
@@ -1,11 +1,34 @@
...
+        }
+
+        break;
+
+      case "clear":
+        for (const del in state) {
+          delete state[del];
+        }
+
+        break;
```
loralevitska, line 27: But `default` case still need. Always use `default` case for each switch construction

src/transformState.js
```diff
@@ -1,11 +1,34 @@
-'use strict';
+"use strict";
 
 /**
  * @param {Object} state
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
loralevitska, line 8: Destructure values from `action`
```suggestion
  for (const action of actions) {
```

loralevitska (approved):
Good job!

PR #2501, opened 2023-06-19
Solution

Luk2asz:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
DorotaLeniecDev:
```suggestion
    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+      case action.type === 'addProperties':
```
DorotaLeniecDev:
```suggestion
      case 'addProperties':
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          state[key] = action.extraData[key];
+        }
+        break;
+
+      case action.type === 'removeProperties':
+        const actionKeysToRemove = action.keysToRemove;
+
+        for (let i = 0; i < actionKeysToRemove.length; i++) {
+          delete state[action.keysToRemove[i]];
+        }
```
DorotaLeniecDev: You can make it a bit simpler by using for of loop:
```suggestion
 for (const key of action.keysToRemove) {
          delete state[key];
 }
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case action.type === 'clear':
+        for (const keyState in state) {
+          delete state[keyState];
+        }
+        break;
+
+      default:
+        return;
```
DorotaLeniecDev: Here you can throw some error to inform user that something went wrong:
```suggestion
        throw Error(`Action type ${action.type} is not found`);
```

DorotaLeniecDev (changes requested):
Good job, few improvements needed

DorotaLeniecDev (approved):
Well done!

PR #2502, opened 2023-06-19
Make tasks

F-u-n-H-o-u-s-e:


src/transformState.js
```diff
@@ -4,8 +4,29 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    if (actions[i].type === 'addProperties') {
+      for (const keyAdd in actions[i].extraData) {
```
demetrxx: `keyToAdd`, `keyToRemove` would be more decriptive

demetrxx (changes requested):
Good job! But let's improve your code a bit. Use switch case instead of if-else pattern

demetrxx (changes requested):
In my previous review, I mentioned that your situation requires using switch case, seems like you missed that.

Please, introduce the approriate changes and re-submit the task.

F-u-n-H-o-u-s-e:
Thank you for repeate me. I am sorry, dont no how i dosent saw this message.

VitaliyBondarenko1982 (approved):

PR #2503, opened 2023-06-19
add solution

tarovyk:


src/transformState.js
```diff
@@ -6,6 +6,29 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const action = actions[i];
```
loralevitska: Destructure values from `action`

src/transformState.js
```diff
@@ -6,6 +6,29 @@
...
+          delete state[action.keysToRemove[key]];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return state;
```
loralevitska:
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const key in action.keysToRemove) {
          delete state[action.keysToRemove[key]];
        }
        break;
        
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
      default:
        return state;
```

loralevitska (changes requested):
Almost perfect) Lets improve

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const { type, extraData, keysToRemove } = actions[i];
+
+    switch (type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

loralevitska (approved):
🔥

PR #2505, opened 2023-06-20
task done

IntensoSergey:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
Anastasiia-Svintsova: You could use [Object.assign](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) here. Using Object.assign() or the spread operator is generally considered a more modern and preferable approach for object assignment, providing improved readability, immutability, compatibility, and handling of undefined or null values.

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
```
Anastasiia-Svintsova: Add case for the `clear` case. In the default case you can  throw the error that the type is not found, for example 
image

Anastasiia-Svintsova (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can destructure the action to get all the needed props of it (type, extraData, keysToRemove)

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return `type in array 'actions' is not found`;
```
polosanya, line 27: I think it's better to throw an error instead

polosanya (approved):
Nice!

PR #2506, opened 2023-06-20
Solution

YuriyBilskiy:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
loralevitska, line 8: Destructure values from `action`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      for (const key of keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
loralevitska: Use [`switch case`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) in cases where you constantly compare one value with others

loralevitska (changes requested):
Almost done) Let's improve

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
sTorba24, line 9: you can do it directly in loop like 

const { type, extraData, keysToRemove } of actions

sTorba24 (approved):
GJ 🚀

PR #2507, opened 2023-06-20
add tusk solution

MaksymBevskyi:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      const extraData = action.extraData;
+
+      for (const key in extraData) {
+        state[key] = extraData[key];
+      }
+    }
+
+    if (action.type === 'removeProperties') {
```
vlad-demchuk: better to use switch when you compare one value

vlad-demchuk (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2508, opened 2023-06-20
create a transformState function

Solovei-Andrii:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const changeState = state;
```
anastasiiavorobiova: You don't need the additional variable for this code

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        delete changeState[key];
+      }
+    } else if (el.type === 'clear') {
+      for (const key in changeState) {
+        delete changeState[key];
+      }
+    }
+  }
+
+  return changeState;
```
anastasiiavorobiova: You don't need the return statement. You have already changed the state object

anastasiiavorobiova (changes requested):
Great start! Added tiny comments

src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+      Object.assign(state, el.extraData);
+    } else if (el.type === 'removeProperties') {
+      for (const key of el.keysToRemove) {
+        delete state[key];
+      }
+    } else if (el.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
Viktor-Kost: try to avoid using else if statements, such it decreases readability. It'd be better to apply switch

Viktor-Kost (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2511, opened 2023-06-21
Solution

Constantin-Haranskiy:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya: You can destructure action to get all the needed props of action (type, extraData, keysToRemove)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'addProperties' :
+        Object.entries(action.extraData).forEach(([key, value]) => {
+          state[key] = value;
```
polosanya: Object.assign can be use here instead
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+
+      case 'removeProperties' :
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
```
Default case is missing

polosanya (changes requested):
Almost done!

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+
+    switch (type) {
+      case 'addProperties' :
+        Object.assign(state, extraData);
+        break;
+
+      case 'clear' :
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
```
loralevitska, line 19: Could be better with [`for ... in` loop](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in)

loralevitska (approved):
Well done)

PR #2514, opened 2023-06-23
Solution

Patryk-Filipiak:


domalewski (changes requested):
everything looks fine 
are you forgot point 4 on checklist?

Patryk-Filipiak:
> everything looks fine are you forgot point 4 on checklist?

default case added :)

domalewski (approved):
looks awesome

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
choeqq, line 20: please add empty line after break so it's easier to read

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```

choeqq (changes requested):

src/transformState.js
Patryk-Filipiak: Are you sure? The state object should be mutable, and the spread operator creates a copy of it without mutating it. I also cannot change the values of the state because parameters of a function cannot be reassigned, in my opinion its the best way to do

src/transformState.js
Patryk-Filipiak: done

src/transformState.js
JoyCoffeeAddict: That's correct, it's very nice that you provided an argument.

JoyCoffeeAddict:
@Erchu420 I think you've forgotten to push your changes

Patryk-Filipiak:
> @Erchu420 I think you've forgotten to push your changes

check now

choeqq (approved):
Good job 🚀

PR #2515, opened 2023-06-24
first commit

Faridun92:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      keysToRemove.forEach(key => {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      });
+    } else if (type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
etojeDenys: better to use here switch case instead of if`s

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,39 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
```
OleziO, line 12: Move all cases to the object. Example of usage: `case Actiontype.AddProperties:`

src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+
```
OleziO, line 36: Remove empty line

OleziO (changes requested):

SanyaBratashchuk (approved):

PR #2517, opened 2023-06-26
Done

SashaKozoriz:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let newState = state;
+
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      newState = Object.assign(newState, action.extraData);
```
etojeDenys: You don't need to use Object.assign, instead use the spread operator to merge the extraData into the state object.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    }
+
+    if (action.type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+    }
+  }
+
+  return newState;
```
etojeDenys: You don't need to return newState, as the function is supposed to modify the given state object directly.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let newState = state;
```
etojeDenys: Avoid reassigning the state object to newState, instead directly modify the state object.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        delete newState[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in newState) {
+        delete newState[key];
+      }
+    }
+  }
```
etojeDenys: better to use here switch\case instead of if`s it will improve readability

etojeDenys (changes requested):

VitaliyBondarenko1982 (approved):

PR #2518, opened 2023-06-26
solution

tndeineka:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VitaliyBodnar: - please use the switch case here instead of multiple 'if'

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (const key of action.keysToRemove) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
```
VitaliyBodnar: can be less
```suggestion
      for (const key of action.keysToRemove) {
        delete state[key];
      }
```

VitaliyBodnar (changes requested):
Good Job🔥, let’s make a few changes.

sTorba24 (approved):
GJ 🚀

PR #2521, opened 2023-06-26
added completed task

olya-girnyak:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya, line 9: You can make destructuring to get all the needed props of action

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 0;
```
polosanya, line 27: I think it is better to throw some error

polosanya (approved):
Nice!

PR #2522, opened 2023-06-27
Solution

Kosnstantin-Levonchuk:


src/transformState.js
```diff
@@ -1,11 +1,31 @@
 'use strict';
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
+// /**
```
SerhiiKirik, line 3: why do you comment on it?

SerhiiKirik (approved):
Well done 🚀

PR #2523, opened 2023-06-27
stateful object solution

vetal-hovenko:


src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: use `switch case` block (hint: it's a good practice to always add `default` case)

src/transformState.js
```diff
@@ -5,7 +5,19 @@
...
+      Object.assign(state, action.extraData);
+    } else if ((action.type === 'removeProperties')) {
+      for (const removedKey of action.keysToRemove) {
```
TarasHoliuk:
```suggestion
      for (const keyToRemove of action.keysToRemove) {
```
it's good practice to name such values in this way during mapping, schematically: `for (const singular of plural)`

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,30 @@
...
+        break;
+
+      case 'clear':
+        for (const clearing in state) {
+          delete state[clearing];
+        }
+        break;
+
+      default:
+        continue;
```
volodymyr-soltys97, line 28: The `continue` statement in the `default` case of the switch is not necessary as it does nothing here. It could be removed for cleaner code.

volodymyr-soltys97 (approved):
Nice work 🔥

volodymyr-soltys97 (approved):

PR #2525, opened 2023-06-30
add task solution

NikitaOrliak:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const value = Object.values(action);
```
mykhalenych: you can desctruct action

mykhalenych (changes requested):

mykhalenych (approved):

PR #2526, opened 2023-06-30
the transformState was completed

Viktoria-Moroz:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+          }
+        }
+        break;
+
+      default:
+        throw new Error('Unexpected action type');
+    }
+  }
+
+  return state;
```
Viktor-Kost: Returning the 'state' is unnecessary as objects in JavaScript are passed by reference. The function will modify the original state object without the need for a return statement.

Viktor-Kost (changes requested):
Nice!
Check one comment I left for you

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const act of actions) {
```
```suggestion
  for (const action of actions) {
    const { type,  keysToRemove, extraData } = action
```

loralevitska (approved):
Well done)

PR #2527, opened 2023-07-01
solution

Hladkyi03:
Solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -6,18 +6,24 @@
  */
 function transformState(state, actions) {
   for (const object of actions) {
-    if (object.type === 'addProperties') {
-      for (const addKey in object.extraData) {
-        state[addKey] = object.extraData[addKey];
-      }
-    } else if (object.type === 'removeProperties') {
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    } else if (object.type === 'removeProperties') {
+      for (const removeKey of object.keysToRemove) {
+        delete state[removeKey];
+      }
+    } else {
+      for (const stateKey in state) {
+        delete state[stateKey];
+      }
+    }
+  }
```
etojeDenys: better to rewrite it using switch\case instead of if\else

etojeDenys (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```
loralevitska, line 8: Destructure `object`

loralevitska (approved):
Well done)

PR #2528, opened 2023-07-02
solution

monikaprochniewicz:


DorotaLeniecDev (changes requested):
Refactor your solution using switch

src/transformState.js
```diff
@@ -6,6 +6,33 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    const { type } = action;
```
DorotaLeniecDev: You could destructure all needed variables at once
```suggestion
     const { type, extraData, keysToRemove } = action;
```

src/transformState.js
```diff
@@ -6,6 +6,33 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
DorotaLeniecDev: For readability you could add spaces between cases 
```suggestion
      case 'addProperties':
        const { extraData } = action;

        Object.assign(state, extraData);
        break;
        
      case 'removeProperties':
        const { keysToRemove } = action;

        for (const key of keysToRemove) {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        }
        break;
        
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
      default:
        break;
```

DorotaLeniecDev (changes requested):
Good job, 2 small suggestion for improvements from my side

darokrk (approved):
Good job 😃  looks nice, approved 🏆

PR #2531, opened 2023-07-03
made transform state

ivan-baranovskyi:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      }
+    } else if (objActions.type === REMOVE) {
+      for (const key of objActions.keysToRemove) {
+        delete state[key];
+      }
+    } else if (objActions.type === REMOVE_ALL) {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
sTorba24: use `switch ... case` instead

sTorba24 (changes requested):
Almost good! Lets improve

sTorba24 (approved):
Much better! Approved :)

PR #2533, opened 2023-07-04
my solution

EkaterinaBessonnaya:
:wink:

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -4,8 +4,31 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -4,8 +4,31 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const keys of action.keysToRemove) {
+          delete state[keys];
+        }
+        break;
```
agniya-i, line 27: Don't forget that `switch/case` should always  **default** case for error handling 🙌

agniya-i (approved):

PR #2537, opened 2023-07-04
add stateful_object solution

Serhii-Dr:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete state[key]);
+        break;
+      case 'clear':
+        if (Object.keys(state).length > 0) {
```
Viktor-Kost: Checking the length of Object.keys(state) is redundant. The forEach loop will not execute if there are no keys, so this check can be removed for cleaner code.

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          Object.keys(state).forEach(key => delete state[key]);
+        }
+        break;
+      default:
+        return state;
```
Viktor-Kost, line 20: In the 'default' case of switch, it is not necessary to return 'state'. Switch statement doesn't require a default return as function will return 'state' at the end regardless.

Viktor-Kost (changes requested):

src/transformState.js
VitaliyBondarenko1982: https://github.com/mate-academy/js_stateful-object/pull/2537/files#r1252276239
not fixed this comment. it means that no need to return state. better to throw some error for example.

VitaliyBondarenko1982 (approved):
Looks good

src/transformState.js
Serhii-Dr: I think i'm not 100% understood this issue
should I just in 'default' case to throw 'new Error' object and not to return state both from switch statement and function at all?

PR #2543, opened 2023-07-06
Solution

kyslytsyn:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97, line 9: You can use desctructurization here for action
`const { type, extraData, keysToRemove } = action`

volodymyr-soltys97 (approved):
Good work 🔥

PR #2545, opened 2023-07-06
Solution

nenatre:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (const level of actions) {
```

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+        for (const key of level.keysToRemove) {
+          delete state[key];
+        }
 
+        break;
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
```
polosanya, line 24: image

polosanya (changes requested):
Almost done!

src/transformState.js
polosanya: Not fixed. Default case is still absent

src/transformState.js
```diff
@@ -5,7 +5,22 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
polosanya:
```suggestion
  for (const action of actions) {
    switch (action.type) {
```
You can make destructuring to get all the needed props of action 
https://javascript.info/destructuring-assignment#object-destructuring

polosanya (changes requested):
Please make sure that you fixed all comments. 
Feel free to ask for help in js_chat if you have any difficulties)

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+      default:
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
```
qoa11a:
```suggestion
      case 'clear':
        for (const key of Object.keys(state)) {
          delete state[key];
        }
        break;
      default:
        throw new Error('Unrecognised action.type');
```
While the `default` case should exist it doesn't mean that you should move the functionality of a specific case under this case.

qoa11a (changes requested):

polosanya (approved):
Well done!

PR #2546, opened 2023-07-07
task is done

Victoriia-Melnyk:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+    if (object.type === 'removeProperties') {
+      for (const word of object.keysToRemove) {
+        delete state[word];
+      }
+    }
+
+    if (object.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
```
OleziO: Replace with switch case construction

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const object = actions[i];
+
+    if (object.type === 'addProperties') {
```

Example of usage: `object.type === ActionTypes.AddProperties`

OleziO (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        };
```
```suggestion
       Object.assign(state, action.extraData)
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
SanyaBratashchuk, line 8: use object destructuring

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        };
+        break;
+      case 'removeProperties':
+        for (const word of action.keysToRemove) {
```
```suggestion
        for (const keyToRemove of action.keysToRemove) {
```

SanyaBratashchuk (changes requested):

TarasHoliuk (approved):

PR #2552, opened 2023-07-11
complete task

Horizon-git:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97: Use desctrurization here 
`const {} = action`
Consider it will be better if you create a copy of initial state

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Awesome

PR #2556, opened 2023-07-12
solution

julius0724:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: use `switch`
P.S. don't forget about `default` case

TarasHoliuk: Also, you may destruct object to get all properties you use below

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

demetrxx (approved):
Well done 👍 Go ahead!

PR #2557, opened 2023-07-12
Complite the task

AndruhaMan:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

anastasiiavorobiova (approved):
Awesome!

PR #2558, opened 2023-07-13
Solution

oksom:


src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
+      case 'addProperties':
```
Anastasiia-Svintsova, line 10: It's better to use constants for string values to avoid errors because of typo

Anastasiia-Svintsova (approved):

PR #2559, opened 2023-07-13
Stateful object

Kacper-Leman:
Soultion

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    } else if (action.type === 'removeProperties') {
+      action.keysToRemove.forEach((key) => {
+        delete state[key];
+      });
+    } else if (action.type === 'clear') {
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,19 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    if (action.type === 'addProperties') {
```
Radoslaw-Czerniawski, line 9: Generally avoid else if, else if, else if conditional chaining, [switch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch) would be better here. 

For example you want to execute the last condition for 'clear', javascript will go and check conditions one by one before finally executing the last, correct one. If there were more this could take a while.

In the switch statement on the other hand, compiler creates a jump table and immediately "jumps" into the right clause rather than checking conditions / clauses one by one.

Radoslaw-Czerniawski (approved):
Very good execution :D 
My suggestion is to in future prs, favor `switch` statement over chaining `else if` conditions.

src/transformState.js
Kacper-Leman: Yea that's true I will use switch next time :D

PR #2561, opened 2023-07-15
Solution

RomanenkoStud:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
volodymyr-soltys97, line 16: The code assumes that the second property for 'removeProperties' is 'keysToRemove'. However, as per the task description, it's not explicitly mentioned what the second property is.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return Error('Bad action type');
```
volodymyr-soltys97: Returning an Error object won't actually throw an error. Consider throwing the error instead, like 'throw new Error('Bad action type')'.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
volodymyr-soltys97: You can use destructurization here for action 
```suggestion
  for (const action of actions) {
    const {} = action;
      switch (action.type) {
```

volodymyr-soltys97 (changes requested):
Good job

nazarmatsevych (approved):

PR #2562, opened 2023-07-15
stateful-object_solution

olha-rypich:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

OleziO (changes requested):
Fix all comments

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const actionTypes = {
```
mykhalenych, line 8: better to create static constants outside the function

mykhalenych (approved):

PR #2563, opened 2023-07-15
Solution

SichVictor:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

VitaliyBondarenko1982 (approved):

PR #2566, opened 2023-07-17
Solution

oksana-vas:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+const actionTypes = {
+  addProperties: 'addProperties',
+  removeProperties: 'removeProperties',
+  clear: 'clear',
+};
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
mykhalenych, line 15: also you can destruct action

const {type, extraData, keysToRemove} = action

mykhalenych (approved):

PR #2567, opened 2023-07-17
Solution

oshapkun:


src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+          state[key] = extraData[key];
+        }
+        break;
+      }
+
+      case 'removeProperties': {
+        const { keysToRemove } = action;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
anastasiiavorobiova: Checking if the property exists before deleting it is redundant. The delete operator does not throw an error if the property does not exist.

src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+      }
+
+      case 'clear': {
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      }
+
+      default: {
```
anastasiiavorobiova, line 39: Throwing an error for an unknown action type could stop the execution of the function prematurely. If this behavior is not intended, consider logging a warning or ignoring the unknown type instead.

anastasiiavorobiova (changes requested):
Great work! Added tiny comments to fix

nazarmatsevych (approved):

PR #2568, opened 2023-07-17
add task solution

1umamaster:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const member of actions) {
```
Esceype: If we have an array of actions, then the member is somehow not very suitable for me in this situation

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    switch (member.type) {
+      case 'addProperties':
+        for (const key in member.extraData) {
+          state[key] = member.extraData[key];
+        } break;
+
+      case 'removeProperties':
+        for (const item of member.keysToRemove) {
+          delete state[item];
+        } break;
```
Esceype: New line for break

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const member of actions) {
+    switch (member.type) {
+      case 'addProperties':
+        for (const key in member.extraData) {
+          state[key] = member.extraData[key];
+        } break;
```
Esceype: What about Object.assign ?)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+      case 'removeProperties':
+        for (const item of member.keysToRemove) {
+          delete state[item];
+        } break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        } break;
```
Esceype: Why do you need another additional loop here, if you need to completely clean the object?

Esceype (changes requested):
GJ! But some small fixes to improve your code style )

Esceype (commented):
Great job, fast fix, cool )

Esceype (approved):
Great job, fast fix, cool )

PR #2569, opened 2023-07-17
task done

pti4ka121:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
DarkMistyRoom, line 9: would be good practice to decompose action object here
`const { type, extraData, ... } = action;`

DarkMistyRoom (commented):
Good job!

DarkMistyRoom (approved):

PR #2571, opened 2023-07-18
Solution added

kyrylomanko:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => delete state[key]);
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
```
polosanya, line 20: Let's add a default case to avoid possible errors
image

polosanya (changes requested):
Almost done!

lerastarynets (approved):
You've got this, keep pushing forward!

PR #2572, opened 2023-07-18
Solution

odolga:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const valueOfType = action['type'];
```
volodymyr-soltys97: The name 'valueOfType' could be more descriptive. Consider renaming it to 'actionType' to better represent its content.

volodymyr-soltys97: You can use a desctructurization here for action
```suggestion
  for (const action of actions) {
    const {} = action;
```

volodymyr-soltys97 (changes requested):
Good job

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        (Object.assign(state, action['extraData']));
```
```suggestion
        (Object.assign(state, action.extraData));
```
There is no need to use dynamic keys. Consider using static keys or using destructuring

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+
+      case 'removeProperties':
+        for (const keyToDelete of action['keysToRemove']) {
```
anastasiiavorobiova: See my previous comment

anastasiiavorobiova (changes requested):
Great work! Keep doing

anastasiiavorobiova (approved):
Great work

PR #2573, opened 2023-07-18
all tests pass

aniashev:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const actionType = actions[i].type;
+
+    if (actionType === 'addProperties') {
```
mykhalenych: change if's to switch/case

mykhalenych (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
Viktor-Kost: you can simplify using `for of`, so you'll not need to get value by index

Viktor-Kost (changes requested):

src/transformState.js
aniashev: thank you for your suggestion! makes sense

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
+    const actionType = object.type;
```
polosanya:
```suggestion
    const { type, extraData, keysToRemove } = object;
```
You can just get all the needed props using destructuring

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const object of actions) {
```

polosanya (changes requested):
Almost done!

Viktor-Kost (approved):

PR #2574, opened 2023-07-18
Develop

Sancz0pansa:


domalewski (approved):
looks awesome

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (let j = 0; j < action.keysToRemove.length; j++) {
+          delete state[action.keysToRemove[j]];
+        }
```
choeqq: please use `for of` loop or `forEach` in that case

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
```
choeqq, line 12: cosider adding empty line after each break statement

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      default:
+        throw new Error('Something went wrong !');
+    }
+  }
 }
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```

choeqq (changes requested):

src/transformState.js
Sancz0pansa: using normal for loop with index is something bad  ?

src/transformState.js
Sancz0pansa: We cant modify function parameters so:

const addons = action.extraData;

state = {...state, ...addons};

won't work.

mvjl000 (approved):

src/transformState.js
JoyCoffeeAddict: @Sancz0pansa it's not bad, but it's more readable to use `for of` if you want to get each value anyway. 

e.g.
```js 
const example = [1, 2, 3];

for (let i =0; i < example.length; i++) {
  const value = example[i]
}
```

vs


```js 
const example = [1, 2, 3];

for (const value of example) {
  // I have a value and the code is readable
}
```

PR #2575, opened 2023-07-18
Added transformState

PavloSich:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
```
loralevitska: Destructure `action`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
```
loralevitska: Add empty line between cases
```suggestion
        break;
        
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        }
        break;
        
      case 'clear':
        for (const key in state) {
          if (state.hasOwnProperty(key)) {
            delete state[key];
          }
        }
        break;
        
```

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
loralevitska: This check redundant this key will be because you loop through it from this object
```suggestion
        for (const key in state) {
          delete state[key];
        }
```

loralevitska (changes requested):
Almost perfect) Let's improve

sTorba24 (approved):
Nice work!

sTorba24 (approved):
GJ 🚀

PR #2576, opened 2023-07-19
added solution

FurArt:
Hi,
but I feel like I'm using EVIL

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -7,7 +7,7 @@
 
 function transformState(state, actions) {
   actions.forEach((action) => {
-    const { type } = action;
+    const { type, extraData, keysToRemove } = action;
 
     switch (type) {
       case 'clear':
@@ -19,11 +19,11 @@ function transformState(state, actions) {
 
         break;
       case 'addProperties':
-        Object.assign(state, action.extraData);
+        Object.assign(state, extraData);
 
         break;
       default:
-        Object.values(action.keysToRemove).forEach((removeProperties) => {
+        Object.values(keysToRemove).forEach((removeProperties) => {
           if (removeProperties in state) {
             delete state[removeProperties];
           }
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -7,7 +7,9 @@
 
 function transformState(state, actions) {
   actions.forEach((action) => {
-    switch (action.type) {
+    const { type } = action;
+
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+        Object.values(action.keysToRemove).forEach((removeProperties) => {
+          if (removeProperties in state) {
+            delete state[removeProperties];
+          }
+        });
+        break;
+    }
+  });
+
+  return state;
```
volodymyr-soltys97: Since JavaScript is pass by reference for objects, you don't need to return the state. Any modifications made to the state object will automatically affect the original state.

src/transformState.js
```diff
@@ -4,8 +4,33 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
```
volodymyr-soltys97: You can use desctructurization here for action
```suggestion
  actions.forEach((action) => {
    const { type, ...} = action;
    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+        break;
+      default:
+        Object.values(action.keysToRemove).forEach((removeProperties) => {
+          if (removeProperties in state) {
+            delete state[removeProperties];
+          }
+        });
+        break;
+    }
+  });
```
volodymyr-soltys97:
```suggestion
  actions.forEach((action) => {
    const { type, extraData, keysToRemove } = action;

    switch (type) {
      case 'clear':
        for (const stateProper in state) {
          if (state.hasOwnProperty(stateProper)) {
            delete state[stateProper];
          }
        }

        break;
      case 'addProperties':
        Object.assign(state, extraData);

        break;
      default:
        Object.values(keysToRemove).forEach((removeProperties) => {
          if (removeProperties in state) {
            delete state[removeProperties];
          }
        });
        break;
    }
  });
```

volodymyr-soltys97 (changes requested):
Let's improve your code

anastasiiavorobiova (approved):
Awesome!

PR #2580, opened 2023-07-20
function transformState

MartaChobaniuk:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
Viktor-Kost, line 26: it is a good practice when you add a default behavior for switch statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

Viktor-Kost (approved):
Nice!
Just apply my recommendation for yourself

PR #2584, opened 2023-07-21
Develop

AndrzejPec:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete state[key]);
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
+        break;
+
+      default:
+        window.alert('Unknown action type: ' + action.type);
```
Meg-Sowka: It would be better to `throw Error` here, as it allows us to handle it properly 😉

Meg-Sowka:
Some changes needed in my opinion ;)

Meg-Sowka (changes requested):
It's almost perfect ;) Some changes needed!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+    }
+  });
 }
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq, line 11: this can also be achieved with [spread operator ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is more commong nowadays

choeqq (changes requested):

src/transformState.js
AndrzejPec: unfortunately spread operator is difficult to use since linter prohibits reasigning state (as function input) and coach used this Object.assign method in today's lesson (live coding) aswell, so I decided to leave it that way

choeqq (approved):
Good job 🚀

PR #2585, opened 2023-07-21
solution

ogospadzipalina:


Vira-V (approved):
This solution cover all tests. I guess that everything is correct. Good job!

src/transformState.js
```diff
@@ -5,7 +5,26 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```

choeqq (approved):

PR #2587, opened 2023-07-23
add task Object

VladyslavKolisnyk:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const prop of action.keysToRemove) {
+          if (state[prop]) {
```
anastasiiavorobiova: The check if (state[prop]) is redundant. The delete operation is safe and does nothing if the property doesn't exist in the object.

anastasiiavorobiova (changes requested):
Almost done! Added a tiny comment to fix

anastasiiavorobiova (approved):
Awesome!

PR #2588, opened 2023-07-23
solution

cardinal1991:


Meg-Sowka (changes requested):
Almost perfect, few changes needed ;)

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        // Ignoring unknown action types
```
choeqq, line 30: you should throw Error here

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+        break;
+    }
+  }
 }
```

src/transformState.js
```diff
@@ -4,8 +4,33 @@
...
+function transformState(states, actions) {
+  const state = states;
+
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        if (action.extraData && typeof action.extraData === 'object') {
+          Object.assign(state, action.extraData);
+        }
+        break;
```
choeqq, line 16: please add empty line after each break statement

src/transformState.js
```diff
@@ -4,8 +4,33 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
-function transformState(state, actions) {
-  // write code here
+function transformState(states, actions) {
+  const state = states;
```
choeqq: I don't get it 🤔 
We're prompted to modify state, which is rarely the case but here we are 😄

choeqq (changes requested):

choeqq (approved):
Well done ✅

PR #2589, opened 2023-07-24
task has been done

vladkorobka:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```

Example of usage:

OleziO (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+
+  actions.map(action => {
+    ActionTypes[action.type] = action.type;
+  });
+
+  for (const action of actions) {
+    switch (action.type) {
+      case ActionTypes.addProperties:
+        for (const data in action.extraData) {
+          state[data] = action.extraData[data];
```
anastasiiavorobiova: Consider using Object.assign instead of the loop for better performance

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ActionTypes = {};
+
+  actions.map(action => {
+    ActionTypes[action.type] = action.type;
```
anastasiiavorobiova: This code seems redundant, consider deleting it. And using the ActionTypes object

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ActionTypes = {};
```
```suggestion
  const ActionTypes = {
    removeProperties: "removeProperties",
    ...etc
  };
```

anastasiiavorobiova (changes requested):
Great work! Almost done!

Viktor-Kost (approved):

PR #2590, opened 2023-07-24
cr a func

OPTIMISTIXX:


src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
Anastasiia-Svintsova: The default case in the switch statement is a no-op and can be removed.

src/transformState.js
```diff
@@ -5,7 +5,42 @@
...
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        if ('extraData' in action) {
+          const { extraData } = action;
+
+          for (const key in extraData) {
+            state[key] = extraData[key];
+          }
```
Anastasiia-Svintsova: Use the Object.assign() method instead

src/transformState.js
```diff
@@ -5,7 +5,42 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
```
Anastasiia-Svintsova: Would be great to create variables for string values to avoid errors due to typos.

Anastasiia-Svintsova (changes requested):

anastasiiavorobiova (approved):
Great work!

PR #2591, opened 2023-07-24
solution

Capoverde:
My solution

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        if (action.extraData && typeof action.extraData === 'object') {
+          Object.assign(state, action.extraData);
+        }
+        break;
+      case 'removeProperties':
+        if (action.keysToRemove && Array.isArray(action.keysToRemove)) {
+          action.keysToRemove.forEach((key) => {
+            if (state.hasOwnProperty(key)) {
+              delete state[key];
+            }
+          });
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      default:
+        break;
+    }
+  });
 }
 
 module.exports = transformState;
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -1,8 +1,7 @@
 'use strict';
 
 const REMOVE_PROPERTIES = 'removeProperties';
-const ACTION_AND_PROPERTIES = 'addProperties';
-const CLEAR = 'clear';
+const ACTION_ADD_PROPERTIES = 'addProperties';
 
 /**
  * @param {Object} state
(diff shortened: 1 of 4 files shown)


src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+      default:
```
domalewski, line 28: default should return error, right?

src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    switch (action.type) {
+      case 'addProperties':
+        if (action.extraData && typeof action.extraData === 'object') {
```
domalewski: this condition is necessary?

domalewski (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+    if (action.type === ACTION_AND_PROPERTIES) {
+      Object.assign(state, action.extraData);
+    } else if (action.type === REMOVE_PROPERTIES) {
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    } else if (action.type === CLEAR) {
+      Object.keys(state).forEach(key => {
+        delete state[key];
+      });
```
choeqq: according to [checklist ](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)you should use switch statement

src/transformState.js
```diff
@@ -1,11 +1,32 @@
 'use strict';
 
+const REMOVE_PROPERTIES = 'removeProperties';
+const ACTION_AND_PROPERTIES = 'addProperties';
+const CLEAR = 'clear';
```
choeqq: nice usage of constants

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+      action.keysToRemove.forEach(key => {
+        delete state[key];
+      });
+    } else if (action.type === CLEAR) {
+      Object.keys(state).forEach(key => {
+        delete state[key];
+      });
+    }
+  });
 }
```

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+const ACTION_AND_PROPERTIES = 'addProperties';
+const CLEAR = 'clear';
+
 /**
  * @param {Object} state
- * @param {Object[]} actions
+ * @typedef {Object} Action
+ * @property {string} type
+ * @property {Object} extraData
+ * @property {strinmg[]} keysToRemove
```
choeqq, line 12:
```suggestion
 * @property {string[]} keysToRemove
```

choeqq (changes requested):

src/transformState.js
```diff
@@ -1,11 +1,32 @@
...
+  for (const key of actions) {
+    switch (key.type) {
+      case ACTION_ADD_PROPERTIES:
+        Object.assign(state, key.extraData);
+        break;
+      case REMOVE_PROPERTIES:
+        key.keysToRemove.forEach(item => delete state[item]);
+        break;
+      default:
+        Object.keys(state).forEach(item => delete state[item]);
```
choeqq: you're missing `clear` case - default case is supposed to throw new Error

choeqq (changes requested):
1 change needed, and you're good to go 🚀

Capoverde:
Looking for feedback from You :)

src/transformState.js
Capoverde: Looking for feedback from You :)

choeqq (approved):
Nicely done ✅

PR #2592, opened 2023-07-24
task solution

HubertKlamann:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    if (obj.type === 'removeProperties') {
+      for (const key of obj.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
```
Matylda-Lewandowska: I think it is better using here [switch](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)
don't forget about 4th point of [chacklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) default should have an error! :)

Matylda-Lewandowska (changes requested):
Hi, this code you had made need little adjustments, but all over is alright. Great job!  :)

src/transformState.js
choeqq: according to [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) you should use switch

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+    }
```
choeqq, line 28: please provide default handler for values that aren't within scope of our functions

choeqq (changes requested):

choeqq (approved):
Good work ✅

PR #2595, opened 2023-07-24
solution

vadymboichenkoAcupower:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
Anastasiia-Svintsova: Use switch/case instead to make the code more organized, and readable. And don't forget about the default case.

Anastasiia-Svintsova: Would be great to move all string values to variables. It's a good practice to avoid errors connected with typos in the future.

Anastasiia-Svintsova (changes requested):

volodymyr-soltys97 (approved):
Nice work 👍

PR #2597, opened 2023-07-25
initial commit

domalewski:


PawelP2001 (approved):
looks good, matches the checklist

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        throw Error('Something went wrong');
+    }
+  }
 }
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq, line 11: that can also be achieved using [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is way more common nowadays

choeqq (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw Error('Something went wrong');
```
JoyCoffeeAddict: I wouldn't throw an error here, because this way you expect that every consumer of this methods handles Errors. 
You could just: 

```suggestion
        console.error(`Error: ${action.type} is incorrect action type`);
```
We typically throw in `async` code, that will be taught later in the course. 

It's not inherently bad to throw, but there is no need for it here.
Also, `throwing` / `console.error` are not for the user, but for other programmers (or you) you should describe the problem that happened (In this case, bad action.type was provided)

JoyCoffeeAddict (approved):

src/transformState.js
domalewski: ESLint disallow to use of console, but you have a great point

PR #2598, opened 2023-07-25
Solution

Dolnys:


domalewski (approved):
everything looks good

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq: that can also be achieved with `{... , variable}` - [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax), which is more popular this days

choeqq: it would be useful to check if `extraData` is object, we don't want o assign string or anything else

choeqq (changes requested):

Dolnys:
I introduced data type checking. However, I kept the Object.assign method because when I tried to solve the problem using the spread operator, the tests did not pass

JoyCoffeeAddict:
Spread operator is more common for similar problem, but this task forces us into Object.assign

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          }
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
```
JoyCoffeeAddict: You could add an empty line after each `break`

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+      default:
+        throw new Error(`Error: ${action.type}`);
```
JoyCoffeeAddict: I wouldn't throw an error here, because this way you expect that every consumer of this methods handles Errors. 
You could just: 

```suggestion
        console.error(`Error: ${action.type} is incorrect action type`);
```
We typically throw in `async` code, that will be taught later in the course. 

It's not inherently bad to throw, but there is no need for it here.

JoyCoffeeAddict (approved):

Dolnys:
I changed throw error to console.log, but I had to add /* eslint-disable no-console */ for ESLint because the use of console logs was disabled in the project. I hope this is okay.

JoyCoffeeAddict:
@Dolnys By the way: for displaying errors you can use `console.error` or `console.warn`. And it's okay to disable this rule, but only in this place.  You should use `// eslint-disable-next-line no-console` to be 100% correct. The difference is , that in most cases you want to disable a rule only in a specified place.

Dolnys:
> @Dolnys By the way: for displaying errors you can use `console.error` or `console.warn`. And it's okay to disable this rule, but only in this place. You should use `// eslint-disable-next-line no-console` to be 100% correct. The difference is , that in most cases you want to disable a rule only in a specified place.

Okay, thanks, I'll remember that. 👍

PR #2599, opened 2023-07-25
Solution to "Transform state"

tniklevych:


src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const operation = action.type;
+
+    switch (operation) {
```
volodymyr-soltys97: Use a destructurization here for action
```suggestion
  for (const action of actions) {
    const {} = action;

    switch (type) {
```

volodymyr-soltys97 (changes requested):
Good job 👍 

anastasiiavorobiova (approved):
Great work!

PR #2600, opened 2023-07-25
Solution

klaudia-klis:


domalewski (approved):
looks good to me

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq, line 11: consider adding empty line after each break statement, it's more readable that way

choeqq (approved):

ihor-jpeg (approved):

PR #2601, opened 2023-07-25
add state transform function

Meg-Sowka:


domalewski (changes requested):
everything looks nice
but you forgot point 4 on checklist

Meg-Sowka:
> everything looks nice but you forgot point 4 on checklist

@domalewski Thanks a bunch! Already fixed it!

domalewski (approved):
looks awesome, nice job

src/transformState.js
```diff
@@ -4,8 +4,30 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq, line 12: this can also be achieved using [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is more common nowadays

src/transformState.js
```diff
@@ -4,8 +4,30 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Action not found!');
+    }
+  }
 }
```

choeqq (changes requested):

src/transformState.js
Meg-Sowka: @choeqq In this particular example I would have to reassign value to the state
state = { ...state, ...action.extraData };
which I cannot do with function parameters, so Object.assign is the best option here.

and the task itself asks us to mutate the current state.

mvjl000 (approved):

PR #2603, opened 2023-07-25
Solution

PawelP2001:


domalewski (approved):
looks good

src/transformState.js
```diff
@@ -5,7 +5,38 @@
...
+    const { type } = action;
+
+    switch (action.type) {
+      case 'addProperties':
+        const { extraData } = action;
+
+        for (const [key, value] of Object.entries(extraData)) {
+          state[key] = value;
+        }
+        break;
```
choeqq: you can also achieve that using just[ spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) or `Object.assign()` , that way there will be no need to creating the loop

src/transformState.js
```diff
@@ -5,7 +5,38 @@
...
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error(`Invalid action: ${type}`);
+    }
+  }
 }
```

choeqq (changes requested):

choeqq (approved):
Good job 🚀

PR #2604, opened 2023-07-25
Solution

Reject-4444:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (true) {
```
denys-danyliuk: try to come up with better condition than just leaving true
I'll give you a hint, in each case you compare action.type to something, why would you do that if whole idea of switch is to compare condition to case

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+
+      case action.type === 'removeProperties':
+        for (const keys of action.keysToRemove) {
+          delete state[keys];
+        };
+        break;
+
+      default:
+        Object.keys(state).forEach(n => delete state[n]);
+        break;
```
denys-danyliuk: this should be another case, not default one

denys-danyliuk (changes requested):

xiimiik (approved):
Nice work!

denys-danyliuk (approved):

xiimiik (approved):

PR #2605, opened 2023-07-25
Solution

Vira-V:


Meg-Sowka (approved):
Everything looks great! :)

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
choeqq, line 16: this can also be achieved using [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is more common nowadays

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        throw new Error(`Unknown action type: ${action.type}`);
+    }
+  }
 }
```

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
```
choeqq, line 17: consider adding empty line after each break statement

choeqq (changes requested):

src/transformState.js
Vira-V: I'm not sure about spread operator in this case. I tried, but tests were failed. So, I left the option with Object.assign().

JoyCoffeeAddict: That is correct solution in this case, spread operator is great on it's own, but the task forces us to think differently

JoyCoffeeAddict (approved):

PR #2606, opened 2023-07-25
Develop

Matylda-Lewandowska:
💮

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(element => delete state[element]);
+        break;
+
+      case 'clear':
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'removeProperties':
+        action.keysToRemove.forEach(element => delete state[element]);
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
choeqq, line 22: please add new line after break - it will be more readable

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```

choeqq (changes requested):

src/transformState.js
Matylda-Lewandowska: Are you sure, that I can use spread operator here? We need to modify state.

choeqq (approved):
Good job 🚀

ihor-jpeg (approved):

PR #2608, opened 2023-07-25
add task solution

arturwisnia:


src/transformState.js
mvjl000, line 1: Checklist point 2 and 4 - it would be better if you used switch here, instead of bunch of if statements

mvjl000: You could also return the state at the bottom of the function ;)

mvjl000 (changes requested):

choeqq (approved):
Good job 🚀

PR #2609, opened 2023-07-26
adding a solution

vensecue:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
-  // write code here
-}
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
 
+        Object.assign(state, extraData);
```
choeqq: this can also be achieved using [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is more common nowadays

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+    }
+  }
+}
```

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
-}
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
 
+        Object.assign(state, extraData);
+        break;
```
choeqq, line 12: consider adding empty line after each break statement - it is easier to read

choeqq (changes requested):

src/transformState.js
Meg-Sowka: @choeqq In this particular example I would have to reassign value to the state
`state = { ...state, ...action.extraData };`
which I cannot do with function parameter, so Object.assign is the best option here.

Meg-Sowka: @choeqq nvm clicked on the wrong link xD 
but the case is the same ;)

src/transformState.js
vensecue: I have checked the code again, also took a look at other students ones, now I understand what was wrong. Hopefully it matches everything now

choeqq (approved):
Good job 🚀

PR #2610, opened 2023-07-26
solution

patryk177m:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
 function transformState(state, actions) {
-  // write code here
+
+  for (const action of actions) {
+
+    switch (action.type) {
+      case 'addProperties':
+
+         Object.assign(state, action.extraData);
+        break;
```
choeqq: indentations look little odd - are you using prettier?

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+
+  for (const action of actions) {
+
+    switch (action.type) {
+      case 'addProperties':
+
+         Object.assign(state, action.extraData);
```
choeqq: this can also be achieved using [spread operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax) which is more common nowadays

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
+  }
+
 }
```

choeqq (changes requested):

choeqq (approved):
Good job 🚀

PR #2611, opened 2023-07-26
solution

SviatoslavVasyletskyi:


domalewski (changes requested):
code looks great just minor changes and you are good to go
for readability add some spaces between case's and so on
default should return error (check point 4 on checklist)
function should return state

src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        if (action.extraData && typeof action.extraData === 'object') {
+          for (const [key, value] of Object.entries(action.extraData)) {
+            state[key] = value;
```
choeqq, line 13: Object.assign() would useful there

choeqq (approved):

PR #2612, opened 2023-07-26
Solutions

SergiyVS:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        }
+        break;
+
+      case 'addProperties':
+        for (const extraKey in action.extraData) {
+          state[extraKey] = action.extraData[extraKey];
+        }
```
Anastasiia-Svintsova: Use [Object.assign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) instead

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+        break;
+
+      default: ;
```
Anastasiia-Svintsova: Add some action for a default case 
image

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
Anastasiia-Svintsova: Use destructuring for the action object

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'clear':
```
Anastasiia-Svintsova: It's a good practice to move string values to variables to avoid errors connected with typos and to promote code maintainability and readability.

Anastasiia-Svintsova (changes requested):

Anastasiia-Svintsova (changes requested):
Good start! Check my comments above

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const actionClear = 'clear';
+  const actionRemProp = 'removeProperties';
```
agniya-i:
```suggestion
  const actionRemoveProp = 'removeProperties';
```

agniya-i (approved):

PR #2613, opened 2023-07-26
Add state object

ghost:


Meg-Sowka (approved):
It's perfect! Good job! :)

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        }
+        break;
+      }
+
+      default: {
+        throw new Error("Incorrect action type");
+      }
+    }
+  }
 }
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    switch (action.type) {
+      case "addProperties": {
+        Object.assign(state, action.extraData);
+        break;
+      }
+
+      case "removeProperties": {
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
```
choeqq, line 18: it would be nice to check if property we want to delete exists

choeqq (changes requested):

choeqq (approved):
Good job 🚀

PR #2615, opened 2023-07-26
add task solution

griveous199:


src/transformState.js
```diff
@@ -2,10 +2,31 @@
...
+      Object.assign(state, action.extraData);
+    } else if (action.type === REMOVE_PROPERTIES) {
+      action.keysToRemove.forEach((key) => {
+        delete state[key];
+      });
+    } else if (action.type === CLEAR) {
+      Object.keys(state).forEach((key) => {
+        delete state[key];
+      });
+    }
```
choeqq: according to [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) you should use switch statement

src/transformState.js
```diff
@@ -2,10 +2,31 @@
...
+const CLEAR = 'clear';
+
 function transformState(state, actions) {
-  // write code here
+  actions.forEach((action) => {
+    if (action.type === ADD_PROPERTIES) {
+      Object.assign(state, action.extraData);
+    } else if (action.type === REMOVE_PROPERTIES) {
+      action.keysToRemove.forEach((key) => {
+        delete state[key];
```
choeqq: I would be nice to check if the object we want to delete exists

src/transformState.js
```diff
@@ -2,10 +2,31 @@
...
+      });
+    } else if (action.type === CLEAR) {
+      Object.keys(state).forEach((key) => {
+        delete state[key];
+      });
+    }
+  });
 }
```

src/transformState.js
```diff
@@ -2,10 +2,31 @@
...
+    } else if (action.type === REMOVE_PROPERTIES) {
+      action.keysToRemove.forEach((key) => {
+        delete state[key];
+      });
+    } else if (action.type === CLEAR) {
+      Object.keys(state).forEach((key) => {
+        delete state[key];
+      });
+    }
+  });
```
choeqq, line 37: please provide some default case - when incorrect method is provided

choeqq (changes requested):

src/transformState.js
```diff
@@ -2,10 +2,40 @@
...
+            delete state[key];
+          }
+        });
+        break;
+      case CLEAR:
+        Object.keys(state).forEach((key) => {
+          delete state[key];
+        });
+        break;
+      default:
```
choeqq, line 34: default should throw new Error when reached

choeqq (changes requested):
1 fix and you're good to go 🚀

src/transformState.js
Nightycake: Yeah but lint don't letting me use console.log or console.warn here

griveous199: nevermind I will update this in a sec

choeqq (approved):
Great job ✅

PR #2616, opened 2023-07-26
Solution

Val-Bryks:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const data in action.extraData) {
```
anastasiiavorobiova: Consider using Object.assign instead of the loop for better performance

anastasiiavorobiova (changes requested):
Great work! Almost done!

anastasiiavorobiova (approved):
Awesome!

PR #2617, opened 2023-07-26
add task solution

MateuszSlezak:


Meg-Sowka (changes requested):
Looks like it works nicely, BUT you should have taken a look at the checklist before approaching it ;)

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const stateHistory = [];
+
+  for (const order of actions) {
+    const ADD_PROPERTIES = order.type === 'addProperties';
+    const REMOVE_PROPERTIES = order.type === 'removeProperties';
+    const CLEAR_ALL = order.type === 'clear';
```
choeqq: those should be constants outside of function - not depending on order.type

src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+    }
+
+    if (CLEAR_ALL) {
+      for (const key in state) {
+        delete state[key];
+      }
+
+      stateHistory.push(Object.assign({}, state));
+    }
+  }
```
choeqq: according to [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) you should use switch statement

src/transformState.js
```diff
@@ -5,7 +5,36 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const stateHistory = [];
```
choeqq, line 12: you weren't prompted to track state history, but it's good idea 😄

choeqq (changes requested):
Almost there 🚀

choeqq (changes requested):
Waiting for any updates @MateuszSlezak 👀

src/transformState.js
MateuszSlezak: so i keep this ;p

choeqq (approved):
Good job ✅

PR #2622, opened 2023-07-27
feat: implement a transforming state (Thu, Jul 27, 2023 02:10:42 PM)

FrontKid:


src/transformState.js
```diff
@@ -4,8 +4,72 @@
...
+  if (isObject(actions)) {
+    throw new Error('Actions is not an object');
+  }
+
+  for (const action of actions) {
+    const { type } = action;
+
+    switch (type) {
+      case 'addProperties':
+        const { extraData } = action;
```
AndriiKuz, line 23: I think in these places using destructuring isn't a good idea. it will be better to use action.extraData.

FrontKid: But why?

AndriiKuz: It is better to use destructuring to obtain several values ​​from an object. If we need only one value, we can directly take it from the object.
it will just be cleaner and have less code. But this is just a recommendation.

AndriiKuz (approved):

roman-mirzoian (approved):
The work is well done and it is clear that you understand the correct approaches to writing code 🔥 
But I agree about destructuring - if you only want to get one field, it's better to use it via an object to avoid overloading your code.

PR #2627, opened 2023-07-28
Stateful object

stunnerboyone:


src/transformState.js
```diff
@@ -4,8 +4,32 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
   // write code here
+  for (let i = 0; i < actions.length; i++) {
```
denys-danyliuk: use for of

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+    switch (actions[i].type) {
+      case 'addProperties':
+        Object.assign(state, { ...actions[i].extraData });
+        break;
+      case 'removeProperties':
+        for (let j = 0; j < actions[i].keysToRemove.length; j++) {
```
denys-danyliuk: same here, it's unnecessary complicated

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        const errorMessage = new Error('Array have not correct action value');
+
+        return errorMessage;
```
denys-danyliuk:
```suggestion
        throw new Error('Array have not correct action value');
```

src/transformState.js
```diff
@@ -4,8 +4,32 @@
...
+    }
+  }
+
+  return state;
```
denys-danyliuk: function doesn't return anything, it just transforms state

denys-danyliuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,28 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
 function transformState(state, actions) {
   // write code here
+  for (const action in actions) {
+    switch (actions[action].type) {
```
andriy-shymkiv: `actions` is array. so which method we should use to run thorough the array?

```suggestion
  for (const action of actions) {
    switch (action.type) {
```

andriy-shymkiv: use this approach everywhere

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+      case 'addProperties':
+        Object.assign(state, { ...actions[action].extraData });
+        break;
+      case 'removeProperties':
+        for (const keyToRemove in actions[action].keysToRemove) {
+          delete state[actions[action].keysToRemove[keyToRemove]];
+        }
```
andriy-shymkiv: this loop is really complicated. after using a `for of` loop above -  you can make this one mutch simpler

andriy-shymkiv (changes requested):
good job! left comments

src/transformState.js
```diff
@@ -4,8 +4,28 @@
...
+
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, { ...action.extraData });
+        break;
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
```
andriy-shymkiv, line 16: looks better now, but `action` is object here, so why did you changed it to `for of`?
remember - use `for of` for arrays, and `for in` for objects

andriy-shymkiv (changes requested):

src/transformState.js
stunnerboyone: image
action is object but action.keysToRemove an array

xiimiik (approved):
GJ

PR #2628, opened 2023-07-29
stateful object v1

oltarasov:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    }
+
+    if (obj.type === 'removeProperties') {
+      for (const toDelete of obj.keysToRemove) {
+        delete state[toDelete];
+      }
+    }
+
+    if (obj.type === 'clear') {
+      Object.keys(state).forEach(key => delete state[key]);
```
VitaliyBondarenko1982: use `for in` loop

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    }
```
VitaliyBondarenko1982: instead `if` consider to use `switch case`

VitaliyBondarenko1982 (changes requested):
Good job.
Consider some improvements

src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
```
```suggestion
  for (const action of actions) {
```

nazarmatsevych (approved):
Good work, left minor comment

PR #2629, opened 2023-07-30
Solution

polinavafik:


DenisChakhalian (approved):
good job!

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+/* - `state` is an object.
+- actions is an array of objects
+type contains  `'addProperties'` `'removeProperties'` `'clear'`;
+- if `type` is `addProperties`, second property is `extraData`.
+- if `type` is `removeProperties`, second property is `keysToRemove`
+- if `type` is `clear` you should remove all the properties
+ Use `switch
+if key is a variable - `object[key]`.
+If it's called key -  `object.key``
+*/
```
qoa11a, line 17: You should remove redundant comments when you don't need them anymore.

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => delete state[key]);
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
```
qoa11a, line 31: It's one of the ways to delete all properties from an object. The other one is to assign an empty object to the `state` variable. In the second case, you'll create a new reference so they're not identical.

qoa11a (approved):
Very good! 👍

PR #2632, opened 2023-07-30
solution

DenisChakhalian:


KatOlista (approved):
good job!

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
+      case 'removeProperties':
+        for (const keyToRemove of action.keysToRemove) {
+          delete state[keyToRemove];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```

qoa11a (approved):
It is nicely done!

PR #2633, opened 2023-07-30
add task solution

yar0man:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+    }
```
DenisChakhalian: checklist: [CODE STYLE]: switch/case should always have default case for error handling.
example in the photo:
image

DenisChakhalian (changes requested):
good job 🚀
fix one thing

DenisChakhalian (approved):
good job!

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
```
roman-mirzoian: It is a little easier to add `extraData` using `Object.assign`.

roman-mirzoian (changes requested):

qoa11a (approved):

PR #2635, opened 2023-07-30
Develop

KatOlista:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
AndrMar1939:
```suggestion
  for (let i = 0; i < actions.length; i++) {
```
'For of' loop would be more readable

AndrMar1939 (approved):
Good job! But let`s make your code better.

src/transformState.js
```diff
@@ -5,7 +5,23 @@
...
+      case 'addProperties':
+        Object.assign(state, actions[i].extraData);
+        break;
+      case 'removeProperties':
+        for (const action of actions[i].keysToRemove) {
+          delete state[action];
+        };
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
```
qoa11a, line 19: It's one of the ways to delete all properties from an object. The other one is to assign an empty object to the `state` variable. In the second case, you'll create a new reference so they're not identical.

src/transformState.js
qoa11a: It's better to use the `for ... of` loop here.

qoa11a (changes requested):
Good job! 👍 
Please, consider rewriting your solution using the `for ... of` loop.

Moroz-Dmytro (approved):
Well done 👍

PR #2637, opened 2023-07-31
Solution

neovendano:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const prop in state) {
+          if (state.hasOwnProperty(prop)) {
+            delete state[prop];
+          }
+        }
```
polinavafik: I would add a blank line after each break, so it would be easier to read 🤝

```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;
        
      case 'clear':
        for (const prop in state) {
          if (state.hasOwnProperty(prop)) {
            delete state[prop];
          }
        }
```

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+        break;
```
polinavafik, line 35: and maybe put 
 default:
      break;

polinavafik (commented):
Great code!!!
There is just some minor things that you could change🎀

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
```
maksym-mishchanchuk: [CODE STYLE]: Nested loops === EVIL

maksym-mishchanchuk (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,37 @@
...
+      case 'removeProperties':
+        stateRemoveProperty(state, action.keysToRemove);
+        break;
+
+      case 'clear':
+        stateClear(state);
+        break;
+
+      default:
+        break;
```
dffUqp, line 35: would be nice to return an error or error message here

dffUqp (approved):
LGTM👍

PR #2640, opened 2023-07-31
add task solution

Vannnkof:


Koliras (approved):

vserhiichuk (approved):
Good job!

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => {
+          delete state[key];
+        });
```

qoa11a (approved):
Good job! 👍

PR #2641, opened 2023-07-31
Solution

AndrMar1939:


voievodik (changes requested):
Interesting version! Good job!

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const finiteStateMachine = ({ type, extraData, keysToRemove }) => {
```
roman-mirzoian: The `transformState` function should do what you described in `finiteStateMachine` and now you are overloading your code with an additional function. If you want to break it into small parts (which is the correct approach), you need to pass the code action on the state to the function (add, remove, clear).

src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```

roman-mirzoian (changes requested):

src/transformState.js
AndrMar1939: I made an investigation. 

- created a heavy Object with 1e7 keys-values;
- created performance.now() at the beginning and at the end transformState(), found the difference between them;

- results: 
1) my first solution has average performance **5500**; 

2) when I removed my inner Function and implement switch-case in every iteration, average performance  was **4800**; _You were right about perfomance!_

3) when I turned back my inner Function and replaced "break" with "return",  average performance  was **4600**;

4) when I again removed my inner Function and replaced switch-case "break" with "return",  average performance  was **4600**;

Conclusions: solutions with addition function and without it have the same average perfomance if "break" is replaced by "return";

Moroz-Dmytro (approved):
👍

src/transformState.js
roman-mirzoian: @AndrMar1939 In addition to performance, you should also pay attention to the readability and maintainability of code. What's the point of creating another function inside a function if it can be written as a single function?)

AndrMar1939: It`s a tough question! Firstly, I thought it was more efficient and clearly than creating a switch-case for each new iteration. 
But considering your arguments, and the arguments in a next problem, I realized that I was wrong. 
Thank you for a review!

PR #2642, opened 2023-07-31
Solution

ivannaromanovych:


davinnchii (approved):
In general everything is fine, but I'd prefer to use .forEach method instead of common for-loop

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+      default:
+        break;
```

roman-mirzoian (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
```

qoa11a (approved):
Nice! 👍

PR #2643, opened 2023-07-31
Develop

yuliiaant:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
```
Wesses, line 10: Maybe you should use `const` values for types, but i am not sure)))))

Wesses (approved):
Good job.
My solution is almost the same is yours, let's see what mentors will say.

dffUqp (approved):
LGTM🔥

PR #2644, opened 2023-07-31
add task solution

Koliras:


kaavka (approved):
I dont even know what to add) Maybe just some identations between cases. Good job.

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+      case 'removeProperties':
+        action.keysToRemove.forEach(key => {
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
```

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+      default: break;
```
qoa11a, line 23:
```suggestion
      default: 
        break;
```
I'd recommend inserting a new line after any of the cases.

qoa11a (approved):
Nice!

PR #2645, opened 2023-07-31
add task solution

voievodik:


src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.map(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.keys(state).map(key => delete state[key]);
+        break;
+      default:
```
rostyslav48, line 19: Great Job👏. However, maybe it is better to make function to return something by default?

rostyslav48 (approved):

src/transformState.js
```diff
@@ -5,7 +5,21 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
```
dffUqp, line 7: function should return state

src/transformState.js
```diff
@@ -5,7 +5,21 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+      case 'removeProperties':
+        action.keysToRemove.map(key => delete state[key]);
+        break;
+      case 'clear':
+        Object.keys(state).map(key => delete state[key]);
+        break;
```
dffUqp, line 18: we mutate our initial state let's prevent this behavior and just return new state from function

cloning objects in js- [link](https://www.freecodecamp.org/news/clone-an-object-in-javascript/)

dffUqp (changes requested):
Almost done:smiley:

src/transformState.js
voievodik: - state is an object. You are supposed to add, change, or delete its properties instead of creating a new object

src/transformState.js
voievodik: should modify the state, doing the following:

- add two properties to the state
- then remove keys bar and hello from the state
- and finally add another one property to the state

roman-mirzoian (approved):
Well done 👍

PR #2646, opened 2023-07-31
Add task solution

davinnchii:


Orpheus29 (approved):
Hello, good job altogether!
You may consider adding a `default` case, as in the checklist:
4. [CODE STYLE]: switch/case should always have default case for error handling.

src/transformState.js
```diff
@@ -6,6 +6,24 @@
  */
 function transformState(state, actions) {
   // write code here
```
roman-mirzoian, line 8: Do not leave unnecessary comments.

src/transformState.js
```diff
@@ -6,6 +6,24 @@
...
+        action.keysToRemove.forEach((key) => {
+          delete state[key];
+        });
+        break;
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
+        break;
+
+      default:
+        break;
```

roman-mirzoian (changes requested):

BudnikOleksii (approved):
Good job!
Don't forget to remove comments.

PR #2648, opened 2023-07-31
add task solution

Wesses:
If you see this please explain me this paragraph from checklist.
[CODE STYLE]: Nested loops === EVIL.
I dont understand how to delete all fields from object without loop while I am already in loop.
Mark my misstakes, ty

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const TYPE_ADD_PROPERTIES = 'addProperties';
+  const TYPE_REMOVE_PROPERTIES = 'removeProperties';
+  const TYPE_CLEAR = 'clear';
+
+  for (const action of actions) {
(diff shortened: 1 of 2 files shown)


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const TYPE_ADD_PROPERTIES = 'addProperties';
+  const TYPE_REMOVE_PROPERTIES = 'removeProperties';
+  const TYPE_CLEAR = 'clear';
```
ydratsevska: You don't need to create new variables. You can use 'addProperties', 'removeProperties',  'clear'.  We need to create new variables to describe value instead hardcoding. It's not such case.

ydratsevska (changes requested):
Good job!!! Please change your variables to  'addProperties', 'removeProperties', 'clear'.

ydratsevska (approved):

roman-mirzoian (approved):
Well done 👍

PR #2649, opened 2023-07-31
Stateful object task

vserhiichuk:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+          state[key] = action.extraData[key];
+        };
+        break;
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
```
Taras2705: good job but there is some mistake
1.[ CODE STYLE]: don't use for in loop for iterating over array

Taras2705 (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        for (const key of Object.keys(action.extraData)) {
+          state[key] = action.extraData[key];
+        };
```
roman-mirzoian: It is a little easier to add `extraData` using `Object.assign`.

roman-mirzoian (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          delete state[key];
+        }
+        break;
+
+      default:
+
```
Moroz-Dmytro, line 27:
```suggestion
```

Moroz-Dmytro (approved):
👍

PR #2651, opened 2023-08-01
Solution

serhiilapin:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let ACTION_REQUIRED = '';
+  
+
+  for (const action of Object.values(actions)) {
+    ACTION_REQUIRED = action['type'];
```
VadimMentorQA, line 12: Instead of accessing object properties with action['type'], use dot notation (action.type) for better readability.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  let ACTION_REQUIRED = '';
```
VadimMentorQA, line 8: The variable ACTION_REQUIRED should be in camel case (actionRequired) according to JavaScript naming conventions.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+  
+
+  for (const action of Object.values(actions)) {
+    ACTION_REQUIRED = action['type'];
+
+    switch (ACTION_REQUIRED) {
+      case 'addProperties':
+        Object.assign(state, action['extraData']);
```
VadimMentorQA, line 16: Instead of accessing object properties with action['extraData'], use dot notation (action.extraData). Ensure that the 'extraData' property is correct as per the task description.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const parameters of action['keysToRemove']) {
```
VadimMentorQA, line 20: Instead of accessing object properties with action['keysToRemove'], use dot notation (action.keysToRemove). Ensure that the 'keysToRemove' property is correct as per the task description.

VadimMentorQA (approved):
Accepted with comments within

PR #2653, opened 2023-08-01
Solution

ydratsevska:


Valentyn-Bilyk (approved):
I think well done

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
qoa11a, line 13: It's one of the ways to delete all properties from an object. The other one is to assign an empty object to the `state` variable. In the second case, you'll create a new reference so they're not identical.

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      default: break;
```
qoa11a, line 26:
```suggestion
      default: 
        break;
```
I'd recommend inserting a new line after any of the cases.

qoa11a (approved):
Good job! Keep up the great work!

PR #2655, opened 2023-08-01
solution

Taras2705:


src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
```
desevoker: You can extract such repeating blocks into a separate function.

Also, you can omit checks for `hasOwnProperty`, `delete` will do nothing if property doesn't exist.

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+      case 'clear' :
+        for (const key of Object.keys(state)) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      default:
+        break;
```
desevoker: That doesn't look good, you shouldn't add something just for the sake of adding it.
Returning an error message there would be much better.

desevoker (approved):
Goob job! 🚀

Left a few recommendations for you.

src/transformState.js
Moroz-Dmytro: Good point

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear' :
+        for (const key of Object.keys(state)) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
Moroz-Dmytro: Same here

src/transformState.js
Moroz-Dmytro: Good point 👍

Moroz-Dmytro (changes requested):
Good work 👍 
But check comments

Moroz-Dmytro (approved):
👍

PR #2656, opened 2023-08-01
task solution

viktoria-roik:


dpidlutska (approved):
LGTM! 🔥

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        });
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+      default:
+        break;
```
Moroz-Dmytro: Better will be to return an error here or some message that something went wrong
image

Moroz-Dmytro (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+
+      case 'clear':
+        Object.keys(state).forEach(key => {
+          delete state[key];
+        });
+        break;
+      default:
+        return new Error('Something went wrong');
+    }
```
dffUqp, line 27:
```suggestion
      case 'clear':
        Object.keys(state).forEach(key => {
          delete state[key];
        });
        break;
        
      default:
        return new Error('Something went wrong');
    }
```

dffUqp (approved):
GJ✔️

PR #2657, opened 2023-08-01
task solution

no4kar:
without nested loop

### DIFF
--- a/src/transformState.js
+++ b/src/transformState.js
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const removeKeyFromStateObj = (key) => {
+    delete state[key];
+  };
+
+  for (const action of actions) {
+    if (Object.hasOwnProperty.call(action, 'type')) {
+      switch (action.type) {
+        case 'addProperties':
+          Object.keys(action.extraData).forEach(
+            (key) => {
+              state[key] = action.extraData[key];
+            });
+          continue;
+
+        case 'removeProperties':
+          action.keysToRemove.forEach(removeKeyFromStateObj);
+          continue;
+
+        case 'clear':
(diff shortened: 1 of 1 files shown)


src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const removeKeyFromStateObj = (key) => {
+    delete state[key];
+  };
+
+  for (const action of actions) {
+    if (Object.hasOwnProperty.call(action, 'type')) {
```
Esceype: Remove this as you have default case

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+      switch (action.type) {
+        case 'addProperties':
+          Object.keys(action.extraData).forEach(
+            (key) => {
+              state[key] = action.extraData[key];
+            });
+          continue;
```
Esceype: Why do you need `continue` here and in other cases ?

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+  };
+
+  for (const action of actions) {
+    if (Object.hasOwnProperty.call(action, 'type')) {
+      switch (action.type) {
+        case 'addProperties':
+          Object.keys(action.extraData).forEach(
+            (key) => {
+              state[key] = action.extraData[key];
+            });
```
Esceype: Rewrite it using Object.assign

Esceype (changes requested):
Great job!!! Read the following comments bellow and let`s improve it a little bit

src/transformState.js
no4kar: how can I be sure that this key('type') exist in nested object?

src/transformState.js
no4kar: thank you. with this looks better.

TarasHoliuk (approved):
:+1:

PR #2659, opened 2023-08-01
add task solution

irahos:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          }
+        });
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach((key) => delete state[key]);
+        break;
+
+      default:
+        throw Error('Invalid input.');
```
KlymSudakov: Check correct error declaration

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        action.keysToRemove.forEach((key) => {
+          if (state.hasOwnProperty(key)) {
```
KlymSudakov: Not sure if there is a need to check this condition - if (state.hasOwnProperty(key))

KlymSudakov (changes requested):

Moroz-Dmytro (approved):
👍

PR #2661, opened 2023-08-02
added task transforming state

petro-boreiko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
```
TarasHoliuk: use `switch` instead of `if` blocks

TarasHoliuk (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
sTorba24, line 25: it should be in case `clear`

sTorba24: for default return error

sTorba24 (changes requested):
Almost good! Lets improve it

Anastasiia-Svintsova (approved):

PR #2663, opened 2023-08-02
add task solution

edgergard:


ihorid (approved):

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (let i = 0; i < actions.length; i++) {
+    switch (actions[i].type) {
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
+          state[key] = actions[i].extraData[key];
+        }
```
dffUqp: we can make it easier by using `Object.assign`

src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
-}
+  for (let i = 0; i < actions.length; i++) {
```
dffUqp: let's replace it with [for...of](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of)

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+      case 'removeProperties':
+        for (const key of actions[i].keysToRemove) {
+          delete state[key];
+        }
+        break;
 
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
```
dffUqp, line 25: we clear all values from our object so can we just assign new empty object to avoid additional loop?

dffUqp (changes requested):
Almost done:smiley:

Moroz-Dmytro (approved):
GJ 👍

PR #2665, opened 2023-08-02
add task solution

dpidlutska:


Denys14Samoilenko (changes requested):
Good job!! 👍🏻 
tests passed, but if you insert code just in VS code,  your function return undefined. because you forgot return value from function.

BudnikOleksii (approved):
Good work🚀

PR #2668, opened 2023-08-02
Solution

Oleh0293:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
...
+          delete state[key];
+        }
+        break;
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        return 'Unexpected error';
```
Moroz-Dmytro, line 25:
```suggestion
      case 'addProperties':
        Object.assign(state, action.extraData);
        break;
        
      case 'removeProperties':
        for (const key of action.keysToRemove) {
          delete state[key];
        }
        break;
        
      case 'clear':
        for (const key in state) {
          delete state[key];
        }
        break;
        
      default:
        return 'Unexpected error';
```

Moroz-Dmytro: It's easier to read. But even better would be to use curly braces(read more [here](https://mrcoles.com/curly-braces-javascript-switch/))

Moroz-Dmytro (approved):
Well done 👍 
But check advice for the future

PR #2670, opened 2023-08-02
add task solution

cty2802:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
```
k1NAs: In my opinion, in this case, the better solution would be to use the `for-of` cycle. This will give you the opportunity not to create additional `variables` to get the necessary values from the `actions` array.

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
+          delete state[removeValue[j]];
+        }
+
+        break;
+
+      default:
+        Object.keys(state).forEach(key => delete state[key]);
```
k1NAs, line 29: It would be cool to add `case 'clear':` for this state, and after thet to display an `error` message in `default:`. It will help you to understend that something went wrong.

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+    switch (typeOfAction) {
+      case 'addProperties':
+        for (const newProperty in extraDatavalue) {
+          state[newProperty] = extraDatavalue[newProperty];
+        }
+        break;
+
+      case 'removeProperties':
+        for (let j = 0; j < removeValue.length; j++) {
```
k1NAs: The same here. Try to use `for-of` or `forEach()` loop instead of just `for` loop

k1NAs (changes requested):
Great Job!!! :muscle:
But there is some moments to fix

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const extraDatavalue = action.extraData;
+    const typeOfAction = action.type;
+    const removeValue = action.keysToRemove;
```
BudnikOleksii: Not all actions have extraData and keysToRemove.

src/transformState.js
```diff
@@ -5,7 +5,37 @@
...
+
+        break;
+
+      case 'clear':
+        Object.keys(state).forEach(key => delete state[key]);
+
+        break;
+
+      default:
+        return 'Error!';
```
BudnikOleksii: If you want to notify a user about errors it's better to throw exceptions, you can check [docs ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/throw?retiredLocale=uk) how to use it in js.
Because from this function, we expect returned object "state", but we get a string instead and no errors in this case.

BudnikOleksii (changes requested):
Good job👍
Left a few comments, check them.

src/transformState.js
```diff
@@ -5,7 +5,37 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const extraDataValue = action.extraData || {};
+    const typeOfAction = action.type;
+    const removeValue = action.keysToRemove || {};
```
roman-mirzoian, line 11: In order not to do unnecessary checks, you can not create variables for these objects, but directly use them where they can be.

roman-mirzoian (approved):
Well done 👍

k1NAs (approved):
Good job! :rocket:

PR #2671, opened 2023-08-02
Solution

00Mass00:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          delete state[property];
+        }
+        break;
+
+      case 'clear':
+        if (action.type === 'clear') {
```
BudnikOleksii: You don't need "if" statement here as you check action.type with switch case.

BudnikOleksii (changes requested):
Nice job👍
Just remove unnecessary if statement.

mr-olezha (approved):
good job bro

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Hey ho!');
```
Moroz-Dmytro, line 27: But give more descriptive error message. It's a bad example

Moroz-Dmytro (approved):
👍

PR #2672, opened 2023-08-02
transform state solution

Denys14Samoilenko:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
+    }
+  }
+
+  return state;
```
vatatan, line 31: we don't need to return "state" here
but still looks cool👍
good job🔥

vatatan (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        break;
```
Moroz-Dmytro: image

Would be nice to throw an error in such case

Moroz-Dmytro (changes requested):
Well done 👍 
But small comment

dffUqp (approved):
LGTM🚀

PR #2675, opened 2023-08-02
add task solution

NazarVotkanych:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+      default:
+        break;
```
dffUqp, line 25: would be nice to return an error or error message here

dffUqp (approved):
LGTM🚀

PR #2678, opened 2023-08-03
task_solution

BeginerAlex:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+            delete state[property];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
```
sTorba24, line 26: you forgot about default case
image
throw error for `default`

sTorba24 (changes requested):
Almost good!

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+
+      default:
```
sTorba24, line 28: return error for default case to explain what happened

sTorba24 (changes requested):

src/transformState.js
BeginerAlex: I tried to return console.log('error') and I also tried to return alert('error') but the linter wouldn't let me do it. What else can I return in this case?

BeginerAlex: I made return 'error'; is this option suitable?

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const property of action.keysToRemove) {
+          if (property in state) {
```
loralevitska, line 16: The if statement is not necessary when deleting a property. If the property does not exist, the delete operation will have no effect. You can safely remove this check.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        break;
+
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+        break;
+
+      default:
+        return 'error';
```
loralevitska, line 29:
```suggestion
        return throw new Error('.....');
```

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```

loralevitska (approved):
Well done)

src/transformState.js
BeginerAlex: when I make so, linter rejects it

PR #2679, opened 2023-08-03
add task solution

k1NAs:


bogemna:
looks good!

bogemna (approved):
good

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
Moroz-Dmytro: You don't need to do this check, if `key` doesn't present, nothing happens

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
```
Moroz-Dmytro: Same here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        }
+        break;
+
+      default:
+        return new Error('Oops, something went wrong!');
```
Moroz-Dmytro: Better to [throw](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/throw) an error, not to return

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
+
+      case 'clear':
+        for (const key of Object.keys(state)) {
```
Moroz-Dmytro: Can you use [for in](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Statements/for...in) loop here?

Moroz-Dmytro (changes requested):
Well done 👍 
Small comments

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        throw new Error('Oops, something went wrong!');
```
Moroz-Dmytro, line 27: But try to give error message that describes what type of error it is

Moroz-Dmytro (approved):
👍

PR #2680, opened 2023-08-03
Solution

MariaSnegireva:


src/transformState.js
```diff
@@ -5,7 +5,39 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case ('addProperties'):
+        const extraData = action.extraData;
+
+        for (const key in extraData) {
```
anastasiiavorobiova: Consider using Object.assign instead of the cycle here

src/transformState.js
```diff
@@ -5,7 +5,39 @@
...
+          state[key] = extraData[key];
+        }
+
+        continue;
+
+      case ('removeProperties'):
+        const keysToRemove = action.keysToRemove;
+
+        for (const key of keysToRemove) {
+          if (state.hasOwnProperty(key)) {
```
anastasiiavorobiova: There is no need to add an additional check here. The data is already valid

anastasiiavorobiova (changes requested):
Great work! Almost perfect!

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case ('addProperties'):
```
```suggestion
const ADD = 'addProperties';

      case ADD: {
```

Consider using variables instead of magic strings

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        const extraData = action.extraData;
+
+        Object.assign(state, extraData);
+
+        continue;
+
+      case ('removeProperties'):
```
anastasiiavorobiova: The same issue here

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+        const keysToRemove = action.keysToRemove;
+
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+
+        continue;
+
+      case ('clear'):
```
anastasiiavorobiova: See my previous comment

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
```
anastasiiavorobiova, line 8: Declare here all needed variables and use them below

anastasiiavorobiova (changes requested):
Great solution! Almost done

loralevitska (approved):
Good job!

PR #2682, opened 2023-08-03
solution

vladyslav-svorin:


src/transformState.js
```diff
@@ -5,7 +5,34 @@
...
+        break;
+
+      case 'addProperties':
+        for (const key in actions[i].extraData) {
+          state[key] = actions[i].extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        if (actions[i].keysToRemove !== undefined) {
```
loralevitska: There's no need for checking if 'keysToRemove' is undefined. If it's undefined, the later for loop just won't execute.

src/transformState.js
```diff
@@ -5,7 +5,34 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let i = 0; i < actions.length; i++) {
+    const mainAction = actions[i].type;
```
loralevitska: Instead of traditional for loop, use of 'for...of' loop is recommended for better readability when iterating over an array.

loralevitska: Destructure action
`const { type, extraData, keysToRemove } = actions[i]`

loralevitska (changes requested):
Good job! There are things that should be fixed:

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+        break;
+
+      case 'addProperties':
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        const toRemove = keysToRemove;
```
VitaliyBodnar: this variable is reduntant here

src/transformState.js
```diff
@@ -5,7 +5,32 @@
...
+
+        for (let j = 0; j < toRemove.length; j++) {
+          delete state[toRemove[j]];
+        }
```
VitaliyBodnar: don't see any sense in using  for loop, better to use` for of` or `forEach` here

VitaliyBodnar (changes requested):
Almost done!

src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+
+      case 'addProperties':
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        keysToRemove.forEach(element => delete state[element]);
+    }
```
polosanya, line 34: image

polosanya (changes requested):
One small but important fix left 🙏

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case 'clear':
```

Example of usage:
  `case ActionType.Clear:`

OleziO (changes requested):

VitaliyBodnar (changes requested):
Good job!

VitaliyBodnar (approved):

PR #2683, opened 2023-08-03
Complete task

Annbpiu:


src/transformState.js
```diff
@@ -5,7 +5,28 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return 'Error';
```
etojeDenys: better to throw the error

etojeDenys (changes requested):

VitaliyBodnar (approved):
Good job!

PR #2686, opened 2023-08-04
Solution

RomanPerepelkin:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const obj of actions) {
+    if (obj.type === 'addProperties') {
```
volodymyr-soltys97: Use destructuring for `obj`

volodymyr-soltys97: Use switch here will be better

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      Object.assign(state, obj.extraData);
+    }
+
+    if (obj.type === 'removeProperties') {
```
volodymyr-soltys97: Move this magic words to variable and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

anastasiiavorobiova (approved):
Awesome!

PR #2687, opened 2023-08-04
Develop

shikoryak:


src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
```
IvanFesenko: Add separete case for delete. By default should be an error in case wrong action

IvanFesenko (changes requested):
Good job. Check my comment

src/transformState.js
```diff
@@ -5,7 +5,24 @@
...
+        break;
+
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+    }
+  }
```
IvanFesenko: Where is default case?
Please check [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)
2023-08-04_15h53_52

IvanFesenko (changes requested):
Not fully fixed

src/transformState.js
```diff
@@ -5,7 +5,28 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
mykhalenych, line 9: you can destruct action
const {type, extraData, keysToRemove) = action

mykhalenych (approved):

PR #2692, opened 2023-08-04
Solution

okanivan:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+            delete state[key];
+          }
+        }
+        break;
+
+      default:
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
```
alexander-ignatow: This is not correct

Such actions should be made on `clear`

What if someone passes `AddProperties` in mistake?
It is good idea to always handle `default` case, but you should do it correctly

alexander-ignatow (changes requested):
Looks good - but you need to fix your `switch` clause

BudnikOleksii (changes requested):
The switch case must have a default case, you can throw an exception if the action type isn't what you expected.

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        Object.assign(state, action.extraData);
+        break;
+
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        }
+        break;
```
tiserett, line 20: preferably to add curly braces in case we have few lines of code inside of case block

tiserett (approved):

PR #2693, opened 2023-08-04
solution

Duster-1:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+      case 'clear':
+        for (const key in state) {
+          delete state[key];
+        }
+
+        break;
+
+      default:
+        break;
+    }
```
nazarmatsevych, line 31: For example you can throw an error in default case with message like: 'Unknown action type'

nazarmatsevych (approved):

PR #2694, opened 2023-08-05
Solution

MOstapchyk:


src/transformState.js
```diff
@@ -6,6 +6,32 @@
  */
 function transformState(state, actions) {
   // write code here
+  const transformedState = state;
+
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformState.js
```diff
@@ -6,6 +6,32 @@
...
+      case 'addProperties':
+        Object.assign(transformedState, action.extraData);
+        break;
+
+      case 'removeProperties':
```
volodymyr-soltys97: Move these magic words to variables and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Nice work 🔥

PR #2696, opened 2023-08-06
finish solution

serkrops:


src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+        }
+
+        break;
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+
+        break;
+      default:
```
bogemna, line 23: Overall, your code seems to align well with the checklist. The only suggestion is to add a default case in your switch statement for error handling, as mentioned in point 4 in checklist

[CODE STYLE]: switch/case should always have default case for error handling.

well done

bogemna (approved):

src/transformState.js
```diff
@@ -5,7 +5,30 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
+
+        break;
```
tiserett: these gaps are redundant here

src/transformState.js
```diff
@@ -5,7 +5,30 @@
...
+      case 'removeProperties':
+        for (const key of action.keysToRemove) {
+          delete state[key];
+        }
+
+        break;
```
tiserett: same with gaps here + if you have few lines inside of case block - better to add curly braces + let's use array methods here and everywhere else ;)

tiserett (changes requested):

BudnikOleksii (approved):
Good job👍

PR #2697, opened 2023-08-07
Solution

AnastasiiaKriuchkova:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+        break;
+      case 'clear':
+        for (const property in state) {
+          delete state[property];
+        }
+
+        break;
+      default:
+        return state;
+    }
```
IvanFesenko: 2023-08-07_16h07_28
[checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md)

IvanFesenko (changes requested):
Good job. Please fix default case

polosanya (approved):
👍

PR #2700, opened 2023-08-07
solution

mykyta01:


src/transformState.js
```diff
@@ -6,6 +6,36 @@
...
+
+    switch (type) {
+      case 'addProperties':
+        for (const key in extraData) {
+          state[key] = extraData[key];
+        }
+        break;
+
+      case 'removeProperties':
+        if (keysToRemove) {
```
lerastarynets: Unnecessary condition. If 'keysToRemove' is not provided, the for...of loop will simply not run. Therefore, the if (keysToRemove) condition is redundant and can be removed.

src/transformState.js
```diff
@@ -6,6 +6,36 @@
...
+          for (const key of keysToRemove) {
+            if (state.hasOwnProperty(key)) {
```
lerastarynets: The hasOwnProperty check is not necessary when deleting a property. JavaScript will not throw an error if you try to delete a property that does not exist. Thus, this line can be safely removed.

src/transformState.js
```diff
@@ -6,6 +6,36 @@
  */
 function transformState(state, actions) {
   // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
lerastarynets: Inconsistent destructuring of 'action'. The task description mentions that the second property of the action object can vary depending on the type, but here you're destructuring 'extraData' and 'keysToRemove' regardless of the type. You should move the destructuring inside the corresponding case blocks.

lerastarynets (changes requested):
Nice start! Let's fix and improve your work a little bit)

src/transformState.js
xiimiik: not fixed

xiimiik (changes requested):

Esceype (approved):
Perfect!!!

PR #2704, opened 2023-08-08
js_stateful-object solution 

Medaed:


src/transformState.js
```diff
@@ -5,7 +5,29 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const actionType = action.type;
+    const actionExtraData = action.extraData;
+    const actionKeysToRemove = action.keysToRemove;
```
VitaliyBodnar: don't see any sense to create these redundant constants

src/transformState.js
```diff
@@ -5,7 +5,29 @@
...
+    if (actionType === 'removeProperties') {
+      for (const property of actionKeysToRemove) {
+        delete state[property];
+      }
+    }
+
+    if (actionType === 'clear') {
+      for (const element in state) {
+        delete state[element];
+      }
```
VitaliyBodnar: it's better to use a switch case statement when you have multiple conditions

VitaliyBodnar (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,32 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const actionType = action.type;
+
+    switch (actionType) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
VitaliyBodnar, line 13: you modified initial state, it's a bad idea
you need to create a clone of the state and modify this cloned object

VitaliyBodnar (changes requested):

src/transformState.js
Medaed: image

PR #2705, opened 2023-08-08
the task was implemented

oirevlis1:


sTorba24 (approved):

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      for (const key of action.keysToRemove) {
+        delete state[key];
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
sTorba24: Use `switch case` instead of `ifs` here

sTorba24 (changes requested):

polosanya (approved):
👍

PR #2707, opened 2023-08-09
task solution

ZadorozhnyiYevhenii:


src/transformState.js
```diff
@@ -5,7 +5,33 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,33 @@
...
+      case 'addProperties':
```
volodymyr-soltys97: Create a variables for these magic words and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

volodymyr-soltys97 (approved):
Well done 🔥

PR #2708, opened 2023-08-10
Stateful object. Tests passed.

momos1703:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```
mykhalenych, line 9: you can destruct action
const {type, extraData, keysToRemove} = action

mykhalenych (approved):

PR #2710, opened 2023-08-10
jsStatefullObject V1

Voloshyn24:


src/transformState.js
```diff
@@ -6,6 +6,31 @@
...
+      case _removeProperties:
+        for (const removeKey of keysToRemove) {
+          delete state[removeKey];
+        }
+        break;
+      case _clear:
+        Object.keys(state).forEach(key => delete state[key]);
+        break;
+      default:
+        return 'Type error!';
```
Anastasiia-Svintsova: It's better to throw an actual JavaScript error to provide more context about what went wrong.
```suggestion
        throw new Error(`Unknown action type: ${type}`);
```

src/transformState.js
```diff
@@ -6,6 +6,31 @@
...
+  const _clear = 'clear';
+
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (type) {
+      case _addProperties:
+        Object.assign(state, extraData);
+        break;
+      case _removeProperties:
```
Anastasiia-Svintsova: Add spaces for better code readability 
```suggestion
        break;
        
      case _removeProperties:
```

Anastasiia-Svintsova (changes requested):
Almost done!

sTorba24 (approved):

PR #2712, opened 2023-08-12
add task solution

pr1syazhniy:


src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
```

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+      case 'addProperties':
+        if (action.extraData && typeof action.extraData === 'object') {
+          for (const [key, value] of Object.entries(action.extraData)) {
+            state[key] = value;
+          }
+        }
+        break;
+      case 'removeProperties':
```
volodymyr-soltys97: Move these magic words to variable and use it everywhere

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

nazarmatsevych (approved):

PR #2713, opened 2023-08-13
Solution

MyroslavSynenko:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case 'addProperties':
+        Object.assign(state, action.extraData);
```
VitaliyBodnar, line 11: you need to create a copy of the state and use it because you are currently modifying the current state

VitaliyBodnar (changes requested):
Almost done!

VitaliyBondarenko1982 (approved):

PR #2714, opened 2023-08-14
Solution

patrykyo20:


src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+      for (const key of keysToRemove) {
+        if (state.hasOwnProperty(key)) {
+          delete state[key];
+        }
+      }
+    } else if (action.type === 'clear') {
+      for (const key in state) {
+        delete state[key];
+      }
+    }
```
VitaliyBodnar: - please use the switch case statement here
- move 'addProperties', 'removeProperties', 'clear' to the constants as it will make a code cleaner

src/transformState.js
```diff
@@ -5,7 +5,26 @@
...
+    if (action.type === 'addProperties') {
+      const extraData = action.extraData;
+
+      for (const key in extraData) {
+        state[key] = extraData[key];
+      }
+    } else if (action.type === 'removeProperties') {
+      const keysToRemove = action.keysToRemove;
 
+      for (const key of keysToRemove) {
```
VitaliyBodnar: redundant variable 
please check other places 
```suggestion
      for (const key of action.keysToRemove) {
```

VitaliyBodnar (changes requested):

src/transformState.js
```diff
@@ -4,8 +4,40 @@
  * @param {Object} state
  * @param {Object[]} actions
  */
+
+const add = 'addProperties';
+const remove = 'removeProperties';
+const clear = 'clear';
```
mvjl000: Variable names like this are typically uppercase
```suggestion
const ADD = 'addProperties';
const REMOVE = 'removeProperties';
const CLEAR = 'clear';
```

src/transformState.js
```diff
@@ -4,8 +4,40 @@
...
+
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    switch (action.type) {
+      case add:
+        for (const key in action.extraData) {
+          state[key] = action.extraData[key];
+        }
```
mvjl000: You could avoid a nested loop here by using `Object.assign`

mvjl000 (changes requested):

mbruhler (approved):
Approved.

`Object.assign` is cool but more modern way to achieve the same this is using `destructuring`

More here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

It's just a note

patrykyo20:
Thank you, this is a valuable tip.

PR #2715, opened 2023-08-14
add task solution

Chumakv:


src/transformState.js
```diff
@@ -5,7 +5,25 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  actions.forEach(action => {
+    switch (action.type) {
```
volodymyr-soltys97, line 13: Use a destructuring for `action`

src/transformState.js
```diff
@@ -5,7 +5,25 @@
...
+      case 'addProperties':
```
volodymyr-soltys97: Move these magic words to variables and use it everywher

volodymyr-soltys97 (changes requested):
Good job 👍 
Let's improve your code

PR #2718, opened 2023-08-15
add task solution

MishaLehotskyi:


src/transformState.js
```diff
@@ -5,7 +5,31 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  const ADD_PROPERTIES = 'addProperties';
+  const REMOVE_PROPERTIES = 'removeProperties';
+  const CLEAR_PROPERTIES = 'clear';
```
andriy-shymkiv: why you hard coded all this type's instead of using switch case?

andriy-shymkiv: [checklist](https://github.com/mate-academy/js_stateful-object/blob/master/checklist.md) rule 2

src/transformState.js
```diff
@@ -5,7 +5,31 @@
...
+
+  for (const action of actions) {
+    if (action.type === ADD_PROPERTIES) {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === REMOVE_PROPERTIES) {
+      for (const remove of action.keysToRemove) {
```
andriy-shymkiv:
```suggestion
      for (const key of action.keysToRemove) {
```

andriy-shymkiv (changes requested):

xiimiik (approved):
Nice work!

PR #2719, opened 2023-08-15
Solution

dmytrosantarskyi:


src/transformState.js
```diff
@@ -5,7 +5,36 @@
...
+        break;
+
+      case 'removeProperties':
+        const keysToRemove = action.keysToRemove;
+
+        keysToRemove.forEach(key => {
+          if (state.hasOwnProperty(key)) {
+            delete state[key];
+          }
+        });
```
xiimiik, line 25: you need to use everything in the same style - either with the forEach method or with a loop

xiimiik (approved):
nice work!

PR #2724, opened 2023-08-26
task solution

TenetaJ:


src/transformState.js
```diff
@@ -5,7 +5,23 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    if (action.type === 'addProperties') {
+      for (const key in action.extraData) {
+        state[key] = action.extraData[key];
+      }
```
witflash: You can just use `Object.assign` to merge two objects.

witflash (approved):
Well done 👍

PR #2735, opened 2023-10-17
fixed

volodymyr-tytarenko:


src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+          delete state[action.keysToRemove[i]];
+        }
+      }
+    }
+
+    if (action.type === 'clear') {
+      for (const value in state) {
+        delete state[value];
+      }
+    }
```
DarkMistyRoom, line 27: If you have predetermined amount of options it is better to use switch/case structure instead of multiple
if/else blocks

src/transformState.js
```diff
@@ -5,7 +5,27 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (let key = 0; key < actions.length; key++) {
```
DarkMistyRoom, line 8: you can use `for of` here as far you don't need to know the index

src/transformState.js
```diff
@@ -5,7 +5,27 @@
...
+    const action = actions[key];
+
+    if (action.type === 'addProperties') {
+      Object.assign(state, action.extraData);
+    }
+
+    if (action.type === 'removeProperties') {
+      for (let i = 0; i < action.keysToRemove.length; i++) {
```
DarkMistyRoom, line 16: use` for of `

DarkMistyRoom (changes requested):

PR #2737, opened 2023-10-30
solution

joanna-konopacka:


src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
+        break;
+
+      case "clear":
+        for (const key in state) {
+          delete state[key];
+        }
+        break;
+
+      default:
+        return "Error: Type Unknown";
```
vlasiuksasha: In the 'default' case of your switch statement, you are returning a string. This will end the execution of the function, even if there are more valid actions in the 'actions' array. Instead, consider logging the error or ignoring the action with an invalid type.

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
+"use strict";
 
-/**
- * @param {Object} state
- * @param {Object[]} actions
- */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
```
vlasiuksasha: The destructuring of 'action' assumes that every action object has 'extraData' and 'keysToRemove' properties. This can potentially cause 'undefined' errors if these properties are absent. Consider destructuring these properties inside their respective case blocks.

src/transformState.js
```diff
@@ -1,11 +1,30 @@
...
+        Object.assign(state, extraData);
+        break;
+
+      case "removeProperties":
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+
+      case "clear":
```
vlasiuksasha, line 24: However, there is a small issue with the "clear" case. When you want to clear all properties from the state object, using a for...in loop may not work as expected. Instead, you should use Object.keys to get an array of keys and then delete each key.

vlasiuksasha (changes requested):
check the comments

Ivan-Shvorak (approved):
LGTM

PR #2741, opened 2023-12-26
Solved the task

natalia-matash:


src/transformState.js
```diff
@@ -6,6 +6,25 @@
  */
 function transformState(state, actions) {
   // write code here
```
OleziO: Remove all comments

OleziO (changes requested):

src/transformState.js
```diff
@@ -6,6 +6,25 @@
...
+  for (const action of actions) {
+    if (action.type === 'clear') {
```
Viktor-Kost: 1. use object destructing for `action`
2. it'd much better to use` switch` case here instead of `if`
3. it is a good practice when you add a default behavior for `switch` statement. Yet, if you don't have such one - just throw an Error with text like "unexpected action"

Viktor-Kost (changes requested):

src/transformState.js
```diff
@@ -5,7 +5,35 @@
  * @param {Object[]} actions
  */
 function transformState(state, actions) {
-  // write code here
+  for (const action of actions) {
+    const { type, extraData, keysToRemove } = action;
+
+    switch (true) {
```
anastasiia-tilikina: don't use true in switch statement. use type instead

src/transformState.js
```diff
@@ -5,7 +5,35 @@
...
+
+      case type === 'removeProperties': {
+        for (const key of keysToRemove) {
+          delete state[key];
+        }
+        break;
+      }
+
+      default:
+        return 'Unexpected acrion';
```
anastasiia-tilikina:
```suggestion
        return 'Unexpected action';
```

anastasiia-tilikina (changes requested):
Almost done

DarkMistyRoom (approved):
Well done!

PR #2743, opened 2024-01-22
add task solution

ArsenSkandalenko:


src/transformState.js
```diff
@@ -6,6 +6,27 @@
  */
 function transformState(state, actions) {
   // write code here
```
OleziO: Remove all comments

OleziO (changes requested):

etojeDenys (approved):
lgtm

